PR #8, opened 2025-05-06
feat: upgrade to zig 0.14

twoeths:
**Motivation**
- upgrade to zig0.14
- consumer hash-tree later, see https://github.com/ChainSafe/hashtree-z/blob/197bd98ad7d72c29183ebea2a4f336b0e5c92da1/build.zig.zon#L5


build.zig.zon
```diff
@@ -6,16 +6,29 @@
     //
     // It is redundant to include "zig" in this name because it is already
     // within the Zig package namespace.
-    .name = "state-transition-z",
+    .name = .state_transition_z,
```
twoeths, line 9: this was generated by removing the above file and issue "zig init" again

PR #9, opened 2025-05-09
feat: add official bun binding

twoeths:
**Motivation**
- consume https://github.com/ChainSafe/bun-ffi-z
- have the same DevEx like napi-rs

**Description**
- add home "bun" folder
- compile zig in bun "package.json" by: "bun run build"
- TBD

**TODOs**
- [x] blocked by https://github.com/ChainSafe/bun-ffi-z/pull/2
- [x] `bun run build`
- [x] `bun test:unit`
- [x] `bun benchmark`
- [x] test CI
- [x] `bun prePublish`. ==> tested in #11
- [x] `bun publish`
- [x] release CI


bun/src/committeeIndices.ts
```diff
@@ -20,6 +20,7 @@ export function computeProposerIndex(
 	rounds: number,
 ): number {
 	validateShufflingParams(activeIndices, seed, rounds);
+	const binding = getBinding();
```
twoeths, line 23: I cannot set binding at import time so has to do it in every api, otherwise got

```
ReferenceError: Cannot access 'binding' before initialization.
```

bun/src/binding.ts
```diff
@@ -116,10 +113,36 @@ const lib = dlopen(binaryPath, {
...
-export const binding = lib.symbols;
+/**
+ * Get the binding.
+ */
+export function getBinding(): ConvertFns<typeof fns> {
+	if (binding == null) {
+		throw new Error("Binding not initialized. Call init() first.");
+	}
+	return binding;
+}
```
wemeetagain, line 140: I don't think we should use this pattern (reinitializable binding), rather just set as const.
This will really simplify usage everywhere.

eg:
```ts
const lib = await openLibrary(...);

export const binding = lib.symbols;
export const close = lib.close;
```

If some consumer really wants to reinitialize a binding/import, they can do something like in this issue:
- https://github.com/oven-sh/bun/issues/12371

wemeetagain: why not? top-level-await should work

twoeths: oh thanks @wemeetagain , I missed that point. Added top level await and it's much nicer now for the consumer apis

bun/src/binding.ts
```diff
@@ -116,10 +113,15 @@ const lib = dlopen(binaryPath, {
...
-export const binding = lib.symbols;
+// Load the compiled Zig shared library
+const lib = await openLibrary(path.resolve("../bun"), fns);
+export const binding: ConvertFns<typeof fns> = lib.symbols;
 
-export function closeBinding(): void {
+/**
+ * Call this api to close the binding.
+ */
+export async function closeBinding(): Promise<void> {
```
wemeetagain, line 125: this can be a synchronous function (you can just export a const like `export const close = lib.close`)

bun/src/binding.ts
```diff
@@ -116,10 +113,15 @@ const lib = dlopen(binaryPath, {
 		],
 		returns: "u32",
 	},
-});
+};
 
-export const binding = lib.symbols;
+// Load the compiled Zig shared library
+const lib = await openLibrary(path.resolve("../bun"), fns);
+export const binding: ConvertFns<typeof fns> = lib.symbols;
```
wemeetagain, line 120: I don't think you need the explicit type here. openLibrary should pass through the proper type to lib.symbols

.github/workflows/test.yml
```diff
@@ -30,53 +36,178 @@ jobs:
...
+          - os: windows-latest
+            target: x86_64-windows-msvc
+    runs-on: ${{ matrix.settings.os }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Download artifacts
+        uses: actions/download-artifact@v4 # no need zig installation, download artifacts from previous job instead
+        with:
+          name: bindings-${{ matrix.settings.target }}
```
wemeetagain, line 202: I suppose we could/should add another command to bun-ffi-z or tweak how it works to better support prebuilt libraries.

Current approach with `bun-ffi-z prepublish` is to rerun zig build for all supported targets.
It would be nice to support the devex of passing in a directory of subdirectories of prebuilt libraries (one directory per target) so that rebuild for all targets is unnecessary.

Then, the publish workflow would look like:
- download all artifacts to eg `artifacts` directory
- run modified bun-ffi-z prepublish to update all `targetPackages` (move each target library into the proper location for a publish)

bun/src/committeeIndices.ts
twoeths: resolved by top-level await

bun/src/binding.ts
```diff
@@ -116,10 +113,13 @@ const lib = dlopen(binaryPath, {
 		],
 		returns: "u32",
 	},
-});
+};
 
+// Load the compiled Zig shared library
+const lib = await openLibrary(path.resolve("../bun"), fns);
```
wemeetagain, line 119: the `path.resolve("../bun")` is strange here. I don't think it will work when the cwd is another consumer of this package.

maybe better is `path.join(import.meta.dirname, '..')`

twoeths: I simplified it to `const lib = await openLibrary(path.resolve("."), fns);` please have a look at the latest change

twoeths:
the `publish` step was tested successfully with `-dry-run` flag, this is the generated `targetPackages` 
[targetPackages.zip](https://github.com/user-attachments/files/20271807/targetPackages.zip)

bun/src/binding.ts
wemeetagain: @twoeths check this comment. Try something with import.meta.dirname here instead of path.resolve/cwd

twoeths: thanks, just switched to `"import.meta.dirname"`

wemeetagain (approved):
🚀

PR #38, opened 2025-08-19
stf: block processing sanity tests

spiral-ladder:


spiral-ladder:
merging into #30 for convenience of review

Issue #21, opened 2025-07-11
Implement spec tests


twoeths:
- download spec tests
- parse data and run state transition
- reference implementation https://github.com/ChainSafe/lodestar/blob/48f2ff71ae076b87ed5b1c0eb12d9862f0fa03bc/packages/beacon-node/test/spec/utils/specTestIterator.ts#L101

twoeths:
one idea to run spec test is to write some more bun binding and utilize all utils we already have in typescript
the first usage of state-transition-z is Bun anyway

ensi321:
> one idea to run spec test is to write some more bun binding and utilize all utils we already have in typescript the first usage of state-transition-z is Bun anyway

I think we should still run state-transition spec test on `state-transition-z`, so that `state-transition-z` can be developed and released independently without relying on lodestar to do spec testing. 

Lodestar should still run all spec tests (including state-transition) on its side.

PR #46, opened 2025-09-09
feat: add minimal framework for spec test

ensi321:
### How to run
`zig build run:download_spec_tests && zig run:write_spec_tests && zig test:spec_tests`

Part of #21


test/spec/test_type/schema/operations.zig
```diff
@@ -0,0 +1,169 @@
...
+    deposit: ssz.capella.Deposit,
+    proposer_slashing: ssz.capella.ProposerSlashing,
+    voluntary_exit: ssz.capella.SignedVoluntaryExit,
+    sync_aggregate: ssz.capella.SyncAggregate,
+    body: ssz.capella.BeaconBlockBody,
+    execution_payload: ssz.capella.ExecutionPayload,
+    address_change: ssz.capella.SignedBLSToExecutionChange,
+};
+
+pub const DenebOperations = struct {
```
twoeths, line 54: if fields in `DenebOperations` and `DenebOperationsOut` are just the same we can deduplicate them, given a Container "Type" is just derived from Container

specifically we can create a function running at comptime `pub fn OutType(comptime T: type) type {` and pass in `DenebOperations` following the same technique here https://github.com/ChainSafe/ssz-z/blob/76cde3c7ea23b33d2789e8e372003272abe0e335/src/ssz/type/container.zig#L284

and instead of declaring `DenebOperationsOut`, we can do
```zig
const DenebOperationsOut = OutType(DenebOperations);
```

test/spec/test_case.zig
```diff
@@ -0,0 +1,106 @@
...
+const deneb = consensus_types.deneb;
+const electra = consensus_types.electra;
+
+// TODO: Most of the runners have fixed ssz types ie. Given a dir, we can already derive which
+// snappy file has what type and have this info written in test_type/schema.
+// There is an exception which is `transition`. One needs to read meta.yaml to figure out
+// whether a block_x.ssz_snappy belong to old fork or new fork. Current design of loadTestCase
+// does not handle this.
+pub fn loadTestCase(comptime Schema: type, comptime SchemaOut: type, dir: std.fs.Dir, allocator: std.mem.Allocator) !SchemaOut {
+    var out: SchemaOut = .{};
```
twoeths, line 21: there is some copy overhead here
having `out` params will resolve that and go with the convention in [ssz-z](https://github.com/ChainSafe/ssz-z/blob/76cde3c7ea23b33d2789e8e372003272abe0e335/src/ssz/type/container.zig#L82)
also we want the consumer to take ownership here

twoeths (commented):
general approach looks good to me, left some suggestions to deduplicate schemas and simplify things a bit, let me know if it does not work

test/spec/utils/test_logger.zig
```diff
@@ -0,0 +1,41 @@
+const std = @import("std");
+
+const GREEN = "\x1b[32m";
+const RED = "\x1b[31m";
+const DIM = "\x1b[2m";
+const RESET = "\x1b[0m";
+
+pub const TestLogger = struct {
```
wemeetagain, line 8: why are you using this TestLogger to run the actual test? Instead of just calling `runTestCase` directly inside the test body?

ensi321: We need a global object to capture the result of every tests and report the statistics at the very end of test executions. 

The `operations_tests.zig` looks something like this. Notice the `TL.summary` at the end.

```
// This file is generated by write_static_tests.zig.
// Do not commit changes by hand.

const std = @import("std");
const types = @import("consensus_types");
const ForkSeq = @import("config").ForkSeq;
const OperationsTestHandler = @import("../test_type/handler.zig").OperationsTestHandler;
const test_case = @import("./operations.zig");
var TL = @import("../utils/test_logger.zig").TestLogger{};

const allocator = std.heap.page_allocator;

test "Static - phase0 operations proposer_slashing invalid_slots_of_different_epochs" {
    try TL.run();
    ...
}

test "Static - phase0 operations proposer_slashing invalid_incorrect_sig_2" {
    try TL.run();
    ...
}

test "__summary__(Operations Tests)" {
    TL.summary("Operations Tests");
}

```

wemeetagain (approved):

PR #59, opened 2025-10-09
chore: restructure parts of `utils` for friendlier lodestar-bun usage

spiral-ladder:
Reasons:

- move parts of `utils` to `stdx` (std eXtended) `utils` can mean too many different things, we have some data structures in `utils` that can be considered foundational in state transition usage. This `stdx` is itself a new module.
- this also makes it easier to import for usage for lodestar-bun since the new `stdx` module does not rely on any imports.


twoeths (commented):
the change makes sense to me, but I don't feel like the name "stdx" much because it's still part of state-transition spec. How about name it "core"? or feel free to suggest other names

twoeths (commented):
on another thought, right now we only have bindings for utils but soon we'll need publish the whole state-transition-z to typescript. In that case we don't need `stdx/core` module anymore? lodestar will just need the whole `state-transition` module

PR #66, opened 2025-10-14
chore: refactor spec test runner

wemeetagain:
- Consolidate logic for the operations spec runner
- Remove external test logger (tests can be more logged via `--summary all`)
- Properly handle bls setting (eg by reading `meta.yaml`)
- Properly handle preset (`-Dpreset=minimal`) (there are currently bugs, the minimal preset needs work)
- Further runners can be added using a similar format:
  - Add a new file/struct in `test/spec/runner` that handles _running_ the runner type. It also needs `Handler`and `handlers` decls. `Handler` must be an enum with a `suiteName` fn.
  - Add a new file/function in `test/spec/writer` that handles _writing_ the tests for that runner type
  - Add the runner to the supported runners in `test/spec/write_spec_tests.zig` and update `writeTests`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -119,7 +119,7 @@ pub fn build(b: *std.Build) void {
     const test_hex = b.addTest(.{
         .name = "hex",
         .root_module = module_hex,
-        .filters = &[_][]const u8{},
+        .filters = b.option([][]const u8, "hex.filters", "hex test filters") orelse &[_][]const u8{},
     });
     const install_test_hex = b.addInstallArtifact(test_hex, .{});
     const tls_install_test_hex = b.step("build-test:hex", "Install the hex test");
@@ -133,7 +133,7 @@ pub fn build(b: *std.Build) void {
     const test_constants = b.addTest(.{
         .name = "constants",
         .root_module = module_constants,
-        .filters = &[_][]const u8{},
+        .filters = b.option([][]const u8, "constants.filters", "constants test filters") orelse &[_][]const u8{},
     });
     const install_test_constants = b.addInstallArtifact(test_constants, .{});
     const tls_install_test_constants = b.step("build-test:constants", "Install the constants test");
@@ -147,7 +147,7 @@ pub fn build(b: *std.Build) void {
     const test_config = b.addTest(.{
         .name = "config",
         .root_module = module_config,
-        .filters = &[_][]const u8{},
+        .filters = b.option([][]const u8, "config.filters", "config test filters") orelse &[_][]const u8{},
     });
     const install_test_config = b.addInstallArtifact(test_config, .{});
     const tls_install_test_config = b.step("build-test:config", "Install the config test");
@@ -161,7 +161,7 @@ pub fn build(b: *std.Build) void {
     const test_consensus_types = b.addTest(.{
         .name = "consensus_types",
         .root_module = module_consensus_types,
-        .filters = &[_][]const u8{},
+        .filters = b.option([][]const u8, "consensus_types.filters", "consensus_types test filters") orelse &[_][]const u8{},
     });
     const install_test_consensus_types = b.addInstallArtifact(test_consensus_types, .{});
     const tls_install_test_consensus_types = b.step("build-test:consensus_types", "Install the consensus_types test");
@@ -175,7 +175,7 @@ pub fn build(b: *std.Build) void {
     const test_preset = b.addTest(.{
         .name = "preset",
         .root_module = module_preset,
-        .filters = &[_][]const u8{},
+        .filters = b.option([][]const u8, "preset.filters", "preset test filters") orelse &[_][]const u8{},
     });
     const install_test_preset = b.addInstallArtifact(test_preset, .{});
     const tls_install_test_preset = b.step("build-test:preset", "Install the preset test");
@@ -189,7 +189,7 @@ pub fn build(b: *std.Build) void {
     const test_state_transition = b.addTest(.{
         .name = "state_transition",
         .root_module = module_state_transition,
-        .filters = &[_][]const u8{},
+        .filters = b.option([][]const u8, "state_transition.filters", "state_transition test filters") orelse &[_][]const u8{},
     });
     const install_test_state_transition = b.addInstallArtifact(test_state_transition, .{});
     const tls_install_test_state_transition = b.step("build-test:state_transition", "Install the state_transition test");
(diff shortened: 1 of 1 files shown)


test/spec/writer/operations.zig
```diff
@@ -0,0 +1,96 @@
...
+    \\    var tc = try Operation.TestCase(.{s}, .{s}, {}).init(allocator, test_dir);
+    \\    defer tc.deinit();
+    \\
+    \\    try tc.runTest();
+    \\}}
+    \\
+    \\
+;
+
+pub fn writeOperationsTests(comptime forks: []const ForkSeq, writer: std.io.AnyWriter) !void {
```
ensi321: I think this file reading logic will be reused so many times for other runners. 
Better write a generic version of it.

test/spec/test_type/operations.zig
```diff
@@ -0,0 +1,246 @@
...
+            .deposit => "deposit",
+            .deposit_request => "deposit_request",
+            .execution_payload => "body",
+            .proposer_slashing => "proposer_slashing",
+            .sync_aggregate => "sync_aggregate",
+            .voluntary_exit => "voluntary_exit",
+            .withdrawal_request => "withdrawal_request",
+            .withdrawals => "execution_payload",
+        };
+    }
```
spiral-ladder: can we just use `@tagName` instead?

test/spec/test_type/operations.zig
```diff
@@ -0,0 +1,246 @@
...
+            pub fn runTest(self: *Self) !void {
+                if (valid) {
+                    try self.process();
+                    try expectEqualBeaconStates(self.post, self.pre.cached_state.state.*);
+                } else {
+                    self.process() catch |err| {
+                        if (err == error.NotImplemented) {
+                            return err;
+                        }
+                        return;
```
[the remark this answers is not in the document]
wemeetagain: We can if thats what makes sense. Alternatively we can throw an `error.SkipZigTest`

src/state_transition/test_utils/generate_state.zig
```diff
@@ -15,6 +17,7 @@ const PubkeyIndexMap = state_transition.PubkeyIndexMap(ValidatorIndex);
 const Index2PubkeyCache = state_transition.Index2PubkeyCache;
 const syncPubkeys = state_transition.syncPubkeys;
 const interopPubkeysCached = @import("./interop_pubkeys.zig").interopPubkeysCached;
+const active_chain_config = if (preset.preset == Preset.mainnet) mainnet_chain_config else minimal_chain_config;
```
spiral-ladder: we're already depending on `build_options`, can we just do this maybe?

```suggestion
const preset_str = @import("build_options").preset;

const active_chain_config = if (std.mem.eql(u8, preset_str, "mainnet")) mainnet_chain_config else minimal_chain_config;
```

if i'm not wrong, the `preset` field is only used to check this, so we can just get rid of that if we do this possibly

test/spec/writer/operations.zig
wemeetagain: ok take a look now (at `write_spec_tests.zig`). More of a framework for handling different runners and handlers and a unified directory walker there.

spiral-ladder (approved):
LGTM

test/spec/runner/Operations.zig
```diff
@@ -0,0 +1,235 @@
...
+        return switch (self) {
+            .block_header => "block",
+            .bls_to_execution_change => "address_change",
+            .execution_payload => "body",
+            .withdrawals => "execution_payload",
+            else => @tagName(self),
+        };
+    }
+
+    pub fn operationObject(self: Operation) []const u8 {
```
ensi321: Would it be more appropriate to call it container instead of object?

test/spec/runner/Operations.zig
```diff
@@ -0,0 +1,235 @@
...
+                    try state_transition.processConsolidationRequest(self.pre.allocator, self.pre.cached_state, &self.op);
+                },
+                .deposit => {
+                    try state_transition.processDeposit(self.pre.allocator, self.pre.cached_state, &self.op);
+                },
+                .deposit_request => {
+                    try state_transition.processDepositRequest(self.pre.allocator, self.pre.cached_state, &self.op);
+                },
+                .execution_payload => {
+                    return error.SkipZigTest;
```
ensi321: I think we can call `processExecutionPayload`. What's the reason of skipping it?

test/spec/runner/Operations.zig
```diff
@@ -0,0 +1,235 @@
...
+        pub fn runTest(self: *Self) !void {
+            if (valid) {
+                try self.process();
+                try expectEqualBeaconStates(self.post, self.pre.cached_state.state.*);
+            } else {
+                self.process() catch |err| {
+                    if (err == error.SkipZigTest) {
+                        return err;
+                    }
+                    return;
```
ensi321: Hmmm, when a test is invalid, we expect `process()` to throw an error. But it could be an issue that it is throwing an error that is different from what we are expecting, eg. we expect `attester_slashing / invalid_att1_empty_indices` to throw error because of empty indices and not because of invalid signature.

We need to provide a mechanism to debug (maybe a simple log is fine?). Here we're just naively returning and swallow that error.

test/spec/test_case.zig
```diff
@@ -12,68 +12,46 @@ const capella = consensus_types.capella;
...
 
-    while (try it.next()) |entry| {
-        // Ignore all non-file
-        if (entry.kind != .file) continue;
-        const name = entry.name;
-        // Ignore all hidden files
-        if (name.len == 0) continue;
-        if (name[0] == '.') continue;
+pub fn loadBlsSetting(allocator: std.mem.Allocator, dir: std.fs.Dir) BlsSetting {
+    var file = dir.openFile("meta.yaml", .{}) catch return .default;
```
ensi321: Would be a good idea to add `meta.yaml` with `bls_setting: 0` and let others change the value as they see fit, instead of having them creating that file

zbuild.zon
```diff
@@ -117,6 +118,7 @@
                     .blst,
                 },
             },
+            .filters = .{},
```
ensi321: Best to create an GitHub issue since the filtering mechanism is not yet implemented.

test/spec/README.md
```diff
@@ -0,0 +1,47 @@
+# Ethereum Consensus Spec Test Framework
+
+This framework generates and runs Zig-based tests for Ethereum consensus specifications, based on the official test formats from [ethereum/consensus-specs](https://github.com/ethereum/consensus-specs/tree/master/tests/formats).
+
+## Overview
+
+The framework automates the creation of test files for various forks and test runners, ensuring compliance with spec tests. It uses Zig to load, process, and validate test cases from SSZ-encoded data.
```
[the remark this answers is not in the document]
wemeetagain: yea this was AI generated and only lightly tweaked

test/spec/runner/Operations.zig
wemeetagain: Another idea, we can separate the test setup from the actual test process by throwing some sort of `error.TestSetup` error for setup steps (eg any allocation that happens beforehand)

wemeetagain: Lets save this for another PR

Issue #72, opened 2025-10-21
Implement getBlockSignatureSets


twoeths:
- this is to be used in beacon node at range sync
- reference implementation: https://github.com/ChainSafe/lodestar/blob/fd1dac853d059ce250d706932daa55e77c33310a/packages/state-transition/src/signatureSets/index.ts#L26

PR #58, opened 2025-10-09
fix: add Block functions

twoeths:
**Motivation**
- we miss `Block` functions, that causes us to use `SignedBlock` everywhere
- this is also not consistent to lodestar
- it'll likely a blocker for spec test

**Description**
- move functions from `SignedBlock` to `Block`, ensure the functions have the same signatures to the current lodestar
- no pointer on wrapped enums because we already have pointers in embedded enumbs
  - instead of
 ```zig
pub const SignedBlock = union(enum) {
    regular: *const SignedBeaconBlock,
    blinded: *const SignedBlindedBeaconBlock,
}
```
  - this PR changes to
```zig
pub const SignedBlock = union(enum) {
    regular: SignedBeaconBlock,
    blinded: SignedBlindedBeaconBlock,
}
```
because the last level already includes pointers
```zig
pub const SignedBeaconBlock = union(enum) {
    phase0: *const ssz.phase0.SignedBeaconBlock.Type,
    altair: *const ssz.altair.SignedBeaconBlock.Type,
    bellatrix: *const ssz.bellatrix.SignedBeaconBlock.Type,
    capella: *const ssz.capella.SignedBeaconBlock.Type,
    deneb: *const ssz.deneb.SignedBeaconBlock.Type,
    electra: *const ssz.electra.SignedBeaconBlock.Type,
```

- no need pointer in wrapped enum types, ie use `Block` instead of `*const Block` in function signatures because we already have pointers inside the last level
```zig
pub const BeaconBlock = union(enum) {
    phase0: *const ssz.phase0.BeaconBlock.Type,
    altair: *const ssz.altair.BeaconBlock.Type,
    bellatrix: *const ssz.bellatrix.BeaconBlock.Type,
    capella: *const ssz.capella.BeaconBlock.Type,
    deneb: *const ssz.deneb.BeaconBlock.Type,
    electra: *const ssz.electra.BeaconBlock.Type,
```

Closes #50

PR #73, opened 2025-10-21
fix: processSyncAggregate

### LINKED ISSUE #51: Spec test processSyncAggregate
Currently lodestar has Block as a param in order to extract signature from there https://github.com/ChainSafe/lodestar/blob/fd1dac853d059ce250d706932daa55e77c33310a/packages/state-transition/src/block/processSyncCommittee.ts#L15

this is not the same to the spec. Need to figure out how the current lodestar implement spec tests for it and do the same thing for state-transition-z

more on that: https://github.com/ChainSafe/state-transition-z/pull/17#discussion_r2350233969

twoeths:
**Motivation**
- make `processSyncAggregate()` signatures the same to the spec so that we can do spec tests

**Description**
- implement `processSyncAggregate()` following the spec, do not call `getSyncCommitteeSignatureSet()` as it's supposed for issue #72
- add valid int test case

Close #51

**TODOs**
- [ ] blocked by #58, need to change the base branch once that's merged

### DIFF
--- a/test/spec/runner/Operations.zig
+++ b/test/spec/runner/Operations.zig
@@ -189,7 +189,7 @@ pub fn TestCase(comptime fork: ForkSeq, comptime operation: Operation, comptime 
                     try state_transition.processProposerSlashing(self.pre.cached_state, &self.op, verify);
                 },
                 .sync_aggregate => {
-                    return error.SkipZigTest;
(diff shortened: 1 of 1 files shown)


twoeths:
closing in favor of #95

PR #75, opened 2025-10-24
fix: operation and sanity spec tests v1.5.0

twoeths:
**Motivation**
- fix operation and sanity spec tests

**Description**
- merge chain config for spec test
- `ReusedEpochTransitionCache` is (implicitly) inited and deinited once
- fix various memory leak issues
- track spec tests in a new CI job

2 remaining skipped tests are:
- operation `processBlockHeader` test, will be fixed in #58
- operation `sync_aggregate`, will fixed in #73

part of #21

### DIFF
--- a/src/state_transition/block/process_attestation_phase0.zig
+++ b/src/state_transition/block/process_attestation_phase0.zig
@@ -66,7 +66,6 @@ pub fn processAttestationPhase0(allocator: Allocator, cached_state: *CachedBeaco
 
 /// AT could be either Phase0Attestation or ElectraAttestation
 pub fn validateAttestation(comptime AT: type, cached_state: *const CachedBeaconStateAllForks, attestation: *const AT) !void {
-    const is_electra = AT == ElectraAttestation;
     const epoch_cache = cached_state.getEpochCache();
     const state = cached_state.state;
     const state_slot = state.slot();
@@ -88,7 +87,7 @@ pub fn validateAttestation(comptime AT: type, cached_state: *const CachedBeaconS
     }
 
     // same to fork >= ForkSeq.electra but more type safe
-    if (is_electra) {
+    if (AT == ElectraAttestation) {
         if (data.index != 0) {
             return error.InvalidAttestationNonZeroDataIndex;
         }


src/state_transition/epoch/process_pending_consolidations.zig
```diff
@@ -52,7 +52,11 @@ pub fn processPendingConsolidations(allocator: Allocator, cached_state: *CachedB
 
     if (next_pending_consolidation > 0) {
         const new_pending_consolidations = pending_consolidations.items[next_pending_consolidation..];
+        // cannot use memcpy due to overlap
```
spiral-ladder: can we use `std.mem.copyForwards`?

src/state_transition/block/process_attestation_phase0.zig
```diff
@@ -3,6 +3,7 @@ const Allocator = std.mem.Allocator;
 const CachedBeaconStateAllForks = @import("../cache/state_cache.zig").CachedBeaconStateAllForks;
 const BeaconStateAllForks = @import("../types/beacon_state.zig").BeaconStateAllForks;
 const ssz = @import("consensus_types");
+const s = @import("ssz");
```
spiral-ladder: I think in near future we should consider what we want to do with namespaces, right now we're using `ssz` for consensus types but then we run into situations like this where we can't use `ssz` the library.

Should we just use `ct` for `consensus_types` and `ssz` to mean `ssz` the library?

src/state_transition/epoch/process_pending_consolidations.zig
spiral-ladder: [7f4574c](https://github.com/ChainSafe/state-transition-z/pull/75/commits/7f4574c54dafadc2df9b30b98841ccd90e8ba47c) addressed here

src/state_transition/block/process_deposit.zig
```diff
@@ -175,7 +175,9 @@ pub fn addValidatorToRegistry(
     try balances.append(allocator, amount);
 }
 
-pub fn isValidDepositSignature(config: *const BeaconConfig, pubkey: BLSPubkey, withdrawal_credential: WithdrawalCredentials, amount: u64, deposit_signature: BLSSignature) !bool {
+/// refer to https://github.com/ethereum/consensus-specs/blob/v1.5.0/specs/electra/beacon-chain.md#new-is_valid_deposit_signature
+/// no need to return error union since consumer does not care about the reason of failure
```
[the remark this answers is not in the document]
wemeetagain: Yes, I think this is a similar thought to https://github.com/ChainSafe/lodestar/issues/6330

Essentially, we may return an `Error!void` and only if absolutely necessary, wrap to convert to `bool` for purposes of spec compliance.

src/config/beacon_config.zig
```diff
@@ -213,7 +213,7 @@ pub const BeaconConfig = struct {
     }
 
     pub fn getDomainForVoluntaryExit(self: *const BeaconConfig, state_slot: Slot, message_slot: ?Slot) ![32]u8 {
-        const domain = if (state_slot < self.chain.DENEB_FORK_EPOCH * preset.SLOTS_PER_EPOCH) {
+        const domain = if (state_slot / preset.SLOTS_PER_EPOCH < self.chain.DENEB_FORK_EPOCH) {
```
ensi321: nit: probably better do @divFloor here

src/state_transition/block/process_attestation_phase0.zig
ensi321: Prefer to reserve `ssz` for ssz library, and something else for consensus_types

src/state_transition/block/process_withdrawals.zig
```diff
@@ -53,14 +67,15 @@ pub fn processWithdrawals(
     }
 
     // Update the nextWithdrawalValidatorIndex
+    const nextWithdrawalValidatorIndex = state.nextWithdrawalValidatorIndex();
```
ensi321:
```suggestion
    const new_withdrawal_validator_index = state.nextWithdrawalValidatorIndex();
```

ensi321 (commented):
looks good. minor comments

src/state_transition/block/process_attestation_phase0.zig
twoeths: tracked in https://github.com/ChainSafe/lodestar-z/issues/82

spiral-ladder (approved):

PR #80, opened 2025-11-05
feat: impl merkle_proof spec test

GrapeBaBa:
**Motivation**

Support merkle_proof spec tests follow the consensus-spec-tests.

**Description**

- Introduces `merkle_proof` runner and writer follow the current design .
- Expands the spec test generator to support the new `merkle_proof` runner: registers the handler, ensures output directories exist, iterates nested suite/case folders, and emits generated cases via the new writer module.


GrapeBaBa:
close as #81

PR #90, opened 2025-11-13
fix: add process_block_header operations spec test

### LINKED ISSUE #50: Enhance Block enum
the current Block definition [here](https://github.com/ChainSafe/state-transition-z/blob/b0fd005820fde96c696bc2d47eef7e56a2f82e45/src/state_transition/types/signed_block.zig#L2) does not have any methods inside, that's a limitation

I think `SignedBlock` should delegate methods to `Block` so that we can use both data structures in any places

_Originally posted by @twoeths in https://github.com/ChainSafe/state-transition-z/pull/17#discussion_r2393400505_

twoeths:
**Motivation**
- the `processBlockHeader()` fn has incorrect signature, it accepts a `SignedBlock` while only a `Block` is needed according to the spec, hence it's being skipped in `operations` test

**Description**
- add missing methods to `Block`
- enable spec tests for it

Closes #50


src/state_transition/types/signed_block.zig
spiral-ladder: For consistency with `beacon_block.zig`, which contains the signed and blinded variants of the `BeaconBlock` type, maybe we should name this `block.zig` as well.

spiral-ladder (approved):

Issue #97, opened 2025-11-15
Implement CompatibleUnion ssz type


twoeths:
refer to https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.2/ssz/simple-serialize.md

PR #99, opened 2025-11-16
feat(ssz): add Progressive List, Progressive BitList, ProgressiveContainer, CompatibleUnion 

### LINKED ISSUE #98: Implement ssz progressive types
refer to https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.2/ssz/simple-serialize.md

guha-rahul:
fixes #98 and #97


src/ssz/type/bit_list.zig
```diff
@@ -314,7 +314,7 @@ pub fn BitListType(comptime _limit: comptime_int) type {
 
             const last_byte_clz = @clz(last_byte);
             if (last_byte_clz == 8) {
-                return error.noPaddingBit;
+                return error.InvalidSSZ;
```
wemeetagain: Why are these errors changed?

src/ssz/type/container.zig
```diff
@@ -255,12 +255,19 @@ pub fn VariableContainerType(comptime ST: type) type {
         _offsets[i] = _fixed_end;
         if (comptime isFixedType(field.type)) {
             _min_size += field.type.fixed_size;
-            _max_size += field.type.fixed_size;
+            if (_max_size != std.math.maxInt(usize)) {
+                _max_size += field.type.fixed_size;
+            }
```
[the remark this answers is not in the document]
guha-rahul: I was experiencing an overflow so i added this guard , it seems like `ProgressiveBitListType` and `ProgressiveListType` both have max_size = std.math.maxInt(usize) so an unbounded field followed by fixed
  fields was causing overflows.

wemeetagain:
@lodekeeper-z can you review this implementation against the one in https://github.com/chainsafe/ssz ?

src/ssz/type/progressive_bit_list.zig
```diff
@@ -0,0 +1,385 @@
...
+
+        pub fn resize(self: *@This(), allocator: std.mem.Allocator, bit_len: usize) !void {
+            const old_byte_len = std.math.divCeil(usize, self.bit_len, 8) catch unreachable;
+            const byte_len = std.math.divCeil(usize, bit_len, 8) catch unreachable;
+            try self.data.resize(allocator, byte_len);
+            self.bit_len = bit_len;
+            // zero out additionally allocated bytes
+            if (old_byte_len < byte_len) {
+                @memset(self.data.items[old_byte_len..], 0);
+            }
```
lodekeeper-z, line 75: **Correctness:** shrinking within the same byte leaves bits above the new `bit_len` intact. Those stale bits become indistinguishable from the SSZ delimiter during serialization. A focused repro creates 8 bits, sets bit 7, resizes to 1, then serializes/deserializes; the round-trip length is 7 instead of 1 (`expected 1, found 7`). Please mask the retained final byte when shrinking (and add that round-trip regression).

src/ssz/type/compatible_union.zig
```diff
@@ -0,0 +1,599 @@
...
+
+        // Check for duplicate selectors
+        for (options, 0..) |option1, i| {
+            for (options, 0..) |option2, j| {
+                if (i < j and option1.@"0" == option2.@"0") {
+                    @compileError("CompatibleUnion has duplicate selector");
+                }
+            }
+        }
+    }
```
lodekeeper-z, line 157: **Semantic parity:** this validates selector shape/uniqueness but never validates that every option pair has compatible Merkleization. ChainSafe/ssz does a pairwise `areTypesCompatible` check in the constructor and rejects incompatible unions. The test below in this file currently constructs `Square {side,color}` and `Circle {radius,color}` as one compatible union even though the reference rejects containers with different field names at the same positions. That defeats the defining invariant of `CompatibleUnion`; please port the recursive compatibility check and add compile-time rejection coverage for incompatible options.

lodekeeper-z (dismissed):
Compared the implementation at `8171ec1` against ChainSafe/ssz `5681f33` (progressive types introduced in `bcbcae3`). I found two correctness gaps that should be fixed before merge. The existing Lodestar-Z SSZ suite passes, but a focused public-API shrink/serialize/deserialize regression fails with `expected 1, found 7`.

lodekeeper-z:
## Code Review Summary

**Verdict: Changes requested** (2 correctness issues)

- `ProgressiveBitList.resize` preserves truncated high bits, so serialization can turn stale data into the delimiter and change the round-trip length. Focused regression: expected 1 bit, got 7.
- `CompatibleUnionType` omits the reference implementation’s pairwise type-compatibility validation, allowing unions whose options do not have compatible Merkleization.

Verification: existing `zig build test:ssz` passes at `8171ec1`; the focused shrink regression fails as described. Compared against ChainSafe/ssz `5681f33` (feature origin `bcbcae3`).

Formal review: https://github.com/ChainSafe/lodestar-z/pull/99#pullrequestreview-5011595864

wemeetagain:
@lodekeeper-z please fix the issues you found

lodekeeper-z:
Fixed both findings in signed commit `d091750b`. Direct push to the contributor-owned head was denied, so I opened https://github.com/guha-rahul/state-transition-z/pull/1 against `Progressive_types`.

The fix clears stale high bits on ProgressiveBitList shrink and ports recursive pairwise CompatibleUnion type validation, including container field names and progressive-container positions.

Verified locally: `zig build test:ssz`, minimal generic SSZ spec tests, and `pnpm lint` (10 existing warnings).

src/ssz/type/type_kind.zig
```diff
@@ -18,9 +22,9 @@ pub fn isBasicType(T: type) bool {
 pub fn isFixedType(T: type) bool {
     return switch (T.kind) {
         .uint, .bool => true,
-        .list => false,
+        .list, .progressive_list, .progressive_bit_list, .compatible_union => false,
         .vector => isFixedType(T.Element),
-        .container => {
+        .container, .progressive_container => {
```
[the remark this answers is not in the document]
wemeetagain: ~~@lodekeeper-z fix this~~ nevermind, looks like this was wrong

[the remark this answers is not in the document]
lodekeeper-z: I checked this against the current ChainSafe/ssz implementation before changing it. This thread is outdated: ProgressiveContainer serialization is identical to Container serialization, so it is fixed-size when every active field is fixed-size and variable-size when any field is variable-size. The reference computes `fixedSize` that way, and the current Zig `isFixedType` matches it. You also corrected the original premise here: https://github.com/ChainSafe/lodestar-z/pull/99#issuecomment-3649514820. Changing this to always return false would regress `FixedProgressiveContainerType`, so I did not make that incorrect change. The focused ProgressiveContainer tests pass with `zig build test:ssz -Dssz.filters=ProgressiveContainerType`.

wemeetagain:
@lodekeeper-z can you ensure that this has proper atomic behavior of each operation and failure of each operation?
Eg like in https://github.com/ChainSafe/lodestar-z/pull/567 and https://github.com/ChainSafe/lodestar-z/pull/574

lodekeeper-z:
Audited the new progressive SSZ operations against the failure-atomic publication/cleanup patterns from #567 and #574, then opened https://github.com/guha-rahul/state-transition-z/pull/2 against this PR's head branch.

The follow-up stages clone/decode/JSON/tree outputs before publication, cleans partial values on failure, reclaims unpublished pooled nodes, and adds deterministic OOM/malformed-input regressions across ProgressiveList, ProgressiveBitList, ProgressiveContainer, and CompatibleUnion.

Verified locally at signed commit `e24fe512`: `zig build test:ssz` passes all 265 tests with no leaks. The follow-up PR is mergeable; its title check is currently running.

GrapeBaBa:
Need a bit more time to review this PR

src/ssz/type/compatible_union.zig
```diff
@@ -0,0 +1,683 @@
...
+
+    return struct {
+        pub const kind = TypeKind.compatible_union;
+        pub const Type: type = ValueType;
+        pub const min_size: usize = _min_size;
+        pub const max_size: usize = _max_size;
+        pub const _union_options = options;
+
+        // Default value uses first option with selector
+        pub const default_value: Type = blk: {
```
GrapeBaBa: This seems not match the spec https://github.com/ethereum/ssz-specs/blob/main/src/ssz/union.py#L261-L263

lodekeeper-z: You are right. The spec explicitly says CompatibleUnion has no default because selector 0 is reserved. I removed the first-option default and added a regression in https://github.com/guha-rahul/state-transition-z/pull/3. The full `zig build test:ssz` suite passes.

wemeetagain:
@lodekeeper-z please resolve the merge conflicts as a commit

lodekeeper-z:
Resolved the current `main` conflict in signed merge commit `097c738b`. Direct push to the contributor-owned head was denied, so I opened https://github.com/guha-rahul/state-transition-z/pull/4 against `Progressive_types`.

The resolution preserves both sets of additions in `src/ssz/memory_safety_test.zig`. Verified with `zig build test:ssz`, tracked Zig formatting, and `pnpm lint` (10 existing warnings). The follow-up PR is mergeable and its title check passes.

src/ssz/type/tree_api.zig
wemeetagain, line 1: @lodekeeper-z please remove this file

lodekeeper-z: Removed `src/ssz/type/tree_api.zig` and its public exports in signed commit `8fa1c0c7`. I opened https://github.com/guha-rahul/state-transition-z/pull/5 against `Progressive_types`. The allocator dispatch now stays local to its production callers and spec-test harness. Verified with `zig build test:ssz`, tracked Zig formatting, and `pnpm lint` (10 existing warnings).

test/spec/ssz/test_case.zig
```diff
@@ -7,16 +7,91 @@ const Node = @import("persistent_merkle_tree").Node;
...
+    pool: *Node.Pool,
+    data: []const u8,
+) !Node.Id {
+    return switch (ST.kind) {
+        .progressive_list, .progressive_bit_list, .compatible_union => ST.tree.deserializeFromBytes(allocator, pool, data),
+        .progressive_container => if (comptime ssz.isFixedType(ST))
+            ST.tree.deserializeFromBytes(pool, data)
+        else
+            ST.tree.deserializeFromBytes(allocator, pool, data),
+        else => ST.tree.deserializeFromBytes(pool, data),
```
wemeetagain: @lodekeeper-z this code still exists because it seems that there's an inconsistency between how progressive types and non-progressives are implemented?

I would expect that if this feature is implemented correctly, this shouldn't be needed.
Eg: the implementation should align with the existing framework standards.

lodekeeper-z: You are right. The helper code was compensating for progressive tree APIs that took a separate allocator and lacked the standard tree serialization surface. I aligned them with the existing pool-based API and removed the kind switches in https://github.com/guha-rahul/state-transition-z/pull/6. `zig build test:ssz` and the minimal generic SSZ spec suite pass locally.

src/ssz/hasher.zig
```diff
@@ -151,6 +168,10 @@ pub const HasherData = struct {
         var chunks = self.chunks;
         chunks.deinit(allocator);
     }
+
+    pub fn getAllocator(self: *HasherData) std.mem.Allocator {
```
wemeetagain: @lodekeeper-z This method is unnecessary. Just use field access.

src/ssz/type/type_kind.zig
```diff
@@ -39,3 +43,13 @@ pub fn isFixedType(T: type) bool {
...
+
+// Progressive list types
+pub fn isProgressiveListType(T: type) bool {
+    return T.kind == .progressive_list;
+}
+
+// Compatible union types
+pub fn isCompatibleUnionType(T: type) bool {
+    return T.kind == .compatible_union;
+}
```
wemeetagain: @lodekeeper-z These functions are unnecessary. Idiomatic usage would be to use T.kind field access.

src/ssz/hasher.zig
lodekeeper-z: Removed `HasherData.getAllocator()` and switched the three callers to direct `scratch.allocator` access in https://github.com/guha-rahul/state-transition-z/pull/7. `zig build test:ssz` passes all 298 tests.

src/ssz/type/type_kind.zig
lodekeeper-z: Removed the redundant `isProgressiveListType` and `isCompatibleUnionType` helpers and their exports in https://github.com/guha-rahul/state-transition-z/pull/7. Added a regression that keeps these helper APIs absent; `zig build test:ssz` passes all 298 tests.

src/ssz/root.zig
```diff
@@ -50,6 +65,12 @@ pub const ListBasicTreeView = tree_view.ListBasicTreeView;
 pub const ListCompositeTreeView = tree_view.ListCompositeTreeView;
 pub const CloneOpts = @import("tree_view/utils/clone_opts.zig").CloneOpts;
 
+test "redundant SSZ helper APIs are not exposed" {
+    try testing.expect(!@hasDecl(types, "isProgressiveListType"));
+    try testing.expect(!@hasDecl(types, "isCompatibleUnionType"));
+    try testing.expect(!@hasDecl(HasherData, "getAllocator"));
+}
+
```
wemeetagain, line 73: @lodekeeper-z this is useless. remove this.

src/ssz/type/progressive_bit_list.zig
```diff
@@ -0,0 +1,446 @@
...
+const BoolType = @import("bool.zig").BoolType;
+const hexToBytes = @import("hex").hexToBytes;
+const bytesToHex = @import("hex").bytesToHex;
+const hexByteLen = @import("hex").hexByteLen;
+const hexLenFromBytes = @import("hex").hexLenFromBytes;
+const mixInLength = @import("hashing").mixInLength;
+const Node = @import("persistent_merkle_tree").Node;
+const progressive = @import("progressive.zig");
+
+pub fn ProgressiveBitList() type {
```
wemeetagain, line 14: @lodekeeper-z make this a pub const. It doesn't need to be a fn, typically we only do this when we need additional comptime variables to affect the struct.

src/ssz/root.zig
lodekeeper-z: Removed the redundant API-presence test in signed commit `8cc4e06b`. Direct push to the contributor-owned head was denied, so I opened https://github.com/guha-rahul/state-transition-z/pull/8. `zig build test:ssz` passes.

src/ssz/type/progressive_bit_list.zig
lodekeeper-z: Changed `ProgressiveBitList` from a zero-argument type factory to a direct `pub const` value type in signed commit `8cc4e06b`, with focused direct-use coverage. The follow-up is https://github.com/guha-rahul/state-transition-z/pull/8; `zig build test:ssz` passes.

wemeetagain (approved):
This looks good.
Thanks for sticking with this for the many many rounds of review.
Merging this to unblock other work.

PR #100, opened 2025-11-16
feat: impl `getAll` / `push` / `sliceTo` / `sliceFrom` API for list tree view

GrapeBaBa:
**Motivation**

Align Zig `TreeView` list functionality with the richer SSZ ListTreeView APIs already exercised on the TypeScript side so downstream consumers in Lodestar can rely on consistent `getAll`, `push`, `sliceTo`, and `sliceFrom` semantics during the upcoming list-view work.

**Description**

Refactored `tree_view.zig` to add cached-length updating `push`, inclusive `sliceTo`, and suffix `sliceFrom` support for list views while keeping basic and composite arrays in sync. Added comprehensive integration coverage under `tree_view.zig`.

### DIFF
--- a/src/ssz/tree_view/list_basic.zig
+++ b/src/ssz/tree_view/list_basic.zig
@@ -101,6 +101,7 @@ pub fn ListBasicTreeView(comptime ST: type) type {
         }
 
         /// Return a new view containing all elements up to and including `index`.
+        /// Caller must call `deinit()` on the returned view to avoid memory leaks.
         pub fn sliceTo(self: *Self, index: usize) !Self {
             try self.commit();
 
--- a/src/ssz/tree_view/list_composite.zig
+++ b/src/ssz/tree_view/list_composite.zig
@@ -92,6 +92,11 @@ pub fn ListCompositeTreeView(comptime ST: type) type {
             return error.NotImplemented;
         }
 
+        /// Appends an element to the end of the list.
+        ///
+        /// Ownership of the `value` TreeView is transferred to the list view.
+        /// The caller must not deinitialize or otherwise use `value` after calling this method,
+        /// as it is now owned by the list.
         pub fn push(self: *Self, value: Element) !void {
             const list_length = try self.length();
             if (list_length >= ST.limit) {
@@ -103,6 +108,7 @@ pub fn ListCompositeTreeView(comptime ST: type) type {
         }
 
         /// Return a new view containing all elements up to and including `index`.
+        /// The caller **must** call `deinit()` on the returned view to avoid memory leaks.
         pub fn sliceTo(self: *Self, index: usize) !Self {
             try self.commit();
 
@@ -130,6 +136,7 @@ pub fn ListCompositeTreeView(comptime ST: type) type {
         }
 
         /// Return a new view containing all elements from `index` to the end.
+        /// The returned view must be deinitialized by the caller using `deinit()` to avoid memory leaks.
         pub fn sliceFrom(self: *Self, index: usize) !Self {
             try self.commit();
 


src/ssz/tree_view.zig
```diff
@@ -104,6 +130,22 @@ pub fn TreeView(comptime ST: type) type {
             out.* = self.data.root.getRoot(self.pool).*;
         }
 
+        /// Get the length of the array. For Vector returns compile-time constant, for List reads from tree.
+        pub fn getLength(self: *Self) !usize {
```
twoeths: make this compatible to ~~`setListLength()`~~ `updateListLength()` api
```suggestion
        pub fn getListLength(self: *Self) !usize {
```

src/ssz/tree_view.zig
```diff
@@ -125,22 +167,102 @@ pub fn TreeView(comptime ST: type) type {
...
+        inline fn listLengthGindex() Gindex {
+            comptime {
+                if (ST.kind != .list) {
+                    @compileError("listLengthGindex can only be used with List types");
+                }
+            }
+            return Gindex.fromDepth(1, 1);
+        }
+
+        fn updateListLength(self: *Self, new_length: usize) !void {
```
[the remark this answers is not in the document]
twoeths: just found we already have `list_length`, may just use it
i'm not sure if we should lazily compute it through

src/ssz/tree_view/root.zig
```diff
@@ -0,0 +1,205 @@
...
+    changed: std.AutoArrayHashMapUnmanaged(Gindex, void),
+
+    /// Number of chunk nodes (starting from index 0) that have been prefetched into the cache.
+    /// Only used for array/list views with basic element types.
+    prefetched_chunk_count: ?usize,
+
+    /// Cached length for list views. null marks the value as invalid/unpopulated, so
+    /// the next `getLength()` fetches from the tree before returning a concrete length.
+    /// Only used for list views.
+    list_length: ?ListLengthUint,
```
twoeths: I don't think we want to track `list_length` for this common data structure
`list_length` should be tracked at its own list TreeView, `ListCompositeTreeView` for example
in `commit()` of `ListCompositeTreeView` we'll do an extra binding for `list_length`

src/ssz/tree_view/list_composite.zig
```diff
@@ -0,0 +1,208 @@
...
+            self.base_view.data.changed.clearRetainingCapacity();
+            self.base_view.data.list_length = null;
+        }
+
+        pub fn hashTreeRoot(self: *Self, out: *[32]u8) !void {
+            try self.commit();
+            out.* = self.base_view.data.root.getRoot(self.base_view.pool).*;
+        }
+
+        pub fn getLength(self: *Self) !usize {
```
twoeths: this is a bit verbose because we never have `setLength()`
maybe either `length()` or `len()`is enough and more compatible to the regular type
don't have strong opinion here but let's see how @spiral-ladder thinks

src/ssz/tree_view/list_composite.zig
```diff
@@ -0,0 +1,228 @@
...
+
+        pub fn deinit(self: *Self) void {
+            self.base_view.deinit();
+        }
+
+        pub fn commit(self: *Self) !void {
+            try self.base_view.commit();
+
+            if (self.list_length_cache) |cached| {
+                if (!isDirtyLength(cached)) return;
```
twoeths: when we only call `set()` for a list, we don't change the length
but the `self.base_view.data.root` only holds the root of the regular tree without `mix_in_length()`
we always have to bind to length node regardless
  - if dirty length, create new length node
  - if not, use the existing length node

this is same to what we have right now in typescript https://github.com/ChainSafe/ssz/blob/f5ed0b457333749b5c3f49fa5eafa096a725f033/packages/ssz/src/type/arrayBasic.ts#L51

also looks like there is no unit test for this case (with set() only)

src/ssz/tree_view/container.zig
```diff
@@ -0,0 +1,95 @@
...
+            if (comptime isBasicType(ChildST)) {
+                var value: ChildST.Type = undefined;
+                const child_node = try self.base_view.getChildNode(child_gindex);
+                try ChildST.tree.toValue(child_node, self.base_view.pool, &value);
+                return value;
+            } else {
+                const child_data = try self.base_view.getChildData(child_gindex);
+
+                return .{
+                    .base_view = .{
```
twoeths: need to double check all of the places like this because base_view is not everything for the `Field`
see also this comment in [CompositeChunks.get](https://github.com/ChainSafe/lodestar-z/pull/100/changes#r2622423541)

src/ssz/tree_view/list_composite.zig
GrapeBaBa: https://github.com/ChainSafe/lodestar-z/blob/main/src/ssz/type/list.zig#L278 I thought it already include `length node`?

src/ssz/tree_view/list_basic.zig
```diff
@@ -0,0 +1,217 @@
...
+        }
+        if (!@hasDecl(ST, "Element") or !isBasicType(ST.Element)) {
+            @compileError("ListBasicTreeView can only be used with List of basic element types");
+        }
+    }
+
+    return struct {
+        base_view: BaseTreeView,
+        list_length_cache: ?ListLengthUint = null,
+        prefetched_chunk_count: ?usize = null,
```
wemeetagain: I think we should continue with a single `TreeViewData` with no additional fields in the wrapping `TreeView` structs.
For example, I don't think we need either of these additional fields here or in the other `*TreeView` structs.

length is gindex 3, so it can use something like
```zig
pub fn length(self: *Self) !usize {
    const length_node = try self.base_view.getChildNode(@enumFromInt(3));
    const length_chunk = length_node.getRoot(self.base_view.pool);
    return std.mem.readInt(usize, length_chunk[0..@sizeOf(usize)], .little);
}
```

Likewise `prefetched_chunk_count` can be removed and a more naive implementation can be used.

We can bench and go further if needed (eg: using a union for `TreeViewData`), but lets only do that if we determine we need it.

src/ssz/tree_view/list_basic.zig
```diff
@@ -0,0 +1,163 @@
...
+            defer if (length_node) |id| self.base_view.pool.unref(id);
+            const root_with_length = try Node.Id.setNode(new_root.?, self.base_view.pool, @enumFromInt(3), length_node.?);
+            errdefer self.base_view.pool.unref(root_with_length);
+
+            length_node = null;
+            new_root = null;
+
+            return try Self.init(self.base_view.allocator, self.base_view.pool, root_with_length);
+        }
+
```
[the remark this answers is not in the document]
wemeetagain: opened an issue for this

wemeetagain (approved):

PR #119, opened 2025-12-08
feat: enhance clone api

twoeths:
**Motivation**
- when upgrading BeaconState, we have to manually clone() each field of `ExecutionPayloadHeader` from bellatrix -> capella and capella->deneb

**Description**
- enhance clone() api to accept a pointer to super-set type
- this is comptime check so should be safe
- list of types are applied: Container, List, Vector


spiral-ladder (dismissed):
LGTM with nits

spiral-ladder (approved):
committed my nitpicks; LGTM

Issue #102, opened 2025-11-18
Add state transition benhmarks


wemeetagain:
We want to benchmark all of the common state transition mainnet workloads

Can be broken into sub-issues as necessary:
- block transition
- epoch transition
- block / epoch single operations (less important)

guha-rahul:
I opened an sub issue to add the block progressing benchmark.

PR #121, opened 2025-12-09
bench: Add benchmark for block and Epoch Processing

guha-rahul:
tracks #102 
- Adds benchmark for block and epoch transition
- fix a bug in [`getAttestationDataSigningRoot`  ](https://github.com/ChainSafe/lodestar/blob/688d5584ead0c6245007ece33175f6658169b662/packages/state-transition/src/signatureSets/indexedAttestation.ts#L15)
- Unlock spec tests to test for proposer and signatures in state transition which was probably failing because of the above issue

### DIFF
--- a/bench/state_transition/process_block.zig
+++ b/bench/state_transition/process_block.zig
@@ -3,44 +3,6 @@
 //! Uses a mainnet state at slot 13180928 and block at slot 13180929.
 //! Run with: zig build run:bench_process_block -Doptimize=ReleaseFast [-- /path/to/state.ssz /path/to/block.ssz]
 
-// printf "Date: %s\nKernel: %s\nCPU: %s\nCPUs: %s\nMemory: %sGi\n" "$(date)" "$(uname -sr)" "$(sysctl -n machdep.cpu.brand_string)" "$(sysctl -n hw.ncpu)" "$(echo "$(sysctl -n hw.memsize) / 1024 / 1024 / 1024" | bc)"
-// Date: Tue Dec  9 2025
-// Kernel: Darwin 25.1.0
-// CPU: Apple M3
-// CPUs: 8
-// Memory: 16Gi
-//
-// zbuild run bench_process_block -Doptimize=ReleaseFast OR zbuild run bench_process_block -Doptimize=ReleaseFast -- /path/to/state.ssz /path/to/block.ssz
-
-// State: slot=13180929, validators=2156873
-// benchmark              runs     total time     time/run (avg ± σ)     (min ... max)                p75        p99        p995
-// -----------------------------------------------------------------------------------------------------------------------------
-// block_header           50       1.515s         30.312ms ± 9.578ms     (26.488ms ... 94.892ms)      30.18ms    94.892ms   94.892ms
-// withdrawals            50       1.449s         28.99ms ± 2.556ms      (26.232ms ... 39.097ms)      29.447ms   39.097ms   39.097ms
-// execution_payload      50       1.745s         34.917ms ± 11.606ms    (26.696ms ... 93.451ms)      38.481ms   93.451ms   93.451ms
-// randao                 50       1.919s         38.384ms ± 5.864ms     (28.478ms ... 51.428ms)      41.875ms   51.428ms   51.428ms
-// randao_no_sig          50       1.69s          33.814ms ± 13.099ms    (27.682ms ... 104.232ms)     32.55ms    104.232ms  104.232ms
-// eth1_data              50       1.83s          36.601ms ± 33.016ms    (26.611ms ... 261.882ms)     32.772ms   261.882ms  261.882ms
-// operations             50       2.277s         45.545ms ± 6.849ms     (41.418ms ... 90.257ms)      46.806ms   90.257ms   90.257ms
-// operations_no_sig      50       1.535s         30.712ms ± 3.142ms     (27.777ms ... 44.238ms)      30.832ms   44.238ms   44.238ms
-// sync_aggregate         50       1.539s         30.795ms ± 2.939ms     (26.857ms ... 41.007ms)      31.772ms   41.007ms   41.007ms
-// sync_aggregate_no_sig  50       1.649s         32.984ms ± 25.843ms    (26.37ms ... 209.967ms)      30.63ms    209.967ms  209.967ms
-// process_block          50       2.39s          47.811ms ± 6.151ms     (43.82ms ... 73.096ms)       47.603ms   73.096ms   73.096ms
-// process_block_no_sig   50       1.657s         33.152ms ± 5.263ms     (29.143ms ... 60.873ms)      32.916ms   60.873ms   60.873ms
-// block(segments)        50       2.555s         51.112ms ± 8.533ms     (44.404ms ... 95.221ms)      53.978ms   95.221ms   95.221ms
-
-// Segmented block breakdown :
-// step                   runs     total time     time/run (avg)
-// ---------------------------------------------------------------------
-// block_total            50       1.041s         20.828ms
-// block_header           50       53.676ms        1.074ms
-// withdrawals            50       1.191ms        0.024ms
-// execution_payload      50       26.506ms        0.530ms
-// randao                 50       38.248ms        0.765ms
-// eth1_data              50       0.055ms        0.001ms
-// operations             50       852.513ms        17.050ms
-// sync_aggregate         50       69.058ms        1.381ms
-
 const std = @import("std");
 const zbench = @import("zbench");
 const state_transition = @import("state_transition");
--- a/bench/state_transition/process_epoch.zig
+++ b/bench/state_transition/process_epoch.zig
@@ -536,7 +536,6 @@ fn runBenchmark(
         .cached_state = cached_state,
     }, .{});
 
-    // Post-Altair
     if (comptime fork.gte(.altair)) {
         try bench.addParam("inactivity_updates", &ProcessInactivityUpdatesBench{
             .cached_state = cached_state,
@@ -597,7 +596,6 @@ fn runBenchmark(
         }, .{});
     }
 
-    // Post-Fulu
     if (comptime fork.gte(.fulu)) {
         try bench.addParam("proposer_lookahead", &ProcessProposerLookaheadBench{
             .cached_state = cached_state,


bench/state_transition/process_block.zig
```diff
@@ -0,0 +1,351 @@
...
+// process_operations     44       2.401s         54.581ms ± 33.324ms    (41.66ms ... 254.774ms)      51.213ms   254.774ms  254.774ms
+// process_sync_aggregate 67       2.086s         31.145ms ± 10.657ms    (26.877ms ... 111.953ms)     30.532ms   111.953ms  111.953ms
+
+const CachedBeaconStateAllForks = state_transition.CachedBeaconStateAllForks;
+const BeaconStateAllForks = state_transition.BeaconStateAllForks;
+const SignedBlock = state_transition.SignedBlock;
+const SignedBeaconBlock = state_transition.SignedBeaconBlock;
+const Body = state_transition.Body;
+const st_mod = @import("state_transition").state_transition;
+const stateTransition = st_mod.stateTransition;
```
spiral-ladder: we already imported `state_transition` above, so

```suggestion
const stateTransition = state_transition.state_transition.stateTransition;
```

although we should probably do something about this ugly namespacing (in a separate PR)

bench/state_transition/process_epoch.zig
```diff
@@ -0,0 +1,491 @@
...
+            self.allocator.destroy(cloned);
+        }
+        var cache = EpochTransitionCache.init(self.allocator, cloned) catch return;
+        defer {
+            cache.deinit();
+            self.allocator.destroy(cache);
+        }
+        state_transition.processJustificationAndFinalization(cloned, cache) catch return;
+    }
+};
```
spiral-ladder:
```suggestion
const ProcessJustificationAndFinalizationBench = struct {
    cached_state: *CachedBeaconStateAllForks,

    pub fn run(self: ProcessJustificationAndFinalizationBench, allocator: std.mem.Allocator) void {
        const cloned = self.cached_state.clone(allocator) catch return;
        defer {
            cloned.deinit();
            allocator.destroy(cloned);
        }
        var cache = EpochTransitionCache.init(allocator, cloned) catch return;
        defer {
            cache.deinit();
            allocator.destroy(cache);
        }
        state_transition.processJustificationAndFinalization(cloned, cache) catch return;
    }
};
```

We can probably do away with the `allocator` in the struct and just use the passed in `allocator`, same goes for the other benchmark structs.

bench/state_transition/process_epoch.zig
```diff
@@ -0,0 +1,491 @@
...
+    try stdout.print("\nStarting process_epoch benchmarks...\n\n", .{});
+
+    var bench = zbench.Benchmark.init(allocator, .{});
+    defer bench.deinit();
+
+    const process_justification_bench = ProcessJustificationAndFinalizationBench{
+        .allocator = allocator,
+        .cached_state = cached_state,
+    };
+    try bench.addParam("justification_finalization", &process_justification_bench, .{});
```
spiral-ladder: let's try to be accurate with naming.

```suggestion
    const process_justification_and_finalization_bench = ProcessJustificationAndFinalizationBench{
        .allocator = allocator,
        .cached_state = cached_state,
    };
    try bench.addParam("justification_and_finalization", &process_justification_and_finalization_bench, .{});
```

bench/state_transition/process_epoch.zig
```diff
@@ -0,0 +1,491 @@
...
+        .allocator = allocator,
+        .cached_state = cached_state,
+    };
+    try bench.addParam("historical_summaries", &process_historical_bench, .{});
+
+    const process_participation_bench = ProcessParticipationFlagUpdatesBench{
+        .allocator = allocator,
+        .cached_state = cached_state,
+    };
+    try bench.addParam("participation_flags", &process_participation_bench, .{});
```
spiral-ladder:
```suggestion
    const process_participation_flag_updates_bench = ProcessParticipationFlagUpdatesBench{
        .allocator = allocator,
        .cached_state = cached_state,
    };
    try bench.addParam("participation_flag_updates", &process_participation_flag_updates_bench, .{});
```

guha-rahul:
@spiral-ladder another question, i just realised that the benchmark numbers are a bit misleading since even though we are getting huge numbers , the `processBlock` takes only <50ms every time. So i assume its the clone and deinit and other zbench overhead. I did a segmented breakdown for the key functions in `processBlock` , should i keep the original zbench benchmarks and add the segmented breakdown at the bottom ?

bench/state_transition/process_block.zig
```diff
@@ -0,0 +1,463 @@
...
+    try BeaconState.deserializeFromBytes(allocator, state_bytes, state_data);
+    const beacon_state = try allocator.create(BeaconStateAllForks);
+    beacon_state.* = @unionInit(BeaconStateAllForks, @tagName(fork), state_data);
+    return beacon_state;
+}
+
+fn loadBlock(comptime fork: ForkSeq, allocator: std.mem.Allocator, block_bytes: []const u8) !SignedBeaconBlock {
+    const ForkTypes = @field(types, @tagName(fork));
+    const SignedBeaconBlockType = ForkTypes.SignedBeaconBlock;
+    const block_data = try allocator.create(SignedBeaconBlockType.Type);
```
spiral-ladder:
```suggestion
    const SignedBeaconBlock ForkTypes.SignedBeaconBlock;
    const block_data = try allocator.create(SignedBeaconBlock.Type);
```

bench/state_transition/process_epoch.zig
```diff
@@ -0,0 +1,702 @@
...
+    });
+
+    try stdout.print("Cached state created at slot {}\n", .{cached_state.state.slot()});
+    try stdout.print("\nStarting process_epoch benchmarks for {s} fork...\n\n", .{@tagName(fork)});
+
+    var bench = zbench.Benchmark.init(allocator, .{ .iterations = 50 });
+    defer bench.deinit();
+
+    // All forks
+    try bench.addParam("justification_finalization", &ProcessJustificationAndFinalizationBench{
```
spiral-ladder:
```suggestion
    try bench.addParam("justification_finalization", &ProcessJustificationAndFinalizationBench{
```

imo it's clear to the reader that these benchmarks are ran for all forks (especially after seeing the other benchmarks below)

Same to the other comments for the rest of the scope

bench/state_transition/process_epoch.zig
```diff
@@ -0,0 +1,702 @@
...
+
+    // Post-Fulu
+    if (comptime fork.isPostFulu()) {
+        try bench.addParam("proposer_lookahead", &ProcessProposerLookaheadBench{
+            .cached_state = cached_state,
+        }, .{});
+    }
+
+    // Actual processEpoch function
+    try bench.addParam("process_epoch", &ProcessEpochBench{ .cached_state = cached_state }, .{});
```
spiral-ladder: What does "Actual processEpoch function" mean?

spiral-ladder (commented):
Looking great - gave a pass through

bench/state_transition/process_epoch.zig
guha-rahul: I meant the non segmented one. Is "continuous" a better name or something more descriptive?

bench/state_transition/process_block.zig
guha-rahul: I already used assigned SignedBeaconBlock above , I am getting error. Should i use `signedbeaconblock` instead of `SignedBeaconBlockType `?

bench/state_transition/utils.zig
```diff
@@ -0,0 +1,49 @@
...
+/// Load and deserialize SignedBeaconBlock from SSZ bytes for a specific fork
+pub fn loadBlock(comptime fork: ForkSeq, allocator: std.mem.Allocator, block_bytes: []const u8) !SignedBeaconBlock {
+    const ForkTypes = @field(types, @tagName(fork));
+    const SignedBeaconBlockType = ForkTypes.SignedBeaconBlock;
+    const block_data = try allocator.create(SignedBeaconBlockType.Type);
+    errdefer allocator.destroy(block_data);
+    block_data.* = SignedBeaconBlockType.default_value;
+    try SignedBeaconBlockType.deserializeFromBytes(allocator, block_bytes, block_data);
+    return @unionInit(SignedBeaconBlock, @tagName(fork), block_data);
+}
```
spiral-ladder: Think this might be saner way to manage the `SignedBeaconBlock` namespace conflict so that it's clear that we're not returning the ssz spec's `SignedBeaconBlock`

```suggestion
pub fn loadBlock(comptime fork: ForkSeq, allocator: std.mem.Allocator, block_bytes: []const u8) !state_transition.SignedBeaconBlock {
    const SignedBeaconBlock = @field(types, @tagName(fork)).SignedBeaconBlock;
    const block_data = try allocator.create(SignedBeaconBlock.Type);
    errdefer allocator.destroy(block_data);
    block_data.* = SignedBeaconBlock.default_value;
    try SignedBeaconBlock.deserializeFromBytes(allocator, block_bytes, block_data);
    return @unionInit(state_transition.SignedBeaconBlock, @tagName(fork), block_data);
}
```

bench/state_transition/utils.zig
```diff
@@ -0,0 +1,49 @@
...
+/// Read slot from raw SignedBeaconBlock SSZ bytes (offset 100)
+pub fn slotFromBlockBytes(block_bytes: []const u8) Slot {
+    std.debug.assert(block_bytes.len >= 108);
+    return std.mem.readInt(u64, block_bytes[100..108], .little);
+}
+
+/// Load and deserialize BeaconState from SSZ bytes for a specific fork
+pub fn loadState(comptime fork: ForkSeq, allocator: std.mem.Allocator, state_bytes: []const u8) !*BeaconStateAllForks {
+    const ForkTypes = @field(types, @tagName(fork));
+    const BeaconState = ForkTypes.BeaconState;
```
spiral-ladder: `ForkTypes` can be confusing since what we're doing here is actually picking a specific fork using `@tagName(fork)` and not pulling all fork types here. IMO let's skip defining this:

```suggestion
    const BeaconState = @field(types, @tagName(fork)).BeaconState;
```

bench/state_transition/process_block.zig
```diff
@@ -0,0 +1,449 @@
...
+// step                   runs     total time     time/run (avg)
+// ---------------------------------------------------------------------
+// block_total            50       1.041s         20.828ms
+// block_header           50       53.676ms        1.074ms
+// withdrawals            50       1.191ms        0.024ms
+// execution_payload      50       26.506ms        0.530ms
+// randao                 50       38.248ms        0.765ms
+// eth1_data              50       0.055ms        0.001ms
+// operations             50       852.513ms        17.050ms
+// sync_aggregate         50       69.058ms        1.381ms
```
spiral-ladder: I tend to think we don't need these details here, rather it should belong in some other docs in future, since benchmarking is highly dependent on hardware. Let's make this top-level comment just describe what we're benchmarking and how to run the benchmarks.

As for 'how to run the benchmarks', let's prefer `zig build ...` rather than `zbuild`, since they're equivalent anyway:

```suggestion
//! Benchmark for fork-specific block processing.
//!
//! We use a mainnet state at slot 13180928 and a mainnet block at slot 13180929.
//! Run with zig build run:bench_process_block -Doptimize=ReleaseFast [-- /path/to/state.ssz /path/to/block.ssz]
```

bench/state_transition/process_block.zig
```diff
@@ -0,0 +1,449 @@
...
+const ProcessBlockHeaderBench = struct {
+    cached_state: *CachedBeaconStateAllForks,
+    signed_block: SignedBlock,
+
+    pub fn run(self: ProcessBlockHeaderBench, allocator: std.mem.Allocator) void {
+        const cloned = self.cached_state.clone(allocator) catch unreachable;
+        defer {
+            cloned.deinit();
+            allocator.destroy(cloned);
+        }
```
spiral-ladder: We can save this for a future PR, but we should probably move this clone (and other clones) out of the run definition since cloning should not be part of the cost of benchmarking `processXXX`

bench/state_transition/process_epoch.zig
```diff
@@ -0,0 +1,680 @@
...
+        if (count == 0) continue;
+        const total_ns = step_durations_ns[idx];
+        const avg_ns: u128 = total_ns / count;
+        const total_ms = @as(f64, @floatFromInt(total_ns)) / 1_000_000.0;
+        const avg_ms = @as(f64, @floatFromInt(avg_ns)) / 1_000_000.0;
+        if (total_ms >= 1000.0) {
+            try stdout.print("{s:<28} {d:<8} {d:>10.3}s   {d:>10.3}ms\n", .{ @tagName(step), count, total_ms / 1000.0, avg_ms });
+        } else {
+            try stdout.print("{s:<28} {d:<8} {d:>10.3}ms   {d:>10.3}ms\n", .{ @tagName(step), count, total_ms, avg_ms });
+        }
```
spiral-ladder: Same comment as https://github.com/ChainSafe/lodestar-z/pull/121/changes#r2671052166

spiral-ladder (changes requested):
Thanks again for the cleanup! Just a few more loose ends. There's still some things to cleanup but we can leave those for a followup PR.

bench/state_transition/process_block.zig
```diff
@@ -0,0 +1,437 @@
...
+// step                   runs     total time     time/run (avg)
+// ---------------------------------------------------------------------
+// block_total            50       1.041s         20.828ms
+// block_header           50       53.676ms        1.074ms
+// withdrawals            50       1.191ms        0.024ms
+// execution_payload      50       26.506ms        0.530ms
+// randao                 50       38.248ms        0.765ms
+// eth1_data              50       0.055ms        0.001ms
+// operations             50       852.513ms        17.050ms
+// sync_aggregate         50       69.058ms        1.381ms
```
spiral-ladder:
```suggestion
```

Maybe I wasn't clear in this [comment](https://github.com/ChainSafe/lodestar-z/pull/121#discussion_r2671060590), but I think we should just remove all these since benchmarks are highly dependent on setup,so it just seems like noise here

PR #123, opened 2025-12-10
feat: add mutable view check

GrapeBaBa:
**Motivation**
Align TreeView's change tracking behavior with TypeScript's TreeViewDU implementation,  only mutable views are tracked in viewsChanged. 

**Description**
- Add `isViewMutable` function in type_kind.zig to determine if a type's TreeView is mutable
- Update `getElement` and `getField` in TreeView to conditionally mark elements as changed based on `isViewMutable`
- Add `getElementReadonly` function for List/Vector types

### DIFF
--- a/src/ssz/type/type_kind.zig
+++ b/src/ssz/type/type_kind.zig
@@ -30,3 +30,24 @@ pub fn isFixedType(T: type) bool {
         },
     };
 }
+
+/// Determines if the TreeView of this type is mutable.
+/// Returns false for:
+/// - Any BasicType (uint, bool)
+/// - ByteVector, ByteList
+/// All other composite types return TreeView wrappers that can be modified.
+pub fn isViewMutable(comptime ST: type) bool {
+    if (isBasicType(ST)) {
+        return false;
+    }
+
+    if (ST.kind == .list or ST.kind == .vector) {
+        if (@hasDecl(ST, "Element")) {
+            const E = ST.Element;
+            if (E.kind == .uint and E.fixed_size == 1) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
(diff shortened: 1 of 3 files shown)


src/ssz/type/type_kind.zig
```diff
@@ -30,3 +30,24 @@ pub fn isFixedType(T: type) bool {
...
+/// All other composite types return TreeView wrappers that can be modified.
+pub fn isViewMutable(comptime ST: type) bool {
+    if (isBasicType(ST)) {
+        return false;
+    }
+
+    if (ST.kind == .list or ST.kind == .vector) {
+        if (@hasDecl(ST, "Element")) {
+            const E = ST.Element;
+            if (E.kind == .uint and E.fixed_size == 1) {
```
lodekeeper-z, line 47: **Blocking if revived:** `TreeView(ByteVectorType(...))` is mutable in this Zig implementation because it exposes `setElement`, so classifying it as immutable prevents parent propagation. I reproduced this with a container holding a 32-byte vector: after `var bytes = try view.getField("bytes"); try bytes.setElement(0, 2); try view.commit();`, the parent root still begins with `0x01` instead of the expected `0x02`. The existing 56 integration tests pass, but this focused regression fails. The TypeScript analogy does not carry over directly: `@chainsafe/ssz`'s `ByteArrayType` returns a raw byte-array view and marks that immutable, while Zig returns a mutable `TreeView` wrapper. Keep byte lists/vectors mutable here unless their accessor returns a non-mutable value view, or otherwise redesign the API so mutations cannot be silently dropped.

lodekeeper-z (commented):
## Code review

I found one blocking correctness issue if this change is revived. I reproduced it with Zig 0.14.1 using a focused regression test; the existing integration suite passes, but the added case fails because a byte-vector child mutation is not propagated to its parent. See the inline comment for details.

The PR is already closed, so I am leaving this as a comment rather than requesting changes.

PR #125, opened 2025-12-11
chore: refactor TreeView

twoeths:
**Motivation**
while the TreeView is a good start, I found it hard to add more features to it:
- each specific TreeView needs its own properties, and we have to put into our common data structure like https://github.com/ChainSafe/lodestar-z/pull/100/files#diff-38966ea8f3099d4e205c423bb107d9bb236fcd474396a9cc1e97de9c78a19f0eR33
- there are naming conflicts, it will happen more when we add more methods to it. Ideally we want `get()` and `set()` apis for Container and Array but since we model them as a single TreeView, we have to add suffixes like `getField`/`setField` and `getElement`/`setElement`

**Description**
improve code isolation of TreeView:
- each type has its own TreeView struct, reuse if needed
- for now we have `ContainerTreeView` and `ArrayTreeView`, we can add more in upcoming PRs
- this is more aligned with the current ViewDU implementation in the current lodestar

**Notes**
- in other languages, I'd achieve this goal using class inheritance. With zig, we use composition instead
- the difference is not a lot of you check out your branch. I basically separate `TreeView` to `ContainerTreeView` and `ArrayTreeView`

### DIFF
--- a/src/ssz/tree_view.zig
+++ b/src/ssz/tree_view.zig
@@ -171,6 +171,8 @@ pub fn ContainerTreeView(comptime ST: type) type {
     return struct {
         base_view: BaseTreeView,
 
+        pub const SszType = ST;
+
         const Self = @This();
 
         pub fn init(allocator: Allocator, pool: *Node.Pool, root: Node.Id) !Self {
@@ -255,6 +257,8 @@ pub fn ContainerTreeView(comptime ST: type) type {
 pub fn ArrayTreeView(comptime ST: type) type {
     return struct {
         base_view: BaseTreeView,
+
+        pub const SszType = ST;
 
         const Self = @This();
 


test/int/ssz/tree_view.zig
```diff
@@ -16,32 +16,32 @@ test "TreeView container field roundtrip" {
...
+    var cp_view = try Checkpoint.TreeView.init(std.testing.allocator, &pool, root_node);
     defer cp_view.deinit();
 
     // get field "epoch"
-    try std.testing.expectEqual(42, try cp_view.getField("epoch"));
+    try std.testing.expectEqual(42, try cp_view.get("epoch"));
 
     // get field "root"
-    var root_view = try cp_view.getField("root");
+    var root_view = try cp_view.get("root");
```
twoeths: simplifying the api is great for DX

PR #138, opened 2025-12-22
feat: impl tree view serialize API 

GrapeBaBa:
**Motivation**

Implement the tree view serialize API

**Description**

- Add `serialize`, `serializeIntoBytes`, and `serializedSize` methods to the TreeViews(array basic/composite, list basic/composite, container), committing pending changes and serializing from the backing merkle tree.
- Implement tree-level `serializeIntoBytes` / `serializedSize` for core SSZ types (uint, bool, bitlist/bitvector, byte list/vector, list/vector/container variants) so TreeViews can delegate to the underlying type’s tree serialization.
- Expand integration test coverage substantially by porting/adding roundtrip + root/size/serialization assertions across TreeView kinds and multiple representative SSZ types.

### DIFF
--- a/src/ssz/tree_view/list_composite.zig
+++ b/src/ssz/tree_view/list_composite.zig
@@ -181,17 +181,6 @@ pub fn ListCompositeTreeView(comptime ST: type) type {
             try self.base_view.setChildNode(@enumFromInt(3), length_node);
         }
 
-        /// Serialize the tree view directly to binary data.
-        /// This commits any pending changes and then serializes from the merkle tree.
-        pub fn serialize(self: *Self) ![]u8 {
-            try self.commit();
-            const size = try self.serializedSize();
-            const output = try self.base_view.allocator.alloc(u8, size);
-            errdefer self.base_view.allocator.free(output);
-            _ = try self.serializeIntoBytes(output);
-            return output;
-        }
-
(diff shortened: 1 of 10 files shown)


src/ssz/tree_view/list_composite.zig
```diff
@@ -179,5 +180,29 @@ pub fn ListCompositeTreeView(comptime ST: type) type {
             errdefer self.base_view.pool.unref(length_node);
             try self.base_view.setChildNode(@enumFromInt(3), length_node);
         }
+
+        /// Serialize the tree view directly to binary data.
+        /// This commits any pending changes and then serializes from the merkle tree.
+        pub fn serialize(self: *Self) ![]u8 {
```
twoeths: I'm not sure if we really need this `serialize()` api
one of the pain point of lodestar is we have to allocate everywhere due to missing "*into()" apis, so with `clone()` we decided to only support "into()" api, also type only supported "serializeIntoBytes()"
then consumer needs to decide how they deal with memory: using stack vs heap, reuse memory etc..

I'd say let's drop the support for `serialize()` api, we can add it back in the future if it's really needed

twoeths (commented):
looks good to me, I dropped one minor comment regarding not to support `serialize()` apis for now

twoeths (approved):

Issue #78, opened 2025-11-03
Implement missing Tree APIs


ensi321:
Here are the ones used by lodestar's beacon-node and state-transition (including those that are used in upcoming glamsterdam). List might be non-exhaustive:

- [x] ListCompositeTreeViewDU.sliceFrom()
- [x] ListCompositeTreeViewDU.getAll()
- [x] ListCompositeTreeViewDU.toValue()
```
const TV = ssz.TreeView(Validators);
const view = try TV.init(...);
var out: TV.SszType.Type = undefined;
try TV.SszType.tree.toValue(view.data.root, view.pool, &out);
```
- [x] ListCompositeTreeViewDU.sliceTo()
- [x] ListBasicTreeViewDU.push()
- [x] ListBasicTreeViewDU.sliceTo()
- [x] ContainerTreeViewDUType.serialize()
- [x] ContainerTreeViewDUType.deserialize()
- [x] ContainerTreeViewDUType.toValue()
```
const TV = ssz.TreeView(BeaconState);
const view = try TV.init(...);
var out: TV.SszType.Type = undefined;
try TV.SszType.tree.toValue(view.data.root, view.pool, &out);
```
- [x] ContainerTreeViewDUType.type.tree_serializedSize()
- [ ] ContainerTreeViewDUType.defaultViewDU()
- [x] TreeViewDU.clone()
- [x] TreeViewDU.serializeToBytes()
- [x] ListCompositeTreeViewDU.push()
- [x] BitArrayTreeViewDU.toBoolArray()
```
const TV = ssz.TreeView(AggregationBits);
const view = try TV.init(...);
var out: TV.SszType.Type = undefined;
try TV.SszType.tree.toValue(view.data.root, view.pool, &out);
```
Prerequisite of #20

wemeetagain:
I tend to think ContainerNodeStruct as its implemented in lodestar (eg as a "node type") is hacky and should be avoided here. Worth thinking about other ways to accomplish the same.

GrapeBaBa:
close it since almost are already implemented, may open separate issue for new

PR #139, opened 2025-12-22
refactor(ssz): drop BaseTreeView, use tuple for ContainerTreeView

twoeths:
## Motivation

The previous `BaseTreeView` design had several structural problems:
- Child TreeViews were returned as shallow copies (`{.base_view=...}`), risking double-free or dangling pointers
- Extra per-type fields (e.g. `length`, cached child arrays) could not be stored in child views without hacks
- All TreeView types were tightly coupled, making `ContainerNodeStruct` (and similar) hard to implement
- Parent `commit()` could not propagate through modified child TreeViews because children were copies, not references

Part of #78.

## Changes

**Drop `BaseTreeView`** — each TreeView is now a self-contained struct. Reusable logic (e.g. `getChildNode`, `setChildNode`, `getLength`, `setLength`) is shared via standalone utilities, not inheritance.

**`ContainerTreeView` now uses a comptime tuple** instead of a runtime Map. Each field slot holds either a reference to a child TreeView or a native basic-type value. This gives type-level access to the child type at a given index, enabling future `ContainerNodeStructTreeView`.

**`ArrayBasicTreeView` / `ArrayCompositeTreeView`** now track:
- A `Map` of dirty nodes (committed to cache only on `commit()`)
- A `length` field updated lazily
- For composite: a `Map` of child TreeView references (so parent `commit()` propagates correctly)


src/ssz/tree_view/utils/type.zig
spiral-ladder: perhaps have this as a top-level struct `CloneOpts.zig`? type.zig is a bit vague, we can expand it later if we want a more general utils file

src/ssz/tree_view/bit_list.zig
```diff
@@ -34,60 +36,76 @@ pub fn BitListTreeView(comptime ST: type) type {
         const chunk_depth: Depth = chunkDepth(Depth, base_chunk_depth, ST);
         const BitOps = BitArray(chunk_depth);
 
-        pub fn init(allocator: Allocator, pool: *Node.Pool, root: Node.Id) !Self {
-            return Self{ .base_view = try BaseTreeView.init(allocator, pool, root) };
+        pub fn init(allocator: Allocator, pool: *Node.Pool, root: Node.Id) !*Self {
+            const ptr = try allocator.create(Self);
```
spiral-ladder: missing `errdefer allocator.destroy(ptr)`?

src/ssz/tree_view/bit_vector.zig
```diff
@@ -31,36 +33,50 @@ pub fn BitVectorTreeView(comptime ST: type) type {
         const chunk_depth: Depth = @intCast(ST.chunk_depth);
         const BitOps = BitArray(chunk_depth);
 
-        pub fn init(allocator: Allocator, pool: *Node.Pool, root: Node.Id) !Self {
-            return Self{ .base_view = try BaseTreeView.init(allocator, pool, root) };
+        pub fn init(allocator: Allocator, pool: *Node.Pool, root: Node.Id) !*Self {
+            const ptr = try allocator.create(Self);
```
spiral-ladder: missing `errdefer allocator.destroy(ptr)`?

src/ssz/tree_view/container.zig
```diff
@@ -3,52 +3,212 @@ const Allocator = std.mem.Allocator;
...
-            return Self{ .base_view = try self.base_view.clone(opts) };
+        pub fn clone(self: *Self, opts: CloneOpts) !*Self {
+            const ptr = try self.allocator.create(Self);
+            if (!opts.transfer_cache) {
+                ptr.* = .{
+                    .allocator = self.allocator,
+                    .pool = self.pool,
+                    .child_data = .{null} ** ST.chunk_count,
+                    .original_nodes = .{null} ** ST.chunk_count,
+                    .root = self.root,
```
spiral-ladder: if we copy `self.root` without ref, will we run into double unrefs if the original `self.root` was unrefed?

src/ssz/tree_view/container.zig
```diff
@@ -3,52 +3,212 @@ const Allocator = std.mem.Allocator;
...
+                .root = root,
+                .changed = .empty,
             };
+            return ptr;
         }
 
-        pub fn clone(self: *Self, opts: BaseTreeView.CloneOpts) !Self {
-            return Self{ .base_view = try self.base_view.clone(opts) };
+        pub fn clone(self: *Self, opts: CloneOpts) !*Self {
+            const ptr = try self.allocator.create(Self);
```
spiral-ladder: missing `errdefer self.allocator.destroy(ptr)`

src/ssz/tree_view/list_basic.zig
```diff
@@ -26,8 +27,11 @@ pub fn ListBasicTreeView(comptime ST: type) type {
         }
     }
 
-    return struct {
-        base_view: BaseTreeView,
+    const TreeView = struct {
+        allocator: Allocator,
+        chunks: Chunks,
+        _orig_len: usize,
+        _len: usize,
```
spiral-ladder: what's the difference between `_orig_len` and `_len`? Might be worth it to add some doc comments here to make the distinction

src/ssz/tree_view/bit_array.zig
```diff
@@ -6,29 +6,101 @@ const Depth = hashing.Depth;
...
+                return;
+            }
+
+            out.* = .{
+                .allocator = self.allocator,
+                .pool = self.pool,
+                .root = self.root,
+                .children_nodes = self.children_nodes,
+                .changed = .empty,
+            };
```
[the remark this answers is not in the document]
twoeths: nice, this makes it simpler

src/ssz/tree_view/list_basic.zig
twoeths: _orig_len is the original length before any modification while _len is the current len, added that to the comments

src/ssz/tree_view/list_basic.zig
```diff
@@ -121,61 +144,70 @@ pub fn ListBasicTreeView(comptime ST: type) type {
...
+        }
+
+        fn updateListLength(self: *Self) !void {
+            if (self._len == self._orig_len) {
+                return;
+            }
+
+            if (self._len > ST.limit) {
+                return error.LengthOverLimit;
+            }
```
spiral-ladder:
```suggestion
            if (self._len >= ST.limit) {
                return error.LengthOverLimit;
            }
```

Seems like a redundant check since every update to `self._len` should include this check by correctness (which we do already check when we update it above) + this should be inclusive of `ST.limit`. If we want to keep this, we can turn this into an assertion.

src/ssz/tree_view/container.zig
```diff
@@ -83,17 +237,26 @@ pub fn ContainerTreeView(comptime ST: type) type {
...
-                    ),
-                );
+                const existing = self.child_data[field_index];
+                if (existing) |child_value| {
+                    if (child_value == value) {
+                        return;
+                    }
+                }
+
+                self.child_data[field_index] = value;
```
spiral-ladder: do we need lines 243-248 or can we just set `self.child_data[field_index] = value`?

src/ssz/tree_view/container.zig
```diff
@@ -3,76 +3,230 @@ const Allocator = std.mem.Allocator;
...
+            if (changed_idx == 0) {
+                return;
+            }
+
+            const new_root = try self.root.setNodesAtDepth(self.pool, ST.chunk_depth, indices[0..changed_idx], nodes[0..changed_idx]);
+            try self.pool.ref(new_root);
+            self.pool.unref(self.root);
+            self.root = new_root;
+
+            self.changed.clearRetainingCapacity();
```
spiral-ladder:
```suggestion
            self.changed.clearRetainingCapacity();
            if (changed_idx == 0) {
                return;
            }

            const new_root = try self.root.setNodesAtDepth(self.pool, ST.chunk_depth, indices[0..changed_idx], nodes[0..changed_idx]);
            try self.pool.ref(new_root);
            self.pool.unref(self.root);
            self.root = new_root;
```

`self.changed` would remain populated if `changed_idx == 0` and we do an early return

src/ssz/tree_view/container.zig
```diff
@@ -3,76 +3,230 @@ const Allocator = std.mem.Allocator;
...
+                        nodes[changed_idx] = child_node;
+                        indices[changed_idx] = i;
+                        self.original_nodes[i] = child_node;
+                        changed_idx += 1;
+                    } else {
+                        var child_view = self.child_data[i] orelse return error.MissingChildView;
+                        try child_view.commit();
+                        const child_changed = if (self.original_nodes[i]) |orig_node| blk: {
+                            break :blk orig_node != child_view.getRoot();
+                        } else true;
```
spiral-ladder:
```suggestion
                        const child_changed = try child_view.commit();
```

Could we perhaps return a `bool` here instead of using `original_nodes` to track if the child changed?

src/ssz/tree_view/container.zig
```diff
@@ -3,76 +3,230 @@ const Allocator = std.mem.Allocator;
...
+                if (child_opt) |child| {
+                    if (!comptime isBasicType(ST.fields[i].type)) {
+                        @constCast(child).deinit();
+                    }
+                    self.child_data[i] = null;
+                }
+            }
+            inline for (0..ST.chunk_count) |i| {
+                // these nodes are unref by root
+                self.original_nodes[i] = null;
```
spiral-ladder:
```suggestion
                }
            }
            inline for (0..ST.chunk_count) |i| {
                // these nodes are unref by root
```

Do we need to set these to `null` if we're only calling this at `deinit`? Plus, could we perhaps just inline this function since `deinit()` is the only place we use this?

PR #140, opened 2025-12-23
feat: refactor for get and update central viewstore

GrapeBaBa:
**Motivation**

Currently the `childData` exists shadow copy issue which cause the get-set-commit not works. This PR try to use a central `ViewStore` to manage all the states for a view and all its children views. Use (view_id, store ptr) to differentiate the view and use view_id to lookup the view.  

The benefit of this approach is more easy to manage the children views memory lifecycle.


GrapeBaBa:
I am working another such as tx scoped store and don't need viewid just use gindex, may submit later

GrapeBaBa:
@wemeetagain This PR not fully verified, just for POC a solution for get and update issue

GrapeBaBa:
in favor of #146

PR #141, opened 2025-12-25
feat: implement bit_list and bit_vector tree view

GrapeBaBa:
**Motivation**

Implement `bit_list` and `bit_vector` tree view for `bit_list` and `bit_vector` ssz type.

**Description**

- Add dedicated bit-packed TreeViews for `BitList` and `BitVector`, and implement the `get`, `set`, `toBoolArray`, and `toBoolArrayInto` API.

### DIFF
--- a/src/ssz/tree_view/bit_array.zig
+++ b/src/ssz/tree_view/bit_array.zig
@@ -1,0 +1,97 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+const hashing = @import("hashing");
+const Depth = hashing.Depth;
+
+const Node = @import("persistent_merkle_tree").Node;
+const Gindex = @import("persistent_merkle_tree").Gindex;
+
+const BaseTreeView = @import("root.zig").BaseTreeView;
+
+/// Provides common bit array operations for both BitVectorTreeView and BitListTreeView.
+pub fn BitArray(comptime chunk_depth: Depth) type {
+    return struct {
+        const bits_per_chunk: usize = 256;
+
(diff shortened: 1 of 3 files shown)


src/ssz/type/bit_list.zig
```diff
@@ -248,7 +248,7 @@ pub fn BitListType(comptime _limit: comptime_int) type {
         pub const Element: type = BoolType();
         pub const limit: usize = _limit;
         pub const Type: type = BitList(limit);
-        pub const TreeView: type = ListBasicTreeView(@This());
```
twoeths: @GrapeBaBa  why does the `ListBasicTreeView` not work? they are all about getting a node, and extract to a `bool` from an index

twoeths: ah I see. So it's the issue of `itemsPerChunk` not the ListBasicTreeView itself.
we should fix `itemsPerChunk()` instead?

src/ssz/tree_view/bit_array.zig
```diff
@@ -0,0 +1,97 @@
...
+        pub fn fillBools(base_view: *BaseTreeView, values: []bool, len: usize) !void {
+            if (values.len != len) return error.InvalidSize;
+            if (len == 0) return;
+
+            const full_chunks = len / bits_per_chunk;
+            const remainder_bits = len % bits_per_chunk;
+            var dest = values;
+
+            for (0..full_chunks) |chunk_idx| {
+                const leaf_node = try base_view.getChildNode(Gindex.fromDepth(chunk_depth, chunk_idx));
```
twoeths: this could be improved to get all nodes given `chunk_depth`

PR #148, opened 2025-12-30
feat: impl the tree view clone API

GrapeBaBa:
**Motivation**

Implement the tree view `clone` like lodestar treeview.

**Description**

- Implemented `clone(dont_transfer_cache: bool)` on the base TreeView and exposed it across all TreeView wrappers, enabling creation of a new view sharing the same committed root and match the previous ts semantics.
- Added comprehensive integration tests covering `clone`.

### DIFF
--- a/src/ssz/tree_view/array_basic.zig
+++ b/src/ssz/tree_view/array_basic.zig
@@ -45,6 +45,10 @@ pub fn ArrayBasicTreeView(comptime ST: type) type {
             };
         }
 
+        pub fn clone(self: *Self, dont_transfer_cache: bool) !Self {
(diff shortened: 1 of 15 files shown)


GrapeBaBa:
`/gemini review`

wemeetagain (commented):
Generally lgtm.

One nit: I think `don't_transfer_cache` is really confusing (I know its a transliteration of existing lodestar code).
Consider removing the negative (`dont_`), possibly wrapping in an opts struct.

Imo something like:
```zig
pub const CloneOpts = struct {
   /// Well documented what it does / means
    transfer_cache: bool = true,
};
...

pub fn clone(self, opts: CloneOpts)
```

wemeetagain (approved):

PR #152, opened 2025-12-31
refactor: polish config module

wemeetagain:
This is an opinionated refactor to polish the code in `src/config`.

#### Pure refactoring
- Add documentation comments to most/all exported structs/methods
- Refactor chain and beacon config into top-level struct files
- Reorganize utility functions as methods, eg: `mergeChainConfig` -> `ChainConfig.merge`
- Rename network specific `chain_config`s
- Use `ChainConfig.merge` in network-specific configs which override other network configs
- Ultimately, the `config` module now only exports `ChainConfig`, `BeaconConfig`, `ForkSeq`, and network configurations, eg: `mainnet`, `hoodi`, etc

#### Bigger tweaks
- Add network-specific `genesis_validator_root` and preinitialized `config: BeaconConfig` - (reminder, we only pay for what we use)
- Remove `ForkSeq.isPost*` functions - These can all be replaced with `ForkSeq.gte` which is equally performant. and IMO just as readable.
- Refactor `BeaconConfig.init` - IMO this is a case of being overly paranoid about a struct being copied. Zig semantics guarantee copy elison of structs that are created and returned like `return .{...};` (Compiler optimizations can remove other cases of copies, but that is somewhat besides the point) `BeaconConfig`, in any case, is the kind of object initialized once at the beginning of an application. This refactor improves DX and also allows us to conveniently provide network-specific `BeaconConfig`s for use in testing / utilities.
- Refactor `BeaconConfig` methods to return const pointers to various internal data. This is a case where we can benefit from returning pointers to this readonly data and avoid copies.
- Add real blob schedules to networks

### DIFF
--- a/src/config/ChainConfig.zig
+++ b/src/config/ChainConfig.zig
@@ -107,8 +107,8 @@ pub const OptionalChainConfig = Optional(ChainConfig);
 ///
 /// For each field: if `overrides.<field>` is non-null, it replaces `config.<field>`.
 /// Otherwise the original value is kept. Precedence is always given to `overrides`.
-pub fn merge(config: ChainConfig, overrides: OptionalChainConfig) ChainConfig {
-    var merged = config;
+pub fn merge(config: *const ChainConfig, overrides: OptionalChainConfig) ChainConfig {
+    var merged = config.*;
     inline for (std.meta.fields(@TypeOf(overrides))) |field| {
         if (@field(overrides, field.name)) |value| {
             @field(merged, field.name) = value;
@@ -147,7 +147,7 @@ fn Optional(comptime T: type) type {
 test merge {
     const mainnet_config = @import("./networks/mainnet.zig").chain_config;
     const old_altair_epoch = mainnet_config.ALTAIR_FORK_EPOCH;
-    const merged_config = merge(mainnet_config, .{
+    const merged_config = mainnet_config.merge(.{
         .ALTAIR_FORK_EPOCH = old_altair_epoch + 1000,
     });
     try std.testing.expect(merged_config.ALTAIR_FORK_EPOCH == old_altair_epoch + 1000);
--- a/src/config/BeaconConfig.zig
+++ b/src/config/BeaconConfig.zig
@@ -194,7 +194,7 @@ pub fn forkSeqAtEpoch(self: *const BeaconConfig, epoch: Epoch) ForkSeq {
 
 /// Return the active fork version for `slot`.
 pub fn forkVersion(self: *const BeaconConfig, slot: Slot) *const [4]u8 {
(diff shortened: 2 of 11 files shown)


twoeths (commented):
this PR changes `beacon_config.zig` to `BeaconConfig.zig`, do we have a convention somewhere? cc @spiral-ladder

spiral-ladder:
> this PR changes `beacon_config.zig` to `BeaconConfig.zig`, do we have a convention somewhere? cc @spiral-ladder

A zig file can just be a struct, in which case the file name is recommended to be PascalCase, eg. see [`BitStack.zig`](https://codeberg.org/ziglang/zig/src/branch/master/lib/std/BitStack.zig). Makes it easier to see it's just a top-level struct at the directory-level.

src/config/ChainConfig.zig
```diff
@@ -0,0 +1,154 @@
+///! Full chain configuration for a specific Ethereum Beacon Chain network (e.g. mainnet).
+///!
+///! This struct mirrors the upstream configuration values used by the consensus spec,
+///! see https://github.com/ethereum/consensus-specs/tree/master/configs.
+const std = @import("std");
+const Preset = @import("preset").Preset;
+
+const ChainConfig = @This();
+
+PRESET_BASE: Preset,
```
spiral-ladder: I think we can omit these being SCREAMING_SNAKE_CASE now if we prefer. Zig stdlib is using snake_case for global vars too ([style guide](https://ziglang.org/documentation/master/#toc-Examples)), though I suppose its not a hard rule and we should do what we prefer. I tend to think snake_case might be more convenient.

src/config/fork_seq.zig
```diff
@@ -0,0 +1,82 @@
...
+    electra = 5,
+    fulu = 6,
+
+    /// Total number of fork variants.
+    pub const count: u8 = @intCast(@typeInfo(ForkSeq).@"enum".fields.len);
+
+    /// Returns the canonical tag name for this fork (e.g. `"capella"`).
+    pub fn name(self: ForkSeq) [:0]const u8 {
+        return @tagName(self);
+    }
```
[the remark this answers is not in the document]
spiral-ladder: we might even consider removing this because this seems like a wrapper around `@tagName(self)`

spiral-ladder (commented):
Nice cleanup, generally in favour of slimming things down such as removing the fork comparison fns

src/config/ChainConfig.zig
wemeetagain: I generally agree. The reason its SCREAMING case here is that this data is kinda owned by the specs, and typically interchanged via yaml, and we're just matching their variable names directly.

Issue #137, opened 2025-12-19
ssz: code path for serialized <-> tree


wemeetagain:
Currently our ssz library has:
- serialized <-> value
- value <-> tree

But not:
- serialized <-> tree

Usecase:
- read a serialized state into tree-backed form
- serialize a tree-backed beacon state to disk

PR #154, opened 2026-01-01
feat: implement tree deserialize API

GrapeBaBa:
**Motivation**
Implement ssz tree deserialize API

**Description**

Implements tree.deserializeFromBytes for SSZ types and add tests

Complete #137


wemeetagain (commented):
Can you add round trip deserialize->tree->serialize and assert byte equality to the ssz spec tests?

wemeetagain (approved):

PR #156, opened 2026-01-04
feat(treeview): add defaultTreeview helper

GrapeBaBa:
**Motivation**

Align lodestar `defaultViewDU` API


spiral-ladder (commented):
LGTM with 1 minor comment

I think gemini's suggestion of asserting `pool != null` is unnecessary

PR #157, opened 2026-01-05
feat(treeview): Implement the basic list sliceFrom

### LINKED ISSUE #135: Implement `sliceFrom` method for basic lists
The `sliceFrom` method is implemented for composite lists but missing for basic lists. Given the PR description mentions implementing `sliceFrom` API for list tree views and there's comprehensive test coverage for composite lists, this appears to be an omission. Basic lists should also support `sliceFrom` for API consistency.
```suggestion

        /// Return a new view containing all elements from `index` to the end of the list.
        pub fn sliceFrom(self: *Self, index: usize) !Self {
            try self.commit();

            const list_length = try self.length();

            // Empty list or starting from the beginning: return a view over the current root.
            if (list_length == 0 or index == 0) {
                return try Self.init(self.base_view.allocator, self.base_view.pool, self.base_view.data.root);
            }

            // If index is past the end, return an empty list view.
            if (index >= list_length) {
                var result = try Self.init(self.base_view.allocator, self.base_view.pool, self.base_view.data.root);
                try result.updateListLength(0);
                return result;
            }

            // Build a new list view containing elements from `index` to `list_length - 1`.
            var result = try Self.init(self.base_view.allocator, self.base_view.pool, self.base_view.data.root);
            try result.updateListLength(0);

            var i: usize = index;
            while (i < list_length) : (i += 1) {
                const value = try self.get(i);
                try result.push(value);
            }

            return result;
        }
```

_Originally posted by @Copilot in https://github.com/ChainSafe/lodestar-z/pull/100#discussion_r2631363354_

GrapeBaBa:
**Motivation**
Add `sliceFrom` for basic list tree view.

**Description**
- Adds `sliceFrom` to `ListBasicTreeView` to return suffix views with correct length/root
- Extends SSZ tree view tests to cover sliceFrom behavior

Fix #135

### DIFF
--- a/src/ssz/tree_view/list_basic.zig
+++ b/src/ssz/tree_view/list_basic.zig
@@ -156,6 +156,132 @@ pub fn ListBasicTreeView(comptime ST: type) type {
             return try Self.init(self.base_view.allocator, self.base_view.pool, root_with_length);
         }
 
+        /// Return a new view containing all elements from `index` to the end.
+        /// The caller **must** call `deinit()` on the returned view to avoid memory leaks.
+        pub fn sliceFrom(self: *Self, index: usize) !Self {
+            try self.commit();
+
+            const list_length = try self.length();
+
+            // If index is 0, return a copy of the current view
+            if (index == 0) {
+                return try Self.init(self.base_view.allocator, self.base_view.pool, self.base_view.data.root);
+            }
+
+            const target_length = if (index >= list_length) 0 else list_length - index;
+
+            if (target_length == 0) {
+                const chunk_root: Node.Id = @enumFromInt(base_chunk_depth);
+                var length_node: ?Node.Id = try self.base_view.pool.createLeafFromUint(0);
+                defer if (length_node) |id| self.base_view.pool.unref(id);
+
+                const new_root = try self.base_view.pool.createBranch(chunk_root, length_node.?);
+                errdefer self.base_view.pool.unref(new_root);
+                length_node = null;
+
+                return try Self.init(self.base_view.allocator, self.base_view.pool, new_root);
+            }
+
+            const start_chunk_index = index / items_per_chunk;
+            const start_offset_in_chunk = index % items_per_chunk;
+            const end_chunk_index = (list_length - 1) / items_per_chunk;
+            const source_chunk_count = end_chunk_index - start_chunk_index + 1;
+
+            const new_chunk_count = (target_length + items_per_chunk - 1) / items_per_chunk;
+
+            const source_nodes = try self.base_view.allocator.alloc(Node.Id, source_chunk_count);
+            defer self.base_view.allocator.free(source_nodes);
+            try self.base_view.data.root.getNodesAtDepth(
+                self.base_view.pool,
+                chunk_depth,
+                start_chunk_index,
+                source_nodes,
+            );
+
+            const new_chunk_nodes = try self.base_view.allocator.alloc(Node.Id, new_chunk_count);
+            defer self.base_view.allocator.free(new_chunk_nodes);
+
+            const chunk_root_value: Node.Id = blk: {
+                // Track newly created leaf nodes for error cleanup.
+                //
+                // - aligned case: at most 1 leaf is newly created (only if we need to zero-pad the last chunk)
+                // - unaligned case: all chunks are newly created leaves
+                var newly_created_count: usize = 0;
+                var newly_created_last_leaf: ?Node.Id = null;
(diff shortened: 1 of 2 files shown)


src/ssz/tree_view/list_basic.zig
```diff
@@ -156,6 +156,167 @@ pub fn ListBasicTreeView(comptime ST: type) type {
...
+        pub fn sliceFrom(self: *Self, index: usize) !Self {
+            try self.commit();
+
+            const list_length = try self.length();
+
+            if (index == 0) {
+                return try Self.init(self.base_view.allocator, self.base_view.pool, self.base_view.data.root);
+            }
+
+            const target_length = if (index >= list_length) 0 else list_length - index;
```
spiral-ladder:
```suggestion
            const target_length = list_length -| index;
```

consider using saturating sub

spiral-ladder (changes requested):
partial review first

src/ssz/tree_view/list_basic.zig
```diff
@@ -156,6 +156,167 @@ pub fn ListBasicTreeView(comptime ST: type) type {
...
+
+                var chunk_bytes = source_nodes[i].getRoot(self.base_view.pool).*;
+                const keep_bytes = elems_in_last * ST.Element.fixed_size;
+                if (keep_bytes < BYTES_PER_CHUNK) {
+                    @memset(chunk_bytes[keep_bytes..], 0);
+                }
+
+                new_chunk_nodes[i] = try self.base_view.pool.createLeaf(&chunk_bytes);
+                newly_created_last_leaf = new_chunk_nodes[i];
+            }
```
spiral-ladder: This is a bit confusing at first glance because we're mixing two conditions within this for loop:

1. the case where there are `target_length % items_per_chunk` elements left to handle, and
2. the case where the `list_length` is perfectly aligned and we can just naively assign `new_chunk_nodes`

Suggestion: handle the first `source_nodes.len - 1` in a while loop in both cases, and then the last element in an if-else:

```suggestion
            var i: usize = 0;
            while (i < source_nodes.len - 1) : (i += 1) {
                new_chunk_nodes[i] = source_nodes[i];
            }

            if (list_length % items_per_chunk == 0) {
                new_chunk_nodes[i] = source_nodes[i];
            } else {
                const elems_in_last = target_length % items_per_chunk;
                std.debug.assert(elems_in_last != 0);

                var chunk_bytes = source_nodes[i].getRoot(self.base_view.pool).*;
                const keep_bytes = elems_in_last * ST.Element.fixed_size;
                if (keep_bytes < BYTES_PER_CHUNK) {
                    @memset(chunk_bytes[keep_bytes..], 0);
                }

                new_chunk_nodes[i] = try self.base_view.pool.createLeaf(&chunk_bytes);
                newly_created_last_leaf = new_chunk_nodes[i];
            }

```

test/int/ssz/tree_view/list_basic.zig
```diff
@@ -735,3 +735,304 @@ test "ListBasicTreeView - sliceTo and serialize" {
...
+
+            var actual_root: [32]u8 = undefined;
+            try sliced.hashTreeRoot(&actual_root);
+
+            try std.testing.expectEqualSlices(u8, &expected_root, &actual_root);
+        }
+    }
+}
+
+test "TreeView basic list sliceFrom works with different element sizes" {
```
spiral-ladder: Could we create `TestCase`s of the different types and use the same test code for both? The `TestCase` can look something like:

```zig
const TestCase = struct {
    List: type,
    Element: type,
}
```

I think we also use this pattern elsewhere in other test code

spiral-ladder (changes requested):
Looks great, besides some questions and comments in the review, also want to point out:

- in general we could probably use more doc comments. The functions we're adding has _a lot_ of parameters, some of which I had to read and do double takes to understand what they are about. **I think we should strive to add useful doc comments for all functions that we have and to keep them updated**. Useful, as in the doc comments tell us in a few lines what the function is about to do, and explain what some of the parameters do.
- I think the tests could be a lot simpler. That's a lot of code that we're adding that we eventually have to maintain (our `list_basic.zig` tests are >1000 lines!) Some cases look like they could be combined in one test with seperate scopes (eg. the 'TreeView basic list sliceFrom returns suffix' and 'TreeView basic list sliceFrom index 0 returns copy' tests are essentially the same code). In general, **we have to be very careful about adding code** to minimize complexity, and delete code wherever we can.

PR #161, opened 2026-01-08
refactor: simplify `BitList` and `BitVector` types

spiral-ladder:
we have both:

- `BitList` versus `BitListType`, and
- `BitVector` versus `BitVectorType`

But really the first type in both cases are just the internal implementations of the exposed types (the second), and they share the same comptime constants too: `limit` for `BitList` and `length` for `BitVector`. So we just unify them into one type each.

This makes it harder by design to accidentally use the internal impls (which we were exporting in `ssz/root.zig`).

Other type-specific changes:

## BitList
- standardize `deinit()` api. Previously, the deinit for the internal `BitList` had this signature

```zig
 pub fn deinit(self: *@This(), allocator: std.mem.Allocator) void {
            self.data.deinit(allocator);
}
``` 

versus `BitListType`:

```zig
        pub fn deinit(allocator: std.mem.Allocator, value: *Type) void {
            value.data.deinit(allocator);
        }
```

We just use the second one now. So instead of 

```zig
    var b: Bits.Type = try Bits.Type.fromBitLen(allocator, 30);
    defer b.deinit(allocator);
```

We now do

```zig
     var b: Bits = try Bits.fromBitLen(allocator, 30);
    defer Bits.deinit(allocator, &b);
```

which is consistent with the other SSZ types.

## BitVector

- deduplicate `byte_len` and `byte_length` - they're just the same but re-assigned and used differently in the external vs inner types.

### DIFF
--- a/src/state_transition/block/process_attestation_phase0.zig
+++ b/src/state_transition/block/process_attestation_phase0.zig
@@ -21,12 +21,6 @@ pub fn processAttestationPhase0(allocator: Allocator, cached_state: *CachedBeaco
 
     try validateAttestation(types.phase0.Attestation.Type, cached_state, attestation);
 
-    // should store a clone of aggregation_bits on Phase0 BeaconState to avoid double free error
-    const AggregationBits = ssz.BitListType(preset.MAX_VALIDATORS_PER_COMMITTEE);
-    var cloned_aggregation_bits: AggregationBits = undefined;
-    try AggregationBits.clone(allocator, &attestation.aggregation_bits, &cloned_aggregation_bits);
-    errdefer AggregationBits.deinit(allocator, &cloned_aggregation_bits);
-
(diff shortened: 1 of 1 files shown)


src/ssz/type/bit_list.zig
```diff
@@ -14,17 +14,35 @@ const maxChunksToDepth = @import("hashing").maxChunksToDepth;
 const Node = @import("persistent_merkle_tree").Node;
 const BitListTreeView = @import("../tree_view/root.zig").BitListTreeView;
 
-pub fn BitList(comptime limit: comptime_int) type {
+pub fn isBitListType(ST: type) bool {
+    return ST == BitListType(ST.limit);
```
spiral-ladder: because `.kind`  and `Element.kind` are both declarations within the `BitListType` we do not need to check for them since this check checks the set of all these declarations

PR #165, opened 2026-01-11
feat: implement `loadState` API and binding

### LINKED ISSUE #159: Implement loadState() api
Given a seed TreeView BeaconSate and `[] const u8`, implement an api to return another TreeView BeaconState sharing the same tree
reference implementation: https://github.com/ChainSafe/lodestar/blob/86298a43e647760d3ca92b37e3db6ddf586b3f7c/packages/state-transition/src/util/loadState/loadState.ts#L20

GrapeBaBa:
**Motivation**

It need `loadState` API to integrate state-transition-z to lodestar

**Description**

- Implement `loadState()` migration semantics aligned with Lodestar: migrate a new BeaconState from SSZ bytes using a seed state, returning the migrated state plus the list of modified validator indices.
- Optimize validator/inactivity_scores handling by reusing seed subtrees where bytes are unchanged, and by computing modified indices via recursive byte-level diff (avoids full SSZ decoding for comparison).

Fix #159

### DIFF
--- a/src/state_transition/load_state.zig
+++ b/src/state_transition/load_state.zig
@@ -195,6 +195,9 @@ fn loadValidators(
     seed_state_validators_bytes: ?[]const u8,
 ) ![]ValidatorIndex {
     if (new_validators_bytes.len % ssz_bytes.VALIDATOR_BYTES_SIZE != 0) return error.InvalidSize;
+    if (seed_state_validators_bytes) |bytes| {
+        if (bytes.len % ssz_bytes.VALIDATOR_BYTES_SIZE != 0) return error.InvalidSize;
+    }
 
     const seed_validators = try types.phase0.Validators.TreeView.init(allocator, pool, seed_validators_node);
     defer seed_validators.deinit();
@@ -917,6 +920,30 @@ test "loadValidators rejects non-multiple-of-VALIDATOR_BYTES_SIZE" {
     try std.testing.expectError(
         error.InvalidSize,
         loadValidators(allocator, StateST, migrated_view, &pool, seed_validators_node, bad_bytes[0..], null),
+    );
+}
+
+test "loadValidators rejects malformed seed_state_validators_bytes" {
+    const allocator = std.testing.allocator;
+    var pool = try Node.Pool.init(allocator, 1024);
+    defer pool.deinit();
+
+    const gen = @import("test_utils/generate_state.zig");
+    const chain_config = gen.getConfig(@import("config").minimal.chain_config, .electra, 0);
+    const state_ptr = try gen.generateElectraState(allocator, &pool, chain_config, 8);
+    defer {
+        state_ptr.deinit();
+        allocator.destroy(state_ptr);
+    }
+
+    const StateST = types.electra.BeaconState;
+    const migrated_view = state_ptr.castToFork(.electra).inner;
+    const seed_validators_node = try validatorsNodeId(state_ptr);
+    const good_new_bytes = [_]u8{0} ** (ssz_bytes.VALIDATOR_BYTES_SIZE * 2);
+    const bad_seed_bytes = [_]u8{0} ** (ssz_bytes.VALIDATOR_BYTES_SIZE + 1);
+    try std.testing.expectError(
+        error.InvalidSize,
+        loadValidators(allocator, StateST, migrated_view, &pool, seed_validators_node, good_new_bytes[0..], bad_seed_bytes[0..]),
     );
 }
 
--- a/src/state_transition/load_state.zig
+++ b/src/state_transition/load_state.zig
@@ -43,7 +43,7 @@ pub fn loadState(
 ) !MigrateStateOutput {
     const fork = try ssz_bytes.getForkFromStateBytes(config, state_bytes);
     const seed_fork = config.forkSeq(try seed_state.slot());
-    const pool = seed_state.pool();
+    const pool = seed_state.nodePool();
 
     return switch (fork) {
         inline else => |f| try loadStateForFork(allocator, pool, f, seed_fork, seed_state, ForkTypes(f).BeaconState, state_bytes, seed_validators_bytes),
@@ -432,12 +432,12 @@ fn inactivityScoresNodeId(state: *AnyBeaconState) !Node.Id {
 
 fn validatorsViewOwned(allocator: Allocator, state: *AnyBeaconState) !*types.phase0.Validators.TreeView {
     const root = try validatorsNodeId(state);
-    return try types.phase0.Validators.TreeView.init(allocator, state.pool(), root);
+    return try types.phase0.Validators.TreeView.init(allocator, state.nodePool(), root);
 }
 
 fn inactivityScoresViewOwned(allocator: Allocator, state: *AnyBeaconState) !*types.altair.InactivityScores.TreeView {
     const root = try inactivityScoresNodeId(state);
-    return try types.altair.InactivityScores.TreeView.init(allocator, state.pool(), root);
+    return try types.altair.InactivityScores.TreeView.init(allocator, state.nodePool(), root);
 }
 
 fn loadValidatorWithSeedReuse(
--- a/src/state_transition/load_state.zig
+++ b/src/state_transition/load_state.zig
@@ -632,6 +632,11 @@ test "loadValidatorWithSeedReuse: reuse vs rebuild" {
     try std.testing.expect(
         try seed_validator.root.getNodeAtDepth(seed_validator.pool, types.phase0.Validator.chunk_depth, withdrawal_i) != try new_validator.root.getNodeAtDepth(new_validator.pool, types.phase0.Validator.chunk_depth, withdrawal_i),
     );
+
+    const fresh_root = try types.phase0.Validator.tree.deserializeFromBytes(&pool, new_validator_bytes[0..]);
+    var fresh_validator = try types.phase0.Validator.TreeView.init(allocator, &pool, fresh_root);
+    defer fresh_validator.deinit();
+    try std.testing.expectEqualSlices(u8, try fresh_validator.hashTreeRoot(), try new_validator.hashTreeRoot());
 }
 
 test "loadState scenarios" {
@@ -806,6 +811,10 @@ test "loadState scenarios" {
             _ = try mv.serializeIntoBytes(&mv_bytes);
             try std.testing.expectEqualSlices(u8, mutated_bytes[base .. base + ssz_bytes.VALIDATOR_BYTES_SIZE], mv_bytes[0..]);
         }
+
+        var fresh_state = try AnyBeaconState.deserialize(allocator, &pool, .electra, mutated_bytes);
+        defer fresh_state.deinit();
+        try std.testing.expectEqualSlices(u8, try fresh_state.hashTreeRoot(), try out.state.hashTreeRoot());
     }
 }
 
--- a/src/state_transition/load_state.zig
+++ b/src/state_transition/load_state.zig
@@ -202,16 +202,16 @@ fn loadValidators(
     new_validators_bytes: []const u8,
     seed_state_validators_bytes: ?[]const u8,
 ) ![]ValidatorIndex {
-    if (new_validators_bytes.len % ssz_bytes.VALIDATOR_BYTES_SIZE != 0) return error.InvalidSize;
+    if (new_validators_bytes.len % types.phase0.Validator.fixed_size != 0) return error.InvalidSize;
     if (seed_state_validators_bytes) |bytes| {
-        if (bytes.len % ssz_bytes.VALIDATOR_BYTES_SIZE != 0) return error.InvalidSize;
+        if (bytes.len % types.phase0.Validator.fixed_size != 0) return error.InvalidSize;
     }
 
     const seed_validators = try types.phase0.Validators.TreeView.init(allocator, pool, seed_validators_node);
     defer seed_validators.deinit();
 
     const seed_count = try seed_validators.length();
-    const new_count = new_validators_bytes.len / ssz_bytes.VALIDATOR_BYTES_SIZE;
+    const new_count = new_validators_bytes.len / types.phase0.Validator.fixed_size;
     const min_count = @min(seed_count, new_count);
 
     var migrated_validators = try seed_validators.clone(.{ .transfer_cache = false });
@@ -235,8 +235,8 @@ fn loadValidators(
     var modified_validators: std.ArrayList(ValidatorIndex) = .empty;
     errdefer modified_validators.deinit(allocator);
 
-    const old_validators_slice = seed_bytes[0 .. min_count * ssz_bytes.VALIDATOR_BYTES_SIZE];
-    const new_validators_slice = new_validators_bytes[0 .. min_count * ssz_bytes.VALIDATOR_BYTES_SIZE];
+    const old_validators_slice = seed_bytes[0 .. min_count * types.phase0.Validator.fixed_size];
+    const new_validators_slice = new_validators_bytes[0 .. min_count * types.phase0.Validator.fixed_size];
     try findModifiedValidators(allocator, old_validators_slice, new_validators_slice, &modified_validators, 0);
 
     try applyModifiedValidators(
@@ -361,9 +361,9 @@ fn applyModifiedValidators(
 ) !void {
     for (modified_validators) |validator_index| {
         const i: usize = @intCast(validator_index);
-        const start = i * ssz_bytes.VALIDATOR_BYTES_SIZE;
-        const new_bytes = new_validators_bytes[start .. start + ssz_bytes.VALIDATOR_BYTES_SIZE];
-        const seed_val_bytes = seed_bytes[start .. start + ssz_bytes.VALIDATOR_BYTES_SIZE];
+        const start = i * types.phase0.Validator.fixed_size;
+        const new_bytes = new_validators_bytes[start .. start + types.phase0.Validator.fixed_size];
+        const seed_val_bytes = seed_bytes[start .. start + types.phase0.Validator.fixed_size];
 
         const seed_validator = try seed_validators.get(i);
         // seed_validator is borrowed from seed_validators; do not deinit.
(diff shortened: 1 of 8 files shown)


src/state_transition/types/beacon_state_tree_view.zig
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    pub fn validators(self: *@This()) !types.phase0.Validators.TreeView {
+        const root = try self.validatorsNodeId();
+        // NOTE(ownership): Returns an *owned* view (caller must deinit).
+        // Borrowed child views built by copying cached TreeViewData can be unsafe (UAF/leaks) once
+        // they perform cached operations (e.g. length/get/clone with cache transfer).
+        // TODO(ssz/tree_view): Fix child-view semantics so borrowed views are safe.
+        return try types.phase0.Validators.TreeView.init(self.allocator(), self.pool(), root);
+    }
```
spiral-ladder: This note feels like it should be a top-level doc comment so that any users of `validators()` see it. This is useful especially when using LSPs

```suggestion
    /// Caller must deinit *owned* view.
    ///
    /// Borrowed child views built by copying cached TreeViewData can be unsafe (UAF/leaks) once
    /// they perform cached operations (e.g. length/get/clone with cache transfer).
    // TODO(ssz/tree_view): Fix child-view semantics so borrowed views are safe.
    pub fn validators(self: *@This()) !types.phase0.Validators.TreeView {
        const root = try self.validatorsNodeId();
        return try types.phase0.Validators.TreeView.init(self.allocator(), self.pool(), root);
    }
```

src/state_transition/types/beacon_state_tree_view.zig
```diff
@@ -0,0 +1,135 @@
...
+
+    pub fn inactivityScores(self: *@This()) !types.altair.InactivityScores.TreeView {
+        const root = try self.inactivityScoresNodeId();
+        // NOTE(ownership): Returns an *owned* view (caller must deinit).
+        // Borrowed child views built by copying cached TreeViewData can be unsafe (UAF/leaks) once
+        // they perform cached operations (e.g. length/get/clone with cache transfer).
+        // TODO(ssz/tree_view): Fix child-view semantics so borrowed views are safe.
```
spiral-ladder: same as the comment for `validators()`

GrapeBaBa:
need to align once #168 merged

spiral-ladder:
Can we add benchmarks for this comparing this versus the typescript impl? Similar to https://github.com/ChainSafe/lodestar-z/pull/351/changes#diff-d61152f5f9c7ce5bf850babd51cad04c7a356548f6d9fde61bd8ec658b6a5eb9

GrapeBaBa:
> Can we add benchmarks for this comparing this versus the typescript impl? Similar to https://github.com/ChainSafe/lodestar-z/pull/351/changes#diff-d61152f5f9c7ce5bf850babd51cad04c7a356548f6d9fde61bd8ec658b6a5eb9



> Can we add benchmarks for this comparing this versus the typescript impl? Similar to https://github.com/ChainSafe/lodestar-z/pull/351/changes#diff-d61152f5f9c7ce5bf850babd51cad04c7a356548f6d9fde61bd8ec658b6a5eb9



> Can we add benchmarks for this comparing this versus the typescript impl? Similar to https://github.com/ChainSafe/lodestar-z/pull/351/changes#diff-d61152f5f9c7ce5bf850babd51cad04c7a356548f6d9fde61bd8ec658b6a5eb9

added. But need to check the #352 first

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+const AnyBeaconState = @import("fork_types").AnyBeaconState;
+const ForkTypes = @import("fork_types").ForkTypes;
+
+const ssz_bytes = @import("ssz_bytes.zig");
+const ssz_container = @import("ssz_container.zig");
+
+const ValidatorIndex = types.primitive.ValidatorIndex.Type;
+
+/// Inactivity score is `uint64` (8 bytes).
+const INACTIVITY_SCORE_SIZE: usize = 8;
```
spiral-ladder: nit: can we use

```suggestion
const INACTIVITY_SCORE_SIZE: usize = ct.primitive.Uint64.fixed_size;
```

instead?

src/state_transition/ssz_bytes.zig
```diff
@@ -0,0 +1,35 @@
...
+/// class BeaconState(Container):
+///   genesis_time: uint64 [fixed - 8 bytes]
+///   genesis_validators_root: Root [fixed - 32 bytes]
+///   slot: Slot [fixed - 8 bytes]
+///   ...
+/// ```
+const SLOT_BYTES_POSITION_IN_STATE: usize = 40;
+
+/// Slot is `uint64`.
+const SLOT_BYTE_COUNT: usize = 8;
```
spiral-ladder:
```suggestion
```

We could remove this and use `types.primitive.Slot.fixed_size` directly below. We also redefine `SLOT_BYTE_COUNT` in various places in `lodestar` so let's avoid this anti-pattern

```
➜ rg SLOT_BYTE_COUNT
packages/state-transition/src/util/sszBytes.ts
9:const SLOT_BYTE_COUNT = 8;
41:  const slot = bytesToInt(bytes.subarray(SLOT_BYTES_POSITION_IN_STATE, SLOT_BYTES_POSITION_IN_STATE + SLOT_BYTE_COUNT));
51:  return bytesToInt(bytes.subarray(SLOT_BYTES_POSITION_IN_STATE, SLOT_BYTES_POSITION_IN_STATE + SLOT_BYTE_COUNT));

packages/beacon-node/src/util/multifork.ts
10:const SLOT_BYTE_COUNT = 8;
50:  return bytesToInt(bytes.subarray(SLOT_BYTES_POSITION_IN_STATE, SLOT_BYTES_POSITION_IN_STATE + SLOT_BYTE_COUNT));
71:    bytes.subarray(SLOT_BYTES_POSITION_IN_LIGHTCLIENTHEADER, SLOT_BYTES_POSITION_IN_LIGHTCLIENTHEADER + SLOT_BYTE_COUNT)

packages/beacon-node/src/db/repositories/lightclientBestUpdate.ts
6:const SLOT_BYTE_COUNT = 8;
29:    const prefixedData = new Uint8Array(SLOT_BYTE_COUNT + valueBytes.length);
31:    prefixedData.set(valueBytes, SLOT_BYTE_COUNT);
38:    const slot = ssz.Slot.deserialize(data.subarray(0, SLOT_BYTE_COUNT));
39:    return this.config.getPostAltairForkTypes(slot).LightClientUpdate.deserialize(data.subarray(SLOT_BYTE_COUNT));
```

src/state_transition/ssz_bytes.zig
```diff
@@ -0,0 +1,35 @@
...
+
+/// SSZ-serialized size of a phase0 Validator. Stable across forks.
+pub const VALIDATOR_BYTES_SIZE: usize = types.phase0.Validator.fixed_size;
+
+pub fn getStateSlotFromBytes(bytes: []const u8) !Slot {
+    if (bytes.len < SLOT_BYTES_POSITION_IN_STATE + SLOT_BYTE_COUNT) return error.InvalidSize;
+    return @intCast(std.mem.readInt(u64, bytes[SLOT_BYTES_POSITION_IN_STATE .. SLOT_BYTES_POSITION_IN_STATE + SLOT_BYTE_COUNT], .little));
```
spiral-ladder:
```suggestion
    return std.mem.readInt(u64, bytes[SLOT_BYTES_POSITION_IN_STATE .. SLOT_BYTES_POSITION_IN_STATE + SLOT_BYTE_COUNT], .little);
```

`Slot` is already `u64` so I think we don't need this `intCast`

src/state_transition/ssz_bytes.zig
```diff
@@ -0,0 +1,35 @@
...
+///   slot: Slot [fixed - 8 bytes]
+///   ...
+/// ```
+const SLOT_BYTES_POSITION_IN_STATE: usize = 40;
+
+/// Slot is `uint64`.
+const SLOT_BYTE_COUNT: usize = 8;
+
+/// SSZ-serialized size of a phase0 Validator. Stable across forks.
+pub const VALIDATOR_BYTES_SIZE: usize = types.phase0.Validator.fixed_size;
```
spiral-ladder: i'm a bit torn between allowing this to follow lodestar ts convention or just importing this at call site directly

src/state_transition/root.zig
```diff
@@ -105,9 +105,14 @@ const EpochShuffling = @import("./utils/epoch_shuffling.zig");
 pub const calculateShufflingDecisionRoot = EpochShuffling.calculateShufflingDecisionRoot;
 pub const processProposerLookahead = @import("./epoch/process_proposer_lookahead.zig").processProposerLookahead;
 
+const load_state_mod = @import("load_state.zig");
+pub const loadState = load_state_mod.loadState;
+pub const MigrateStateOutput = load_state_mod.MigrateStateOutput;
+
```
spiral-ladder:
```suggestion
const load_state = @import("load_state.zig");
pub const loadState = load_state.loadState;
pub const MigrateStateOutput = load_state.MigrateStateOutput;

```

src/state_transition/root.zig
```diff
@@ -105,9 +105,14 @@ const EpochShuffling = @import("./utils/epoch_shuffling.zig");
...
 test {
     testing.refAllDecls(@This());
     testing.refAllDecls(seed);
     testing.refAllDecls(state_transition);
     testing.refAllDecls(EpochShuffling);
+    testing.refAllDecls(load_state_mod);
```
spiral-ladder:
```suggestion
    testing.refAllDecls(load_state);
```

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,749 @@
...
+    owned_len = 0;
+    return try types.phase0.Validator.TreeView.init(allocator, pool, root);
+}
+
+fn findModifiedValidators(
+    validators_bytes: []const u8,
+    validators_bytes2: []const u8,
+    modified_validators: *std.ArrayList(ValidatorIndex),
+    validator_offset: usize,
+) !void {
```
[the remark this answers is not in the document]
spiral-ladder: making this non-recursive might be worth it for style in a separate PR, but i think the generic comment is a bit pedantic

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+    return try types.phase0.Validator.TreeView.init(allocator, pool, root);
+}
+
+fn findModifiedValidators(
+    allocator: Allocator,
+    validators_bytes: []const u8,
+    validators_bytes2: []const u8,
+    modified_validators: *std.ArrayList(ValidatorIndex),
+    validator_offset: usize,
+) !void {
```
spiral-ladder, line 527: Think it's worth doc commenting this (and `findModifiedinactivityScores`), because it's not immediately clear what `validator_offset` is
```suggestion
  /// Appends absolute indices (offset by `validator_offset`) of validators which
  /// differ between the two equal-length, VALIDATOR_BYTES_SIZE-aligned slices.
fn findModifiedValidators(
    allocator: Allocator,
    validators_bytes: []const u8,
    validators_bytes2: []const u8,
    modified_validators: *std.ArrayList(ValidatorIndex),
    validator_offset: usize,
) !void {
```

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+}
+
+fn validatorsViewOwned(allocator: Allocator, state: *AnyBeaconState) !*types.phase0.Validators.TreeView {
+    const root = try validatorsNodeId(state);
+    return try types.phase0.Validators.TreeView.init(allocator, state.nodePool(), root);
+}
+
+fn inactivityScoresViewOwned(allocator: Allocator, state: *AnyBeaconState) !*types.altair.InactivityScores.TreeView {
+    const root = try inactivityScoresNodeId(state);
+    return try types.altair.InactivityScores.TreeView.init(allocator, state.nodePool(), root);
```
spiral-ladder: Given that these are used for tests (and only called once each), can we just inline them?

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+    }
+
+    const seed_validators = try types.phase0.Validators.TreeView.init(allocator, pool, seed_validators_node);
+    defer seed_validators.deinit();
+
+    const seed_count = try seed_validators.length();
+    const new_count = new_validators_bytes.len / ssz_bytes.VALIDATOR_BYTES_SIZE;
+    const min_count = @min(seed_count, new_count);
+
+    var migrated_validators = try seed_validators.clone(.{});
```
spiral-ladder: `CloneOpts`  default `transfer_cache` to `true` so I think we'd need to do this here:


```suggestion
    var migrated_validators = try seed_validators.clone(.{ .transfer_cache = false });
```

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+    errdefer migrated_validators.deinit();
+
+    const seed_bytes = try getSeedValidatorsBytes(allocator, seed_validators, seed_state_validators_bytes);
```
spiral-ladder: Can we retain our comment from typescript lodestar for this? Either here, or as a doc comment, is good


```suggestion
    // 80% of validators serialization time comes from memory allocation
    // seedStateValidatorsBytes is an optimization at beacon-node side to avoid memory allocation here
    const seed_bytes = try getSeedValidatorsBytes(allocator, seed_validators, seed_state_validators_bytes);
```

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+    const root = try validatorsNodeId(state);
+    return try types.phase0.Validators.TreeView.init(allocator, state.nodePool(), root);
+}
+
+fn inactivityScoresViewOwned(allocator: Allocator, state: *AnyBeaconState) !*types.altair.InactivityScores.TreeView {
+    const root = try inactivityScoresNodeId(state);
+    return try types.altair.InactivityScores.TreeView.init(allocator, state.nodePool(), root);
+}
+
+fn loadValidatorWithSeedReuse(
```
spiral-ladder, line 457: Retain doc comments from typescript lodestar. Rephrase as you wish but keep the essence of it (why we do this)



```suggestion
// Load validator from bytes given a seed validator.
// - Reuse pubkey and withdrawal credentials if possible to save memory
// - If it's a new validator, deserialize it
fn loadValidatorWithSeedReuse(
```

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+        inactivity_scores_bytes;
+
+    try findModifiedInactivityScores(allocator, old_scores_slice, new_scores_slice, &modified_validators, 0);
+    try applyScoreDiffs(migrated_scores, inactivity_scores_bytes, modified_validators.items);
+    migrated_scores = try syncScoresLength(allocator, migrated_scores, inactivity_scores_bytes, diff_ctx.old_validator_count, diff_ctx.new_validator_count);
+
+    try migrated_view.set("inactivity_scores", migrated_scores);
+}
+
+fn loadValidators(
```
spiral-ladder, line 197: This function is quite involved and full of helper functions, it would be very useful to doc comment this (as well as all the inner functions, such as `applyModifiedValidators`)

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+    if (seed_state_validators_bytes) |bytes| {
+        return .{ .bytes = bytes, .owned = false };
+    }
+
+    const size = try seed_validators.serializedSize();
+    const out = try allocator.alloc(u8, size);
+    errdefer allocator.free(out);
+    _ = try seed_validators.serializeIntoBytes(out);
+    return .{ .bytes = out, .owned = true };
+}
```
spiral-ladder: I find this function misleading and confusing, we're calling this a getter but we're doing a lot more work than just getting (possibly allocating). We should maybe just do this inline, and the benefit of that is we wouldn't need an `owned` to tell zig if we need to deallocate or not

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+    old_validator_count: usize,
+    new_validator_count: usize,
+) !*types.altair.InactivityScores.TreeView {
+    if (new_validator_count >= old_validator_count) {
+        var idx: usize = old_validator_count;
+        while (idx < new_validator_count) : (idx += 1) {
+            const start = idx * INACTIVITY_SCORE_SIZE;
+            const chunk: *const [INACTIVITY_SCORE_SIZE]u8 = @ptrCast(inactivity_scores_bytes[start .. start + INACTIVITY_SCORE_SIZE].ptr);
+            const value = std.mem.readInt(u64, chunk, .little);
+            try migrated_scores.push(@intCast(value));
```
spiral-ladder:
```suggestion
            try migrated_scores.push(value);
```

`InactivityScore` type is `u64` so we can skip casting

src/state_transition/load_state.zig
```diff
@@ -0,0 +1,971 @@
...
+        }
+
+        try std.testing.expectEqual(case.modified.len, got.items.len);
+        for (case.modified, got.items) |e, g| {
+            try std.testing.expectEqual(@as(ValidatorIndex, @intCast(e)), g);
+        }
+    }
+}
+
+test "loadValidators rejects non-multiple-of-VALIDATOR_BYTES_SIZE" {
```
spiral-ladder: These reject scenarios have roughly the same setup; we can likely put them in the same test named `test "loadValidators: rejection scenarios"` and then put each scenario in their own scope with a comment saying what it tests

spiral-ladder (commented):
sorry this took awhile, took a deeper look. Bench looks good (on M2 Mac), just some comments on the code

```
bindings/perf/loadState.test.ts
  loadState: native vs TS (mainnet)
    ✔ native (internal serialize seed)                                    3.803396 ops/s    262.9229 ms/op        -          9 runs   3.15 s
    ✔ TS (internal serialize seed)                                        1.552710 ops/s    644.0352 ms/op        -          4 runs   3.47 s
    ✔ native (prebuilt seedValidatorsBytes)                               22.97088 ops/s    43.53338 ms/op        -         36 runs   2.09 s
    ✔ TS (prebuilt seedValidatorsBytes)                                   4.441523 ops/s    225.1480 ms/op        -          6 runs   1.94 s
```

wemeetagain (approved):

spiral-ladder (approved):
LGTM, thank you for the changes! @GrapeBaBa

PR #166, opened 2026-01-12
test: nuke int; move state transition tests to state transition module

spiral-ladder:
In a similar vein with moving the `ssz` tests to the ssz module, we move the state transition tests to `state_transition` because they aren't actually integration tests, and with that we nuke `int` entirely.

We rename `int_slow` to `int` - technically our era tests are the only 'integration' tests anyway!

Why this re-organization:

- `int` was a ambiguous test directory name to begin with - there were some tests spread between their own module and within the `int` directory, despite the tests actually not being integration in nature. This was more obvious to see in the `ssz` tests refactored in #162.

- we can now test each module independently. Previously, if we made changes to `ssz`, we're forced to do both `zig build test:ssz` and `zig build test:int` because our tests were scattered. It also unnecessarily runs the `state_transition` tests since they're both put in `int`. With this structure we can test only the module we are working on and made changes in.

- test locality - it's obvious right away if a certain key procedure does not have tests, if the tests are located in proximity with the implementation.

This is a big PR in terms of files changed but really, it's all just file deletions and moving code around.

### DIFF
--- a/src/state_transition/block/process_execution_payload.zig
+++ b/src/state_transition/block/process_execution_payload.zig
@@ -30,6 +30,7 @@ pub fn processExecutionPayload(
     external_data: BlockExternalData,
 ) !void {
     const state = cached_state.state;
+    const beacon_config = cached_state.config;
     const epoch_cache = cached_state.getEpochCache();
     var partial_payload = PartialPayload{};
     switch (body) {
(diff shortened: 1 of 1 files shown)


twoeths (dismissed):

wemeetagain (dismissed):
yes lets please wait to merge until we have merged https://github.com/ChainSafe/lodestar-z/pull/168

PR #176, opened 2026-01-16
feat: add bindings for compute_proposer_index

guha-rahul:
Adds  napi bindings for `computeProposerIndex` , the TS version 
https://github.com/ChainSafe/lodestar/blob/b79f41e05427f9ce5f5371a93ad0a9f6b89d6f2b/packages/state-transition/src/util/seed.ts#L109
uses Rust bindings.

### DIFF
--- a/bindings/test/proposer_index.test.ts
+++ b/bindings/test/proposer_index.test.ts
@@ -15,33 +15,17 @@ describe("computeProposerIndex", () => {
     effectiveBalanceIncrements[i] = 32 + 32 * (i % 64);
   }
 
-  it("should compute proposer index for phase0", () => {
-    const result = bindings.computeProposerIndex("phase0", effectiveBalanceIncrements, indices, seed);
-    expect(result).toBe(789);
-  });
+  it("should compute proposer index for all forks", () => {
+    const testCases = [
+      {expected: 789, forks: ["phase0", "altair", "bellatrix", "capella", "deneb"] as const},
+      {expected: 161, forks: ["electra", "fulu"] as const},
+    ];
 
-  it("should compute proposer index for electra", () => {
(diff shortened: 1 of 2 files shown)


bindings/napi/proposer_index.zig
```diff
@@ -0,0 +1,68 @@
...
+const ByteCount = state_transition.committee_indices.ByteCount;
+const preset = state_transition.preset;
+
+/// Compute the proposer index for a given fork, effective balance increments, indices, and seed.
+///
+/// Arguments:
+/// - arg 0: fork (string)
+/// - arg 1: effective balance increments for all validators
+/// - arg 2: active validator indices
+/// - arg 3: random seed
```
spiral-ladder: nice! This is clearer at a glance what we expect from just looking at the fn (and also helps LSP users)

spiral-ladder (approved):
LGTM 🫡

bindings/napi/proposer_index.zig
guha-rahul: that makes sense, I will need to do the same for the getters in BeaconStateView then.

bindings/test/proposer_index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  const indexCount = 1000;
+
+  const indices = new Uint32Array(indexCount);
+  for (let i = 0; i < indexCount; i++) {
+    indices[i] = i;
+  }
+
+  const effectiveBalanceIncrements = new Uint16Array(indexCount);
+  for (let i = 0; i < indexCount; i++) {
+    effectiveBalanceIncrements[i] = 32 + 32 * (i % 64);
```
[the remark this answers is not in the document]
guha-rahul: yes but basically these are values ranging from 32 to 2048.

Issue #169, opened 2026-01-15
re-organize utils directory


spiral-ladder:
I initially started on this in https://github.com/ChainSafe/lodestar-z/pull/167 but as commits started rolling I figured it might be tough to review due to the sheer volume, even if it's largely a re-org change than a logic change. The changes are also relatively unrelated, so it might be easier to get stuff into main via many small PRs, rather than one large PR.

The (non-exhaustive) list to remove `utils` directory, in the form of commits in that PR:

- [ ] [rm seemingly unused isValidAttestationSlot](https://github.com/ChainSafe/lodestar-z/pull/167/commits/f165c53995035f7dd15b32ebec3dd8206b48f8a9)
- [ ] [mv utils/finality.zig -> penalties.zig](https://github.com/ChainSafe/lodestar-z/pull/167/commits/0193778af2809985b5b6c51f4e27fadbc16b6014) 
- [ ] [mv initializeProposerLookahead into upgrade_state_to_fulu.zig; delete…](https://github.com/ChainSafe/lodestar-z/pull/167/commits/c87cfac3c2a21356a5c5f93818014887d04b80f4) 
- [ ] [utils/root_cache.zig -> cache/RootCache.zig](https://github.com/ChainSafe/lodestar-z/pull/167/commits/258325548924f47693a6feeed5fe337c7f28a9b7)
- [ ] [rm utils/sha256.zig; just use std](https://github.com/ChainSafe/lodestar-z/pull/167/commits/7ebc7083cc724cc76296a05616ae3480a58fd281)
- [ ] [utils/validator.zig -> validator.zig](https://github.com/ChainSafe/lodestar-z/pull/167/commits/bf62f03e5638354faccca148555d6c0a2c84d7a9)
- [ ] [utils/epoch_shuffling -> EpochShuffling.zig](https://github.com/ChainSafe/lodestar-z/pull/167/commits/a75c908b26d2bab9a3b87cc6a350c5d2c19bf784)
- [ ] [utils/epoch.zig -> epoch.zig](https://github.com/ChainSafe/lodestar-z/pull/167/commits/04eab670ff110d8a0dd038d036f7f1de667ca751)
- [ ] [move innerShuffleList from utils/shuffle.zig -> EpochShuffling.zig](https://github.com/ChainSafe/lodestar-z/pull/167/commits/6b473b0ab5eae3772db0a4586bff7f0402386a56)

This issue can be considered done when we **entirely remove the need for a `utils` directory.**

[...]

PR #187, opened 2026-01-22
refactor: move `utils/validator.zig` to top level

spiral-ladder:
part of #169


twoeths (commented):
I'm not super convinced on this, `validator.zig` all contains utils of validators so why not let it stay in `src/state_transition/utils`?
also what's the criteria to move it to the root folder of `src/state_transition`? to me it should only contain main structs/functions to be consumed by other module

nazarhussain:
After discussion with @spiral-ladder we concluded that moving `validator.zig` file from the `utils` is not an immediate value. But we also need to plan reducing the number utils we have, as currently most of the state-transition is under the utils directory, which is not the right maintainable approach. Will plan and open a separate PR for it.

PR #216, opened 2026-02-13
feat: asyncAggregateWithRandomness

wemeetagain:
- Add asyncAggregateWithRandomness, matching chainsafe/blst-ts signature
- ported tests from chainsafe/blst-ts


spiral-ladder (dismissed):
lgtm for now; should prob come back to clean this up at some point

PR #219, opened 2026-02-23
feat: bls batch ops

wemeetagain:
The module (`bindings/napi/bls_batch.zig`) provides BLS batch verification — one synchronous and two asynchronous entry points — plus pool management. All indexed variants resolve public keys by validator index from the shared pubkey cache (`pubkeys.zig`), deserialize and validate signatures on the main thread, then perform the expensive pairing math either inline (sync) or on libuv worker threads (async).

## Verification Functions

A `kind` discriminant selects the set format:

| Kind | Constant | Set fields | Description |
|---|---|---|---|
| `indexed` | `0` | `{index, message, signature}` | Single pubkey looked up by validator index |
| `aggregate` | `1` | `{indices[], message, signature}` | Multiple indices aggregated into one pubkey per set |
| `single` | `2` | `{publicKey, message, signature}` | Raw 48-byte pubkey bytes, no index lookup |

| Function | Signature | Thread |
|---|---|---|
| `verify(kind, sets)` | `(kind, sets) → boolean` | main (blocking) |
| `asyncVerify(kind, sets)` | `(kind, sets) → Promise<boolean>` | worker, returns Promise |
| `asyncVerifySameMessage(sets, message)` | `(sets, message) → Promise<boolean>` | worker, returns Promise |

- **`verify`** batch-verifies using randomized pairing aggregation on the main thread.
- **`asyncVerify`** does the same but dispatches the pairing math to a libuv worker thread.
- **`asyncVerifySameMessage`** is an optimization for attestation-like workloads: it uses `aggregateWithRandomness` (Pippenger multi-scalar multiplication) to combine all pubkeys and signatures with random scalars, then performs a single pairing check. Sets are `{index, signature}` plus a shared message.

## Memory Strategy

**Sync path (`verify`):** All buffers are stack-allocated — `msgs[128]`, `pks[128]`, `sigs[128]`, `rands[128]`. Zero heap allocations on the hot path.

**Async path (`asyncVerify`, `asyncVerifySameMessage`):** A fixed-size pool (`JobPool`) of pre-allocated buffer triples `(msgs, pks, sigs)`, all allocated at init time. On dispatch, a job is popped from the free-list; on completion, pushed back. The hot path does zero allocator calls. Worker threads stack-allocate per-job temporaries (random scalars, Pippenger scratch).

The pool is bounded by `MAX_SETS_PER_JOB = 128` (comptime constant from `blst.MAX_AGGREGATE_PER_JOB`). This is comptime because the same-message Pippenger path requires stack-allocated pointer arrays — heap-allocated pointer arrays crash in blst's assembly under optimized builds.

## Pool Management

- `init(maxJobs)` — pre-allocates `maxJobs` job slots, each sized to 128 sets. Call once at startup.
- `canAcceptWork()` — returns `true` if a free slot is available. JS-side can use this for backpressure before dispatching.
- `deinit()` — frees all pool memory. Not exported to JS; wired into the NAPI environment cleanup hook in `root.zig`.

## Error Handling

All async dispatch functions clean up on failure:
- Pool buffers are pushed back if parsing or queuing fails.
- NAPI async work handles are deleted if `queue()` fails.
- `JobPool.init` tracks partially-allocated slots and frees them on failure.
- Worker thread errors reject the promise with `"BLST_ERROR: Batch verification failed"`.
- Empty sets return `false` (sync) or resolve to `false` (async) immediately.

## TypeScript API

Exported as `blsBatch` from `@chainsafe/lodestar-z/bls-batch`. Types are in `bindings/src/bls-batch.d.ts`.

```typescript
interface BlsBatch {
  readonly indexed: 0;
  readonly aggregate: 1;
  readonly single: 2;

  verify(kind: 0, sets: IndexedSet[]): boolean;
  verify(kind: 1, sets: AggregateSet[]): boolean;
  verify(kind: 2, sets: SingleSet[]): boolean;

  asyncVerify(kind: 0, sets: IndexedSet[]): Promise<boolean>;
  asyncVerify(kind: 1, sets: AggregateSet[]): Promise<boolean>;
  asyncVerify(kind: 2, sets: SingleSet[]): Promise<boolean>;

  asyncVerifySameMessage(sets: SameMessageSet[], message: Uint8Array): Promise<boolean>;

  init(maxJobs: number): void;
  canAcceptWork(): boolean;
}

### CODE REVIEW

File: bindings/napi/bls_batch.zig [hunk 1]
```diff
@@ -0,0 +1,676 @@
...
+
+// ---------------------------------------------------------------------------
+// Init & backpressure
+// ---------------------------------------------------------------------------
+
+/// Pre-allocate the buffer pool.  Call once at startup before dispatching work.
+///   maxJobs — maximum number of concurrent async jobs (= number of buffer slots)
+/// Each slot is sized to MAX_SETS_PER_JOB (128) verification sets.
+pub fn blsBatch_init(env: napi.Env, cb: napi.CallbackInfo(1)) !napi.Value {
+    try pool.init(try cb.arg(0).getValueUint32());
```
nazarhussain: We should check if pool is already initialized and deinit it first.

bindings/napi/bls_batch.zig
```diff
@@ -0,0 +1,522 @@
...
+    return try env.getUndefined();
+}
+
+pub fn blsBatch_canAcceptWork(env: napi.Env, _: napi.CallbackInfo(0)) !napi.Value {
+    return try env.getBoolean(pool.canAcceptWork());
+}
+
+pub fn deinit() void {
+    pool.deinit();
```
GrapeBaBa: How to cancel existing ongoing jobs and release the resource?

bindings/napi/bls_batch.zig
```diff
@@ -0,0 +1,520 @@
...
+        } else {
+            // Incremental aggregation: accumulate in projective coords
+            // one pubkey at a time from the cache — no temp buffer needed.
+            // TODO add this to upstream blst-z?
+            var agg: blst.c.blst_p1 = undefined;
+            blst.c.blst_p1_from_affine(&agg, &(try getPubkey(first_idx)).point);
+            for (1..indices_len) |j| {
+                const idx = try (try indices_val.getElement(@intCast(j))).getValueUint32();
+                blst.c.blst_p1_add_or_double_affine(&agg, &agg, &(try getPubkey(idx)).point);
+            }
```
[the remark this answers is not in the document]
wemeetagain: The reason we can do that in shuffle is that the napi contract / interface is to pass a `TypedArray`.
Here, the contract is to pass a `number[]`.

Look at the lodestar usage, but I'm not sure it makes sense to use a typed array here?

lodekeeper-z (commented):
## Review: feat: bls batch ops

Well-structured module. The split between sync (stack-allocated buffers, main thread) and async (pool-allocated, libuv workers) is clean, and the memory strategy is documented clearly. The Pippenger same-message path with randomness is the right optimization for attestation-heavy workloads. 34 test cases covering all three set kinds + error paths. CI green on core builds.

Several items flagged by Nazar, GrapeBaBa, and Cayman remain unresolved. Summarizing the blockers + adding new observations:

### 🔴 Cancelled async work leaks pending Promises

`asyncComplete` skips `pool.push(data)` on cancellation (correct — pool may be deinitialized), but also skips resolving/rejecting the deferred:

```zig
fn asyncComplete(env: napi.Env, status: napi.status.Status, data: *AsyncJobData) void {
    defer {
        data.work.delete() catch {};
        if (status != .cancelled) pool.push(data);
    }
    // ... only settles deferred below this point
```

If `status == .cancelled`, the function hits the defer and returns — the Promise never settles. This leaves JS callers with a permanently-pending Promise that can't be caught or awaited. The fix: reject the deferred on cancellation before returning.

### 🔴 Double-init of pool (Nazar's unresolved comment)

Calling `blsBatch_init(8)` twice allocates a second pool without freeing the first — memory leak. Options:
1. `deinit()` then `init()` (risky if async jobs are in-flight)
2. Return an error if already initialized
3. No-op if already initialized

Option 2 is safest — it surfaces the programming error. Add: `if (self.slots.len > 0) return error.AlreadyInitialized;` at the top of `init`.

### 🔴 No max bound on `max_jobs` (Nazar's unresolved comment)

Each pool slot pre-allocates ~22 KB (`128 * (32 + 48 + 96)`). A JS caller passing `max_jobs = 1_000_000` would attempt to allocate 22 GB + the `slots`/`stack` arrays. Add a reasonable upper bound (256 as Nazar suggested, or make it configurable with a hard cap).

### 🟡 `queueJob` errdefer resolves instead of rejecting

Cayman already noted this. The errdefer resolves the Promise with `false` on setup failure, which is semantically wrong — the caller can't distinguish "verification failed" from "job setup failed". Should reject with an error. #230 (Nazar's style fixes) addresses this — the errdefer now rejects the promise.

### 🟡 `deserializeSig` / `deserializePubkey` don't validate TypedArray length

`info.data[0..]` passes the entire buffer to `deserialize`. If the buffer is shorter than 96 (sig) or 48 (pk) bytes, `deserialize` reads past the provided data into whatever follows in the backing ArrayBuffer. It'll likely return a deserialization error from invalid point data, but the read itself is out-of-bounds relative to the caller's intent. #230 adds length checks — this should land with or before this PR.

### 🟡 Pairing buffer alignment: `align(32)` vs `align(Pairing.buf_align)`

```zig
var pairing_buf: [Pairing.sizeOf()]u8 align(32) = undefined;
```

Everywhere else in the codebase uses `align(Pairing.buf_align)` (which is 8). `align(32)` is overly conservative but correct — it just wastes stack alignment. Should use `Pairing.buf_align` for consistency.

### 🟡 Compound assert in `batchVerify`

```zig
std.debug.assert(pks.len == n and sigs.len == n and rands.len >= n);
```

Should be 3 separate asserts per TigerStyle — if one fails, you can't tell which invariant broke. (Also flagged by Gemini.)

### Observations (non-blocking)

**1. Stack usage in sync path:** `blsBatch_verify` stack-allocates `MAX_SETS_PER_JOB` arrays: 128 × (32 + 48 + 96 + 8) = 23.5 KB + the pairing buffer. Well within limits.

**2. Pool memory:** At the suggested 256 max_jobs, total pool memory is ~5.6 MB. Reasonable for a validator client.

**3. `parseAggregateSets` direct blst C calls:** The incremental aggregation using `blst_p1_from_affine` / `blst_p1_add_or_double_affine` / `blst_p1_to_affine` bypasses the Zig wrappers. This is fine for performance, but means this code won't benefit from any safety checks added to the Zig layer. Worth a TODO to upstream an incremental aggregate API to blst-z.

**4. Test coverage is solid.** All three set kinds × sync/async × valid/invalid + error handling. Missing: backpressure test (`canAcceptWork` returning false when pool is exhausted), cancellation test (hard to test from JS).

### Relationship to other PRs

- **#230** (Nazar's style fixes) is based on this branch and addresses several items above. Should merge together or in sequence.
- **#235** (native multi-threaded blst) changes the threading model. The async path here uses libuv workers, while #235/#248 introduce a native Zig thread pool in blst-z. These are complementary (NAPI async for JS → libuv, native pool for internal pairing parallelism), but the blst-z dependency version needs to stay in sync.

Overall: solid architecture, good test coverage, needs the safety fixes before merging. Most are addressed by #230 — coordinate merge order. 👍

wemeetagain:
Closing this branch since the current implementation has improved

PR #230, opened 2026-03-04
chore(style): apply style guide fixes to bls_batch

nazarhussain:
- Rename `gpa` to `debug_allocator` (no abbreviations)
- Split compound assertion in batchVerify into 3 separate assertions
- Add precondition assertions to all parse functions (slice len >= n)
- Add bounds assertion to JobPool.push (prevent OOB on double-push)
- Add assertions to sameMessageExecute (n > 0, n <= MAX_SETS_PER_JOB)
- Add input length validation to deserializeSig (96 bytes) and deserializePubkey (48 bytes) before calling into blst
- Guard blsBatch_init against double-init (memory leak) and validate maxJobs bounds (1..256)
- Fix queueJob errdefer to reject promise with error instead of resolving with false (indistinguishable from valid failure)
- Add comment explaining 32-byte vs 8-byte random size difference

### DIFF
--- a/bindings/napi/bls_batch.zig
+++ b/bindings/napi/bls_batch.zig
@@ -44,9 +44,9 @@ const MAX_SETS_PER_JOB = blst.MAX_AGGREGATE_PER_JOB;
 /// MAX_SETS_PER_JOB; verified against the actual blst requirement at pool init.
 const MAX_SCRATCH_SIZE = 128 * MAX_SETS_PER_JOB;
 
-var gpa: std.heap.DebugAllocator(.{}) = .init;
+var debug_allocator: std.heap.DebugAllocator(.{}) = .init;
 const allocator = if (builtin.mode == .Debug)
-    gpa.allocator()
+    debug_allocator.allocator()
 else
     std.heap.c_allocator;
 
@@ -68,14 +68,16 @@ fn getPubkey(index: u32) !*const PublicKey {
 
 fn deserializeSig(sig_value: napi.Value) !Signature {
     const info = try sig_value.getTypedarrayInfo();
-    var sig = Signature.deserialize(info.data[0..]) catch return error.DeserializationFailed;
+    if (info.data.len != 96) return error.InvalidSignatureLength;
+    var sig = Signature.deserialize(info.data[0..96]) catch return error.DeserializationFailed;
     sig.validate(true) catch return error.InvalidSignature;
     return sig;
 }
 
 fn deserializePubkey(pk_value: napi.Value) !PublicKey {
     const info = try pk_value.getTypedarrayInfo();
-    const pk = PublicKey.deserialize(info.data[0..]) catch return error.DeserializationFailed;
+    if (info.data.len != 48) return error.InvalidPublicKeyLength;
+    const pk = PublicKey.deserialize(info.data[0..48]) catch return error.DeserializationFailed;
     pk.validate() catch return error.InvalidPublicKey;
     return pk;
 }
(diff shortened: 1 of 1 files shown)


lodekeeper-z (commented):
## Review: chore(style): apply style guide fixes to bls_batch

Good set of changes. The assertion density improvements, input validation, and the `queueJob` errdefer fix are all solid. The rename from `gpa` to `debug_allocator` is a clear clarity win. A few observations:

### What's good

1. **`deserializeSig` / `deserializePubkey` length checks** — Checking `info.data.len` before slicing is a real correctness fix, not just style. Without it, a JS caller passing a 47-byte TypedArray would get a slice of 48 or 96 bytes reading past the buffer. The explicit `[0..96]` / `[0..48]` slice eliminates the ambiguity of `[0..]` (which was the entire buffer, potentially larger than needed). Clean.

2. **Split compound assert in `batchVerify`** — Breaking `assert(a and b and c)` into three separate asserts means a failure pinpoints which invariant broke. This is exactly what TigerStyle advocates.

3. **Precondition asserts on all parse functions** — `parseIndexedSets`, `parseAggregateSets`, `parseSingleSets`, `parseSameMessageSets` now all assert their buffer lengths upfront. Consistent and defensive.

4. **`queueJob` errdefer: reject instead of resolve(false)** — The old code resolved the promise with `false` on setup failure, which is semantically wrong — the caller can't distinguish "verification failed" from "job didn't even run". Rejecting with an error message is the correct JS contract.

5. **`sameMessageExecute` bounds asserts + comment** — `assert(n > 0)` and `assert(n <= MAX_SETS_PER_JOB)` are good guards before the stack-allocated arrays. The comment explaining why 32-byte randoms (vs 8-byte in `batchVerify`) is helpful context.

### Issue: partial `init` failure + `AlreadyInitialized` guard

Gemini flagged this and it's valid. If `pool.init(max_jobs)` fails partway through (e.g., OOM on the third slot's `sigs` allocation), the `errdefer` chain in `init` frees `self.slots` and `self.stack` — but doesn't reset them to empty slices. So after a failed init:
- `pool.slots` is a dangling pointer with `pool.slots.len > 0`
- The new `if (pool.slots.len > 0) return error.AlreadyInitialized` prevents retry
- A subsequent `deinit()` would double-free

Fix: add `self.* = .{};` as the outermost errdefer in `init`, or (simpler) use a separate `initialized: bool = false` flag instead of checking `slots.len`. The cleanest approach:

```zig
fn init(self: *JobPool, max_jobs: usize) !void {
    errdefer self.* = .{};  // Reset to zero-state on any failure
    // ... rest unchanged
}
```

This ensures `deinit` on a failed-init pool is a no-op (empty slices → nothing to free), and re-init is possible.

### Minor

- **`MAX_POOL_JOBS = 256`** — Reasonable upper bound. Each job pre-allocates ~170 KB (32×128 + 48×128 + 96×128), so 256 jobs ≈ 43 MB. Worth a comment noting the memory implication.

- **`JobPool.push` assert** — Gemini notes this is debug-only. True, but the push/pop pattern here is bounded by the pool size and only called from the NAPI completion callback (single-threaded JS context), so a logic bug triggering double-push would be a programming error, not a runtime condition. `debug.assert` is appropriate per TigerStyle — it's a programmer invariant, not a user input. The existing `pop` already returns `?*AsyncJobData` (nullable) for the runtime check. Consistent.

Overall: clean, focused, correct. The init guard is the only actionable item. 👍

bindings/napi/bls_batch.zig
```diff
@@ -122,6 +126,9 @@ fn batchVerify(
 
 /// Parse {index, message, signature} sets.
 fn parseIndexedSets(sets: napi.Value, n: usize, msgs: [][32]u8, pks: []PublicKey, sigs: []Signature) !void {
+    std.debug.assert(msgs.len >= n);
+    std.debug.assert(pks.len >= n);
+    std.debug.assert(sigs.len >= n);
```
spiral-ladder, line 131: Q: why do we want `>=` here instead of exactly `==`?

lodekeeper-z: `>=` is the minimal precondition for the parser helpers: they only write `0..n`, so a caller may safely pass a larger scratch buffer. In this PR the current call sites already pass `data.*[0..n]`, so `==` would also hold there; I’d use `==` only if we want the helper contract to require already-trimmed slices rather than just enough capacity.

wemeetagain:
- closing since #219 is closed

PR #247, opened 2026-03-17
feat: add StructContainerTreeView, model Validator as StructContainerType

wemeetagain:
## Summary

Replaces PR #232. Adds `StructContainerTreeView` — a view-level optimization for fixed-size containers that stores the value as a flat Zig struct instead of decomposing it into the merkle tree.

**The optimization:** fields are read and written in O(1) via direct struct field access. Dirty fields are tracked with a `StaticBitSet`. On commit, only changed fields are spliced into the existing merkle tree via `setNodesAtDepth` — changing 1 field out of 8 creates ~3 pool nodes instead of ~15. `serializeIntoBytes` and `toValue` read directly from the struct without committing.

**The pool stays pure.** No special node types, no type erasure, no pointer stuffing. This is entirely a view-layer optimization. The tree in the pool is a normal merkle tree built by `FixedContainerType.tree.fromValue`.

### Design

- `StructContainerTreeView` stores `value: T` (owned flat struct) + `changed: StaticBitSet(N)` (1 byte for Validator's 8 fields)
- `get` is a single `@field(self.value, name)` — no tree traversal, no optional unwrap
- `set` writes to the struct and sets a bit — skips if value unchanged
- `commit` iterates set bits, creates leaf nodes only for dirty fields, splices via `setNodesAtDepth`
- `StructContainerType` is `FixedContainerTypeWith(ST, .struct_container)` — 3 lines, no delegation boilerplate

### What changed

| File | Change |
|------|--------|
| `src/ssz/type/container.zig` | Parameterize `FixedContainerType` with `TreeViewType` enum; add `StructContainerType` |
| `src/ssz/tree_view/container.zig` | Add `StructContainerTreeView` with tests |
| `src/consensus_types/phase0.zig` | Validator → `StructContainerType` |
| `src/state_transition/test_utils/generate_state.zig` | Validator field access uses struct semantics |

## Test plan

- [x] `zig build test:ssz` — all tests pass including:
  - Basic get/set/commit roundtrip
  - Partial update produces correct merkle root
  - Clone isolates changes
  - Repeated commits don't leak pool nodes (tight 64-slot pool)
  - Double commit is idempotent
  - Deinit with uncommitted changes
  - fromValue roundtrip
- [ ] `zig build test:spec_tests -Dpreset=minimal`

### DIFF
--- a/src/consensus_types/phase0.zig
+++ b/src/consensus_types/phase0.zig
@@ -19,7 +19,7 @@ pub const Checkpoint = ssz.FixedContainerType(struct {
     root: p.Root,
 });
 
-pub const Validator = ssz.FixedContainerType(struct {
+pub const Validator = ssz.StructContainerType(struct {
     pubkey: p.BLSPubkey,
     withdrawal_credentials: p.Root,
     effective_balance: p.Gwei,
--- a/src/ssz/root.zig
+++ b/src/ssz/root.zig
@@ -33,6 +33,7 @@ pub const VariableVectorType = types.VariableVectorType;
 
 pub const FixedContainerType = types.FixedContainerType;
 pub const VariableContainerType = types.VariableContainerType;
+pub const StructContainerType = types.StructContainerType;
 
 pub const getPathGindex = types.getPathGindex;
 
@@ -42,6 +43,7 @@ pub const HasherData = hasher.HasherData;
 
 const tree_view = @import("tree_view/root.zig");
 pub const ContainerTreeView = tree_view.ContainerTreeView;
+pub const StructContainerTreeView = tree_view.StructContainerTreeView;
 pub const ArrayBasicTreeView = tree_view.ArrayBasicTreeView;
 pub const ArrayCompositeTreeView = tree_view.ArrayCompositeTreeView;
 pub const ListBasicTreeView = tree_view.ListBasicTreeView;
--- a/src/ssz/tree_view/container.zig
+++ b/src/ssz/tree_view/container.zig
@@ -427,6 +427,438 @@ pub fn ContainerTreeView(comptime ST: type) type {
     return TreeView;
 }
 
+pub fn StructContainerTreeView(comptime ST: type) type {
+    const T = ST.Type;
+
+    const TreeView = struct {
+        allocator: Allocator,
+        pool: *Node.Pool,
+        root: Node.Id,
+        value: T,
+        changed: std.StaticBitSet(ST.chunk_count),
+
+        pub const SszType = ST;
+
+        const Self = @This();
+
+        pub fn init(allocator: Allocator, pool: *Node.Pool, root: Node.Id) !*Self {
+            try pool.ref(root);
+            errdefer pool.unref(root);
+
+            const ptr = try allocator.create(Self);
+            errdefer allocator.destroy(ptr);
+
+            try ST.tree.toValue(root, pool, &ptr.value);
+
+            ptr.allocator = allocator;
+            ptr.pool = pool;
+            ptr.root = root;
+            ptr.changed = std.StaticBitSet(ST.chunk_count).initEmpty();
+            return ptr;
+        }
+
+        /// Clone creates a new view sharing the same committed state.
+        /// Uncommitted changes are committed first so the clone sees the current state.
+        pub fn clone(self: *Self, opts: CloneOpts) !*Self {
+            _ = opts;
+            try self.commit();
+
+            try self.pool.ref(self.root);
+            errdefer self.pool.unref(self.root);
+
+            const ptr = try self.allocator.create(Self);
+            ptr.* = .{
+                .allocator = self.allocator,
+                .pool = self.pool,
+                .root = self.root,
+                .value = self.value,
+                .changed = std.StaticBitSet(ST.chunk_count).initEmpty(),
+            };
+            return ptr;
+        }
+
+        pub fn deinit(self: *Self) void {
+            self.pool.unref(self.root);
+            self.allocator.destroy(self);
+        }
+
+        pub fn commit(self: *Self) !void {
+            if (self.changed.count() == 0) {
+                return;
+            }
+
+            // Phase 1: Create leaf nodes for changed fields.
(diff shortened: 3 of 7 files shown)


twoeths (commented):
@wemeetagain the performance of `before_process_epoch` is almost the same to main
this is due to having to deserialize value from a node https://github.com/ChainSafe/lodestar-z/blob/ce6370e75b980d0657ac246083b437e76fcd1e4b/src/ssz/tree_view/chunks.zig#L335
while with #232 you just get it from the pointer of node https://github.com/ChainSafe/lodestar-z/pull/232/changes#diff-2320e93e35c98f65b4cef2c729d8729fda9c280365baa569c2bf6c961230d745R839

lodekeeper-z (approved):
## Review: feat: add StructContainerTreeView, model Validator as StructContainerType

Clean design. The core insight — caching the full struct value alongside the merkle tree and using a `StaticBitSet` dirty tracker to splice only changed fields on commit — gives O(1) field access without touching the pool's node model. This is a strictly view-level optimization: no new node types, no pointer encoding in left/right slots, no `BranchStructRef` indirection. That's a meaningful simplification over the competing approach in #232.

### What's good

1. **Zero pool coupling.** The pool doesn't know or care about `StructContainerTreeView`. It sees normal leaf/branch nodes. This means every existing operation — `setNodesAtDepth`, `getNodeAtDepth`, `clone`, serialization — works unchanged. No new node lifecycle to reason about. No type-erased vtable stored in pointer-encoded left/right slots.

2. **`StructContainerType` via `TreeViewType` enum.** The parameterization of `FixedContainerTypeWith` to select between `ContainerTreeView` and `StructContainerTreeView` is minimal and clean. The SSZ type layer is unchanged — `StructContainerType` delegates everything to `FixedContainerType` except which tree view to use.

3. **Commit is well-structured.** The three-phase approach (create leaf nodes → splice into tree → update root ownership) with `errdefer` cleanup and the `changed_idx = 0` disarm trick is correct. The ownership chain is clear: nodes start at refcount 0, `setNodesAtDepth` takes ownership via rebind, then we ref the new root and unref the old.

4. **`set()` skip-if-unchanged.** The byte-level equality check in `set()` avoids dirtying unchanged fields. Smart for `process_epoch` where many validators get "written" with the same value.

5. **Test suite.** Covers the important cases — partial update correctness, clone isolation, pool node leak detection (tight pool test), double-commit idempotency, deinit with uncommitted changes, fromValue roundtrip.

### Observations

**1. `toValue` skips commit — is this intentional?**

```zig
pub fn toValue(self: *Self, allocator: Allocator, out: *ST.Type) !void {
    _ = allocator;
    out.* = self.value;
}
```

This returns the in-memory value directly, which includes uncommitted changes (since `set()` writes to `self.value` immediately). That's actually fine — the struct is the source of truth, and uncommitted just means the *tree* hasn't been updated yet. But `serializeIntoBytes` also reads from `self.value` without committing, so serialization of a view with uncommitted changes gives the dirty value, not the tree's value. This is consistent but worth a doc comment clarifying that `toValue`/`serializeIntoBytes` reflect uncommitted state while `hashTreeRoot`/`getFieldRoot` commit first.

**2. Memory size per view: ~`sizeof(T)` + bitset overhead**

For `Validator` (121 bytes as SSZ, struct is 129 bytes with Zig padding), each `StructContainerTreeView` carries the full struct. With ~1M validators in a `ListCompositeTreeView`, how does this interact? Each `get()` call on the list creates a new `StructContainerTreeView` (via `init`) which does `toValue` (reads 8 fields from the tree). If `process_epoch` accesses many validators, we're paying for the tree readback on every list access.

I suspect the real perf win comes from combining this with twoeths's observation: skipping the per-field deserialization by reading directly from a `branch_struct` node. Your approach is a stepping stone — the view API is right, but the tree-backed `init` path will need the pointer-in-node optimization from #232 for the hot loop. Is that the intended roadmap? #244 ("Cayman/232 review") suggests you're working on reconciling the two approaches.

**3. `fromValue` double-builds the tree**

The inlined `fromValue` builds the full merkle tree via `ST.tree.fromValue(pool, value)` (creating all 8 leaf nodes + branch nodes), then stores `value.*` in the struct. On the next `commit()`, if any field is dirty, it creates *new* leaf nodes and splices them in. The initial tree nodes for those fields are then unreachable and freed. Not a bug, but if `fromValue` is followed immediately by mutations (common pattern: create validator, set fields), the initial tree work for the mutated fields is wasted. An optimization for later: `fromValue` could skip tree building entirely and defer it to the first `commit()`.

**4. Devil's advocate: when does this approach lose to #232?**

#232 stores the struct pointer directly in the node pool (via `branch_struct` node type), so any code that navigates the tree to the validator node gets O(1) access to the struct *without creating a view*. In your approach, every access requires instantiating a `StructContainerTreeView`, doing the tree readback, and heap-allocating the view. For one-off reads ("get validator N's effective balance"), #232 wins.

Your approach wins on simplicity, correctness, and non-invasiveness. The pool stays clean. But for `process_epoch` with ~1M validators, the per-access overhead matters. The two approaches may actually compose: use #232's `branch_struct` for storage + your `StructContainerTreeView` API for mutations. Worth discussing.

### Nits

- Several lines exceed 100 columns (the `set()` byte comparison, `commit()` field iteration). Gemini flagged these — worth wrapping for style guide compliance.
- `toValue` ignores `allocator` param. Consider documenting why (or removing it if the interface allows — though `assertTreeViewType` probably requires it).

### vs #232

For anyone following both PRs: #247 is the simpler, safer approach (view-level only, no pool changes). #232 is more invasive (new `branch_struct` node type, pointer encoding, type-erased vtable in the pool) but gives O(1) access without view instantiation. twoeths's benchmark comment on #247 suggests performance is "almost the same as main" for `before_process_epoch` because the `init` path still deserializes from the tree — this confirms the readback cost needs addressing regardless of which approach lands.

I'd suggest landing #247 first (it's simpler, correct, well-tested, and the view API is right), then layering #232's `branch_struct` storage optimization underneath if benchmarks justify it.

Overall: clean, correct, well-tested. 👍

wemeetagain:
- closing in favor of #346

PR #248, opened 2026-03-18
feat(bls): switch multi-threading to a queue-based dispatch

spiral-ladder:
Prereq of https://github.com/ChainSafe/lodestar/pull/8900, deployed today

## Results

- Numbers on grafana seem good now, no more CPU overusage, no more retries
- state transition metrics seem normal versus unstable now
- CI numbers on the upstream consuming branch seems fair

With the older model of concurrency, the `dispatch_mutex` was causing contention between the different workers. I separately attempted to deploy feat3 with a [single worker version for a while](https://github.com/ChainSafe/lodestar/pull/8900/changes/f6a6b160d9240477bdf2ff1dd5eac6c796c67cde) that seems to confirm this, so it was a contention issue rather than any correctness issue with our bls implementation.

The new model uses a queue-based system using `JobQueue` with each `WorkItem` owning its own data `PairingBuf`, etc. This is currently running on `feat3` with no issues so far.

### DIFF
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -77,7 +77,9 @@ const WorkItem = struct {
 /// Creates a thread pool with the specified number of workers.
 /// The caller owns the returned pool and must call `deinit` when done.
 pub fn init(allocator_: Allocator, opts: Opts) (Allocator.Error || std.Thread.SpawnError)!*ThreadPool {
-    std.debug.assert(opts.n_workers >= 1 and opts.n_workers <= MAX_WORKERS);
+    std.debug.assert(opts.n_workers >= 1);
+    std.debug.assert(opts.n_workers <= MAX_WORKERS);
+
     const pool = try allocator_.create(ThreadPool);
     pool.* = .{
         .allocator = allocator_,
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -107,21 +107,18 @@ pub fn deinit(pool: *ThreadPool) void {
 
 fn workerLoop(pool: *ThreadPool) void {
     while (true) {
-        var item: *WorkItem = undefined;
-
-        {
+        const item: *WorkItem = blk: {
             pool.queue.mutex.lock();
             defer pool.queue.mutex.unlock();
 
             while (true) {
                 if (pool.shutdown.load(.acquire)) return;
                 if (pool.queue.pop()) |wi| {
-                    item = wi;
-                    break;
+                    break :blk wi;
                 }
                 pool.queue.cond.wait(&pool.queue.mutex);
             }
-        }
+        };
 
         item.exec_fn(item);
         item.done.set();
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -2,6 +2,12 @@
 //!
 //! Provides multi-threaded versions of aggregation and verification functions
 //! using a persistent pool of worker threads to avoid thread creation overhead.
+//!
+//! Multiple callers can dispatch work concurrently. Each job owns its own
+//! pairing buffers. Workers pull work items from a shared queue and use atomic
+//! counters within each job to grab individual signature sets to process,
+//! similar to how the Rust `blst` crate's `verify_multiple_aggregate_signatures`
+//! works with `threadpool::ThreadPool`.
 const ThreadPool = @This();
 
 const std = @import("std");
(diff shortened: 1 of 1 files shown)


twoeths (commented):
I'd love to have a benchmark for this to prove that it's better than what we currently have
need to run on one of our servers to make sure

lodekeeper-z (approved):
## Review: feat(bls): switch multi-threading to a queue-based dispatch

Strong improvement. The fundamental change — replacing the per-worker slot assignment + `dispatch_mutex` with a shared job queue — eliminates the serialization bottleneck for concurrent callers. The old design forced all `verifyMultipleAggregateSignatures` / `aggregateVerify` calls through a single mutex, meaning a second caller blocked entirely until the first finished dispatching + collecting results. Now multiple callers can submit work concurrently, each with their own job state and result buffers, while workers pull from a shared queue. This is the right architecture for a BLS verification pool that will serve multiple concurrent validation pipelines.

### What's good

1. **`WorkItem` + `@fieldParentPtr` pattern** — Clean type erasure. The queue only knows about `WorkItem` (function pointer + done event + next pointer), and each job type (`VerifyMultiWorkItem`, `AggVerifyWorkItem`) embeds it. Workers call `exec_fn` without knowing the concrete type. This is exactly how you'd do vtable dispatch in Zig.

2. **Stack-local pairing buffers in workers** — Each worker's `exec` function allocates a `PairingBuf` on its own stack, does work, commits, then copies to `result_bufs[slot]`. This eliminates the old shared `pool.pairing_bufs` array and the `pool.has_work` tracking. Cleaner ownership semantics.

3. **Simplified `mergeAndVerify`** — Now takes a contiguous slice of result buffers (`result_bufs[0..n_results]`) instead of scanning `pool.has_work[]` with gaps. No more searching for the first worker that actually did work. Linear merge over a dense array.

4. **`workerLoop` using `Condition.wait`** — Workers now block on a condition variable instead of polling individual `ResetEvent`s. More idiomatic, and naturally handles the shutdown signal.

### Observations (non-blocking)

**1. `result_bufs: [MAX_WORKERS]PairingBuf` on the caller's stack**

Each `PairingBuf` is ~3.2 KB (blst pairing context), so `[MAX_WORKERS]PairingBuf` = 16 × 3.2 KB ≈ 51 KB on the stack of every call to `verifyMultipleAggregateSignatures` or `aggregateVerify`. Plus the `work_items` and `item_ptrs` arrays. Default thread stacks are 8 MB on Linux so this won't blow, but it's worth knowing — especially if this ever gets called from a context with reduced stack size (e.g., Zig's default `std.Thread` stack of 16 MB is fine, but some embed contexts limit it).

Since only `n_active` slots are used, you could allocate `n_active` PairingBufs dynamically, but the current approach is simpler and avoids allocation failure paths. Acceptable tradeoff.

**2. `PairingBuf` copy from worker stack to result array**

Workers do `job.result_bufs[slot] = buf;` — a ~3.2 KB memcpy per worker. This is negligible compared to the actual pairing computation (which is O(ms)), but worth noting. An alternative would be workers allocating directly into the result slot, but that would require knowing the slot upfront (before doing work), which conflicts with the "atomic grab next index" pattern.

**3. `pushBatch` builds a LIFO stack**

Items are linked as `item.next = self.head; self.head = item;` — so the last item pushed is the first popped. This means workers that wake up first get the last-pushed items. Doesn't matter for correctness (work distribution is via `counter.fetchAdd` anyway), but FIFO ordering would be more conventional. Could use a tail pointer. Not worth changing.

**4. Concurrency correctness**

The synchronization chain is sound: worker writes `result_bufs[slot]` → worker calls `item.done.set()` → main thread's `item.done.wait()` returns → main reads `result_count` and `result_bufs`. The `ResetEvent.set()`/`wait()` pair provides the necessary happens-before relationship. The `fetchAdd(.acq_rel)` on `result_count` gives unique slots.

### Devil's advocate: thread count = MAX_WORKERS

The old code spawned `n_workers - 1` background threads (worker 0 was the calling thread). The new code spawns `n_workers` threads and the calling thread never participates in work execution — it just submits and waits. This means:
- Old: `init(.{ .n_workers = 4 })` → 3 OS threads + calling thread = 4 workers
- New: `init(.{ .n_workers = 4 })` → 4 OS threads, calling thread blocked in `submitAndWait`

So for the same `n_workers` value, you now have one extra OS thread and the calling thread is idle during verification. For a server workload where the caller has other work to do asynchronously, this is fine. For a synchronous call pattern where the caller just waits, you're wasting the calling thread's core. In practice, BLS verification is almost always async in a consensus client (work is dispatched from an event loop), so this is probably the right tradeoff. Just documenting the behavioral change.

Overall: clean, correct, better architecture. 👍

Issue #10, opened 2025-05-15
Support Windows


twoeths:
as monitored in https://github.com/ChainSafe/state-transition-z/actions/runs/15014083017/job/42188053404
on Windows, `state-transition-utils.dll` was not produced, only `state-transition-utils.lib` instead

```
Run zig build -Dtarget=x86_64-windows-msvc -Doptimize=ReleaseSafe
  zig build -Dtarget=x86_64-windows-msvc -Doptimize=ReleaseSafe
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    ZIG_LOCAL_CACHE_DIR: C:\Users\runneradmin\AppData\Local\zig

Run dir zig-out\lib
  dir zig-out\lib
  shell: C:\Windows\system3[2](https://github.com/ChainSafe/state-transition-z/actions/runs/15014083017/job/42188053404#step:6:2)\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    ZIG_LOCAL_CACHE_DIR: C:\Users\runneradmin\AppData\Local\zig
 Volume in drive D is Temporary Storage
 Volume Serial Number is 0629-A8E1
 Directory of D:\a\state-transition-z\state-transition-z\zig-out\lib
05/14/2025  06:47 AM    <DIR>          .
05/14/2025  06:47 AM    <DIR>          ..
05/14/2025  0[3](https://github.com/ChainSafe/state-transition-z/actions/runs/15014083017/job/42188053404#step:6:3):48 AM            17,236 state-transition-utils.lib
05/14/2025  03:48 AM         1,267,206 state-transition.lib
```

hence bun cannot load its binding

PR #266, opened 2026-03-24
fix: env usage

spiral-ladder:
depends on https://github.com/ChainSafe/zapi/pull/10/

### DIFF
--- a/bindings/napi/BeaconStateView.zig
+++ b/bindings/napi/BeaconStateView.zig
@@ -31,11 +31,12 @@ pub fn BeaconStateView_ctor(env: napi.Env, cb: napi.CallbackInfo(0)) !napi.Value
     const cached_state = try allocator.create(CachedBeaconState);
     errdefer allocator.destroy(cached_state);
 
-    _ = try env.wrap(
+    try env.wrap(
         cb.this(),
         CachedBeaconState,
         cached_state,
         BeaconStateView_finalize,
+        null,
         null,
     );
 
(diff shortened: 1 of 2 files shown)


lodekeeper-z (commented):
The `bindings - test` CI is failing because this PR updates call sites to pass 6 args to `env.wrap()` (adding the trailing `null` for the ref parameter), but the zapi dependency in `build.zig.zon` still points to the old version where `wrap()` only accepts 5 args.

This needs the zapi `bing/null-ref` branch pinned in `build.zig.zon` (same as PR #253 does):
```zig
.zapi = .{
    .url = "git+https://github.com/chainsafe/zapi?ref=bing/null-ref#fff894d0e045e20cc54c574ae83a39f310574486",
    .hash = "napi_z-0.1.0-iqJb71A-AgD71vDhkJJ3E6BgFRYOTL1loErZREgSFmm8",
},
```

Also — this PR and #253 overlap on the same `env.wrap()` call sites. Might make sense to fold this fix into #253 or coordinate merge order.

lodekeeper-z (approved):
LGTM. Clean API update matching the zapi #10 Ref refactor (now merged).

Changes are mechanical:
- `env.wrap()` return type changed from `!napi.Value` to `!void` → drop the `_ = try` pattern
- Extra `null` param for the new `ref` parameter

Core build & tests all pass. Only `bindings - test` CI fails as expected until zapi dependency is updated in `zbuild.zon`.

spiral-ladder:
env usage is already fixed in main. Closing

PR #269, opened 2026-03-25
feat(stf): zero-alloc `processSyncAggregate`

spiral-ladder:
Introduces a zero-alloc `processSyncAggregate`.

Done by:

separating `intersectValues` in `BitVector` into:

- `intersectValuesAlloc`: the original `intersectValues`.
- `intersectValues`: a new function that takes a `out: [length]T` as a param.

The maximum size of the `BitVector` is 512, which is negligible on the
stack. We introduce an alloc-free version as a minor optimization.

Then since we know the pubkeys are going to be 512 * 96 = 48KiB worth of allocations,
we can skip heap allocations and simply use an array for `participant_buf`.

### DIFF
--- a/bench/state_transition/process_block.zig
+++ b/bench/state_transition/process_block.zig
@@ -215,7 +215,6 @@ fn ProcessSyncAggregateBench(comptime fork: ForkSeq, comptime opts: BenchOpts) t
 
             state_transition.processSyncAggregate(
                 fork,
-                allocator,
                 cloned.config,
                 cloned.epoch_cache,
                 cloned.state.castToFork(fork),
--- a/src/ssz/type/bit_vector.zig
+++ b/src/ssz/type/bit_vector.zig
@@ -126,7 +126,7 @@ pub fn BitVector(comptime _length: comptime_int) type {
         /// Allocates and returns an `ArrayList` of indices where the bit at the index of `self` is set to `true`.
         ///
         /// Caller must call `deinit` on the returned list
-        pub fn intersectValues(
+        pub fn intersectValuesAlloc(
             self: *const @This(),
             comptime T: type,
             allocator: std.mem.Allocator,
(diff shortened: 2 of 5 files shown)


lodekeeper-z (approved):
Clean optimization — stack-allocated buffers for sync aggregate processing. The new `intersectValues` correctly mirrors the existing Kernighan's algorithm logic.

One minor observation: `getSyncCommitteeSignatureSet` still heap-allocates `pubkeys` (line 161) — could be stack-allocated too for consistency, but fine as a follow-up.

48KB total stack usage (4KB indices + ~24KB pubkeys in `processSyncAggregate`, + 4KB in `getSyncCommitteeSignatureSet`) is well within safe limits.

lodekeeper-z (commented):
CI is failing — there's a second `processSyncAggregate` call site in `bench/state_transition/process_block.zig:421` that still passes `allocator`. Needs the same fix as the one at line 218.

```
bench/state_transition/process_block.zig:421:33: error: expected 6 argument(s), found 7
```

Also: `getSyncCommitteeSignatureSet` still takes `allocator` as a param (line 117 in the diff) but no longer uses it for `intersectValues` — worth removing it from that function's signature too for consistency? Though it might still be used for other things in that function, haven't checked.

lodekeeper-z (commented):
CI is failing — there's a second `processSyncAggregate` call site in `bench/state_transition/process_block.zig:421` that still passes `allocator`. The one at line 215 was updated but this one was missed.

```diff
-                state_transition.processSyncAggregate(
-                    fork,
-                    allocator,
-                    cloned.config,
+                state_transition.processSyncAggregate(
+                    fork,
+                    cloned.config,
```

Otherwise the PR looks good — nice zero-alloc optimization 👍

lodekeeper-z (approved):
Clean change. The zero-alloc approach with stack buffers is correct — `SYNC_COMMITTEE_SIZE` is compile-time known and fits comfortably on the stack.

Nice side-effect: this also fixes a memory leak in `getSyncCommitteeSignatureSet` where the old `intersectValues` returned an `ArrayList` whose `.items` was extracted but the list itself was never `deinit`'d.

Kernighan's bit iteration in the new `intersectValues` matches the existing allocating version — consistent behavior. 👍

lodekeeper-z (approved):
Clean PR. Stack-allocated buffers for sync committee processing — eliminates heap allocations for both `participant_indices` and `pubkeys`.

Nice side-effect: fixes a memory leak in `getSyncCommitteeSignatureSet` where the old code did `(try ...intersectValues(...)).items` which took the backing slice but never deinitialized the ArrayList.

Kernighan's bit trick in the new `intersectValues` is a good touch — skips zero bytes efficiently.

LGTM ✅

lodekeeper-z:
CI is failing on all 3 platforms — the bench file `bench/state_transition/process_block.zig:421` still calls `processSyncAggregate` with 7 args (now expects 6 after your refactor). Looks like just the benchmark needs updating to match the new signature.

matthewkeil:
@spiral-ladder is this still relevant with all the other great work youve done on blst?

spiral-ladder:
@matthewkeil i think it is but low prio, since it's just an optimization for state transition

one thing to discuss here is if its worth the change in the ssz api for this

GrapeBaBa (approved):
LGTM

wemeetagain (approved):

PR #277, opened 2026-03-25
feat(persistent_merkle_tree): add batchGetRoot for 2x faster tree hashing

lodekeeper-z:
## Summary

Add `batchGetRoot()` to `Node.Id` — a BFS level-ordered batch commit that leverages hashtree's AVX/SSE batch hashing instead of recursive single-pair `hashOne()` calls.

## Algorithm

1. **BFS collect**: Walk from root, collecting dirty branch nodes per level
2. **Bottom-up batch hash**: Process levels from deepest to shallowest using batch `hash()`
3. **Write back**: Update hash and state for each processed node

Three wins stack: hashtree AVX batching (27%), no recursion overhead, and cache locality.

## Benchmark Results

| Tree | Recursive | Batch | Speedup |
|------|-----------|-------|---------|
| 500K leaves, 5% dirty | 16.8ms | 8.5ms | **2.0x** |
| 500K leaves, 95% dirty | 80ms | 39ms | **2.1x** |
| 1M leaves, 95% dirty | 210ms | 85ms | **2.5x** |

## Impact

The hidden `hashTreeRoot` cost in epoch processing is ~96ms on mainnet (2.18M validators). This optimization brings it down to ~40ms, saving **~56ms/epoch**.

## Next Steps

- Integrate into epoch commit path (replace `getRoot()` calls with `batchGetRoot()` where depth is known)
- End-to-end epoch benchmark

### DIFF
--- a/src/persistent_merkle_tree/Node.zig
+++ b/src/persistent_merkle_tree/Node.zig
@@ -3,6 +3,7 @@ const std = @import("std");
 const Allocator = std.mem.Allocator;
 
 const hashOne = @import("hashing").hashOne;
+const hashBatch = @import("hashing").hash;
 const getZeroHash = @import("hashing").getZeroHash;
 const max_depth = @import("hashing").max_depth;
 const Depth = @import("hashing").Depth;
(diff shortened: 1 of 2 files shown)


src/persistent_merkle_tree/Node.zig
```diff
@@ -413,6 +414,95 @@ pub const Id = enum(u32) {
...
+            return &hashes[@intFromEnum(node_id)];
+        }
+
+        const lefts = pool.nodes.items(.left);
+        const rights = pool.nodes.items(.right);
+
+        // Collect dirty branch nodes per level via BFS.
+        // levels[i] holds dirty node IDs at depth i (0 = root level).
+        var levels: [max_depth]std.ArrayListUnmanaged(Id) = undefined;
+        for (0..depth) |i| {
```
GrapeBaBa: could use `memset` instead

wemeetagain:
> Integrate into epoch commit path (replace getRoot() calls with batchGetRoot() where depth is known)

Please include this in this PR.

> End-to-end epoch benchmark

We have this. Please run the benchmark before/after to see what kind of speedup of epoch transition we get.

lodekeeper-z:
I tested this against current `main` before pushing an integration update.

The PMT internals have changed since this PR was opened (`root`/packed payload/node kinds and shared DAGs), so the existing `batchGetRoot` no longer compiles after merging `main`. A direct port also produced incorrect roots for nested container fields. Making the traversal safe for heterogeneous trees and shared subtrees removed the original advantage in the existing end-to-end benchmark.

Fulu state, 2,178,249 validators, 50 iterations, ReleaseFast:

| metric | current main + PR baseline | DAG-safe batch integration | change |
|---|---:|---:|---:|
| `state_root` | 30.006 ms | 31.265 ms | +4.2% |
| segmented epoch total | 123.553 ms | 124.141 ms | +0.5% |

I have not pushed the regression. The older level-only traversal measured roughly neutral here, but it can grow exponentially on a valid shared-subtree DAG and exhaust memory.

Would you prefer that I refresh this PR as a primitive-only change against the current PMT, or pursue a redesigned allocation-reusing batch commit before integrating it into container/state hashing?

wemeetagain:
you can try out both approaches and see if either make sense

lodekeeper-z:
I tried both directions against current `main`.

**Primitive-only refresh:** the API can be made correct for heterogeneous trees and shared DAGs, but it has no production caller, so it provides no epoch/state-root benefit by itself. The current safe implementation also needs substantially more graph/scratch machinery than the original depth-indexed version.

**Allocation-reusing integration:** I replaced the per-call allocations and repeated readiness scans with pool-owned retained scratch plus a linear Kahn traversal. Focused PMT and container tests pass, including unequal-depth sharing, duplicate edges, depth bounds, OOM retry, and warm no-allocation reuse.

The end-to-end result is still negative for the target operation (Fulu state, 2,178,249 validators, 50 iterations, ReleaseFast):

| metric | current `main` | allocation-reusing integration | change |
|---|---:|---:|---:|
| `state_root` | 30.369 ms | 30.963 ms | +1.96% |
| segmented epoch total | 123.689 ms | 122.413 ms | -1.03% (noise-dominated) |

The state-root regression is consistent with the prior DAG-safe experiment, and the integration also adds a cold-path allocation/OOM point plus retained per-pool scratch. I therefore did not push either version: the primitive-only form has no demonstrated consumer benefit, while the integrated form makes the operation we are targeting slower.

wemeetagain:
Ok closing then

PR #282, opened 2026-03-27
perf: use saturating arithmetic in balance updates for SIMD vectorization

lodekeeper-z:
## Current status — historical claims corrected

The measurements below are historical ReleaseFast results, not evidence of SIMD vectorization or an end-to-end gain on the current ReleaseSafe build. The assertion-bearing head has not demonstrated a ReleaseSafe improvement. [Latest maintainer measurements](https://github.com/ChainSafe/lodestar-z/pull/282#discussion_r3988193013) report a benefit for a different variant with the assertions removed; that variant is not this PR head.

The earlier “zero risk of behavior change” claim is withdrawn. Replacing `try std.math.add` with `std.debug.assert` plus saturating addition changes overflow handling from a returned `error.Overflow` to safety-checked illegal behavior (normally a panic) in ReleaseSafe. In ReleaseFast, violating the assertion reaches unchecked `unreachable`; saturation is not a guaranteed fallback while the assertion remains. Removing the assertion would give defined saturating addition, but still would not preserve the checked-add error contract. The ETH-supply argument alone is not a proof covering every state accepted by this API.

The branch is unchanged pending agreement on the intended overflow contract and current-compiler performance evidence.

---

## Motivation

`processRewardsAndPenalties` iterates over all validator balances (2.18M on mainnet). The inner loop used `try std.math.add(u64, ...)` which generates a branch per addition, preventing LLVM from auto-vectorizing.

## Change

Replace `try std.math.add` with saturating add (`+|`). Overflow is mathematically impossible:
- Max ETH supply: ~120M ETH ≈ 1.2 × 10¹⁷ Gwei
- u64 max: 1.8 × 10¹⁹
- Headroom: 154×

## Results

**Isolated microbenchmark** (500k validators, 100 iterations, ReleaseFast):
```
try std.math.add: 618.6 µs/iter
saturating +|:    179.6 µs/iter
speedup:          3.44×
```

**Full epoch benchmark** (`bench_process_epoch`, mainnet state at slot 13336576, 2.18M validators, 50 iterations, ReleaseFast):

| Metric | Checked (`try std.math.add`) | Saturating (`+\|`) | Δ |
|---|---|---|---|
| `rewards_and_penalties` (isolated) | 119.4ms ± 6.2ms | 117.5ms ± 5.7ms | −1.6% |
| Full epoch (segmented) | 794.1ms | 785.5ms | −1.1% |

The inner loop speedup (3.44×) is real but `processRewardsAndPenalties` is dominated by reward/penalty computation, so the end-to-end improvement is modest. That historical result does not establish the safety or current-build performance claims; see the current-status correction above.

### DIFF
--- a/src/state_transition/epoch/process_rewards_and_penalties.zig
+++ b/src/state_transition/epoch/process_rewards_and_penalties.zig
@@ -31,18 +31,19 @@ pub fn processRewardsAndPenalties(
     const balances = try state.balancesSlice(allocator);
     defer allocator.free(balances);
 
-    // Use saturating arithmetic instead of checked (`std.math.add`):
-    // - Overflow is impossible: max ETH supply ~120M ETH = ~1.2e17 Gwei,
-    //   u64 max = ~1.8e19, giving 154x headroom even with all rewards summed.
-    // - Saturating ops are branchless, enabling SIMD auto-vectorization of the
-    //   balance loop (~3.4x speedup on the inner loop for 500k+ validators).
+    // Use saturating arithmetic for performance. Saturating operations are branchless, which
+    // enables SIMD auto-vectorization of the balance update loop. This is safe because a
+    // balance overflow is mathematically impossible: the maximum total ETH supply (~1.2e17 Gwei)
+    // is more than 150x smaller than the max u64 value (~1.8e19 Gwei).
     if (slashing_penalties) |slashings| {
         for (rewards, penalties, balances, 0..) |reward, penalty, *balance, i| {
             const slashing: u64 = if (i < slashings.len) slashings[i] else 0;
+            std.debug.assert(balance.* <= std.math.maxInt(u64) - reward);
             balance.* = ((balance.* +| reward) -| penalty) -| slashing;
         }
     } else {
         for (rewards, penalties, balances) |reward, penalty, *balance| {
+            std.debug.assert(balance.* <= std.math.maxInt(u64) - reward);
             balance.* = (balance.* +| reward) -| penalty;
         }
     }
--- a/src/state_transition/epoch/process_rewards_and_penalties.zig
+++ b/src/state_transition/epoch/process_rewards_and_penalties.zig
@@ -31,14 +31,19 @@ pub fn processRewardsAndPenalties(
     const balances = try state.balancesSlice(allocator);
     defer allocator.free(balances);
 
+    // Use saturating arithmetic instead of checked (`std.math.add`):
+    // - Overflow is impossible: max ETH supply ~120M ETH = ~1.2e17 Gwei,
+    //   u64 max = ~1.8e19, giving 154x headroom even with all rewards summed.
+    // - Saturating ops are branchless, enabling SIMD auto-vectorization of the
+    //   balance loop (~3.4x speedup on the inner loop for 500k+ validators).
     if (slashing_penalties) |slashings| {
         for (rewards, penalties, balances, 0..) |reward, penalty, *balance, i| {
             const slashing: u64 = if (i < slashings.len) slashings[i] else 0;
-            balance.* = (try std.math.add(u64, balance.*, reward)) -| penalty -| slashing;
+            balance.* = ((balance.* +| reward) -| penalty) -| slashing;
         }
     } else {
         for (rewards, penalties, balances) |reward, penalty, *balance| {
-            balance.* = (try std.math.add(u64, balance.*, reward)) -| penalty;
+            balance.* = (balance.* +| reward) -| penalty;
         }
     }
 
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ node_modules/
 # Claude Code local config
 .claude/settings.local.json
 .claude/plans/
+zig-pkg/


src/state_transition/epoch/process_rewards_and_penalties.zig
```diff
@@ -31,14 +31,20 @@ pub fn processRewardsAndPenalties(
...
 
+    // Use saturating arithmetic for performance. Saturating operations are branchless, which
+    // enables SIMD auto-vectorization of the balance update loop. This is safe because a
+    // balance overflow is mathematically impossible: the maximum total ETH supply (~1.2e17 Gwei)
+    // is more than 150x smaller than the max u64 value (~1.8e19 Gwei).
     if (slashing_penalties) |slashings| {
         for (rewards, penalties, balances, 0..) |reward, penalty, *balance, i| {
             const slashing: u64 = if (i < slashings.len) slashings[i] else 0;
-            balance.* = (try std.math.add(u64, balance.*, reward)) -| penalty -| slashing;
+            std.debug.assert(balance.* <= std.math.maxInt(u64) - reward);
```
nazarhussain, line 41: Could not verify the 3.4x improvement but could measure 1.8x but in `ReleaseFast`. With `ReleaseSafe` there is no improvement recorded. But running under `ReleaseFast` caused many other boundary failures so would not recommend switching to it.

The culprit is the `debug.assert` which have a serious overhead in these particular case as it's under the iteration. 

If we want to go with mathematical  optimization then we should remove the assert here.

nazarhussain (changes requested):

src/state_transition/epoch/process_rewards_and_penalties.zig
lodekeeper-z: The assertion does retain an overflow check in `ReleaseSafe`; switching the whole client to `ReleaseFast` is not necessary or appropriate for this optimization.

I tried removing both assertions locally after merging current `main` (`58a9fcbc`) into this PR, using the current Zig 0.16.0 toolchain. There is a wrinkle: the extracted balance-update loop still compiled to scalar saturating arithmetic, not SIMD, even with non-aliasing arguments and AVX2/native targets. Removing the assertions alone therefore does not establish the PR's vectorization claim on the current compiler.

Matched `ReleaseSafe` microbenchmark, 500k balances, 200 calls per sample, nine samples with alternating baseline/candidate order:

| Path | With assertions | Without assertions |
|---|---:|---:|
| No slashings | 462.4 µs | 446.8 µs |
| Full-length slashings | 555.4 µs | 575.1 µs |

An earlier run showed the same direction (462.4 → 444.4 µs; 550.3 → 567.7 µs). These are isolated-loop measurements on this host, not a full-epoch benchmark or a reproduction of your machine's results. The small differences should not be generalized into an end-to-end speedup.

The candidate passed all 156 state-transition tests in `ReleaseSafe`. Both variants also matched a scalar reference across 24 length/slashing cases, including empty and short slashing slices.

I have left the remote head unchanged rather than push a performance refresh with mixed results. The old 3.44× figure is not evidence for current `ReleaseSafe`. My recommendation is to establish a reproducible current-compiler benchmark and inspect the actual production loop's codegen before landing this as a SIMD optimization; neither retaining the per-element assertion nor simply removing it has demonstrated that benefit here.

nazarhussain: Re-measured quiet, 4 stable runs at 2.18M balances, ReleaseSafe:

| variant | µs/iter |
|---|---:|
| `try std.math.add` (main) | ~1020 |
| assert + `+\|` (this PR) | ~1029 |
| `+\|`, no assert | ~937 |

So dropping the assert is worth ~8%, not the 14% I quoted.


**On the assertion retaining an overflow check** — I'd push back on that framing. It doesn't retain the check, it downgrades it:

- Before: `try std.math.add` → `error.Overflow`, recoverable, surfaces as a JS exception.
- After, ReleaseSafe: `panic: reached unreachable code` → process abort.
- After, ReleaseFast: an optimizer assumption; the balance silently saturates to `u64::MAX`.

lodekeeper-z: You are right: my wording conflated keeping a branch/check with preserving the error contract. The diff replaces a returned `error.Overflow` with an assertion, which is a materially different failure mode in ReleaseSafe.

One further distinction for ReleaseFast: `std.debug.assert` is `if (!ok) unreachable` (also in the current Zig 0.16 standard library). Violating it is unchecked illegal behavior, so the assertion-bearing version does **not** guarantee saturation on overflow. The no-assert `+|` variant has defined saturation, but that still differs from returning an error.

I corrected the PR body to withdraw the zero-behavior-change claim and separate the historical ReleaseFast measurements from the current ReleaseSafe evidence. Your new measurements apply to the no-assert variant, not the current head; I have not treated them as a reproduced benchmark or an end-to-end gain.

I am leaving the code unchanged rather than silently selecting a different overflow contract. My recommendation is to preserve checked-add error semantics unless we explicitly agree to narrow the accepted-state contract; the ETH-supply argument alone does not prove the invariant for every state accepted by the API.

PR #296, opened 2026-03-29
fix: use body hash in blockToHeader + resolve stale TODOs

lodekeeper-z:
## Summary

**Bug fix**: `blockToHeader` was computing `body_root` as `hash_tree_root(block)` instead of `hash_tree_root(block.body)`. This would produce incorrect `BeaconBlockHeader` values. The function is currently unused but will be needed for fork choice and beacon-node integration.

**TODO cleanup**: Resolves 3 stale TODOs with actionable notes:
- `attester_status.zig`: Missing functions (`parseAttesterFlags`, etc.) are for validator monitor, not STF — noted for beacon-node layer
- `block_root.zig`: `getTemporaryBlockHeader` already implemented as `blockToHeader` in `process_block_header.zig`; `signedBlockToSignedHeader` deferred to beacon-node layer

🤖 Generated with AI assistance

### DIFF
--- a/src/state_transition/utils/attester_status.zig
+++ b/src/state_transition/utils/attester_status.zig
@@ -29,4 +29,6 @@ pub fn hasMarkers(flags: u8, markers: u8) bool {
     return (flags & markers) == markers;
 }
 
-// TODOs: implement missing functions
+// Note: TS lodestar also has parseAttesterFlags, toAttesterFlags, parseParticipationFlags
+// These are used for validator monitoring and debug tooling, not STF correctness.
+// Add them when implementing the validator monitor (beacon-node layer).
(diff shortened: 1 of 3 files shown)


src/state_transition/utils/attester_status.zig
```diff
@@ -29,4 +29,6 @@ pub fn hasMarkers(flags: u8, markers: u8) bool {
     return (flags & markers) == markers;
 }
 
-// TODOs: implement missing functions
+// Note: TS lodestar also has parseAttesterFlags, toAttesterFlags, parseParticipationFlags
```
GrapeBaBa: don't need this change IMO

PR #312, opened 2026-04-12
[wip]feat(log): implement structured logging module

GrapeBaBa:
## Summary

Add a full-featured structured logging module for lodestar-z

### Key features

- **Comptime static dispatch pipeline**: Filter → Diagnostic → Append, all resolved at compile time with zero runtime overhead for disabled log levels
- **Three layout formats**: TextLayout (human-readable with ANSI color), JsonLayout, LogfmtLayout
- **Multiple appender types**: WriterAppend (sync), AsyncAppend (lock-free ring buffer + background drain), RollingFileWriter (size/date rotation), OpenTelemetryAppend (OTLP JSON)
- **Composable filters**: LevelFilter, ScopeFilter, EnvFilter (parses `RUST_LOG`-style directives like `warn,fork_choice=debug`)
- **Diagnostics**: StaticDiagnostic (pre-bound attrs), ThreadLocalDiagnostic (per-thread context)
- **std.log bridge**: Drop-in `std_options` so existing `std.log.scoped()` calls route through the pipeline
- **Global dispatcher builders**: `initConsoleDispatcher`, `initFileDispatcher`, `initCombinedDispatcher` for common setups
- **Microsecond timestamps**: ISO 8601 UTC format (`2024-08-11T22:44:57.172105Z`)
- **stderr by default**: Following Unix convention (stdout = data, stderr = diagnostics)
- **Extracted `ring_buffer.zig`**: Generic lock-free SPSC ring buffer, reusable outside logging

### Architecture

```
Logger(N) → AnyDispatcher → Dispatcher(Dispatches)
                               ├─ Dispatch(Filters, Diagnostics, Appends)
                               │    ├─ LevelFilter / ScopeFilter / EnvFilter
                               │    ├─ StaticDiagnostic / ThreadLocalDiagnostic
                               │    └─ WriterAppend(Layout) / AsyncAppend(Layout)
                               └─ Dispatch(...)
```

### Module structure

Consolidated from 20+ separate files into 6 focused modules:
- `record.zig` — Scope, Level, Attr, Record, comptime interface checks
- `filter.zig` — LevelFilter, ScopeFilter, EnvFilter
- `diagnostic.zig` — StaticDiagnostic, ThreadLocalDiagnostic
- `layout.zig` — TextLayout, JsonLayout, LogfmtLayout, ISO 8601 formatting
- `append.zig` — WriterAppend, AsyncAppend, RollingFileWriter, OpenTelemetryAppend
- `dispatch.zig` — Dispatch, Dispatcher, AnyDispatcher, Logger

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -149,6 +149,20 @@ pub fn build(b: *std.Build) void {
     module_bls.linkLibrary(dep_blst.artifact("blst"));
     b.modules.put(b.dupe("bls"), module_bls) catch @panic("OOM");
 
+    const module_ring_buffer = b.createModule(.{
+        .root_source_file = b.path("src/ring_buffer.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+    b.modules.put(b.dupe("ring_buffer"), module_ring_buffer) catch @panic("OOM");
+
+    const module_log = b.createModule(.{
+        .root_source_file = b.path("src/log/root.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+    b.modules.put(b.dupe("log"), module_log) catch @panic("OOM");
+
     const module_state_transition = b.createModule(.{
         .root_source_file = b.path("src/state_transition/root.zig"),
         .target = target,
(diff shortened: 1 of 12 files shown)


spiral-ladder:
First impression is 6k LOC is a lot

It looks like a big part of this pr implements a generalized logger supporting a few formats, which wouldn't be wrong if this were a fully featured logging library but as a logger contained within our lodestar-z suite I'm hesitant it needs all the features here

IMO I think we would need a very good reason to merge such a big PR all at one go. As a general rule I would prefer for us to prefer stdlib, and only implement abstractions on top if we find some parts lacking. It's easy to merge all these abstractions into the codebase but it's really hard to take out later if we find some parts are overengineered or redundant

With that said we can probably put this work in the backlog and revisit the ideas if we do need the features, and then only take and use what we need, working from small pieces to a big module rather than merging a big PR at once

GrapeBaBa:
> First impression is 6k LOC is a lot
> 
> It looks like a big part of this pr implements a generalized logger supporting a few formats, which wouldn't be wrong if this were a fully featured logging library but as a logger contained within our lodestar-z suite I'm hesitant it needs all the features here
> 
> IMO I think we would need a very good reason to merge such a big PR all at one go. As a general rule I would prefer for us to prefer stdlib, and only implement abstractions on top if we find some parts lacking. It's easy to merge all these abstractions into the codebase but it's really hard to take out later if we find some parts are overengineered or redundant
> 
> With that said we can probably put this work in the backlog and revisit the ideas if we do need the features, and then only take and use what we need, working from small pieces to a big module rather than merging a big PR at once

It is actually not ready to review. It is indeed a general logging library and should be a separate project, and I also found some parts of the design is not good to go further.

PR #316, opened 2026-04-15
ci: run mainnet spectests on 4core

spiral-ladder:


spiral-ladder:
seems like this doesn't work @_@

PR #329, opened 2026-04-28
fix(ssz): ignore BitVector padding bits

lodekeeper-z:
## Summary
- Mask padding bits only in the final byte of non-byte-aligned BitVectors in `getTrueBitIndexes`, `getSingleTrueBit`, and `intersectValues`.
- Prevent `getTrueBitIndexes` from writing beyond a correctly sized logical-length output buffer when an in-memory value contains dirty padding. For `BitVector[4]`, unmasked storage can emit eight indexes into a four-element buffer; checked builds panic, while unchecked builds risk an out-of-bounds write.
- Cover dirty padding with a four-bit regression and a logical-length output buffer.

## Scope and production context
This is defensive handling of dirty **in-memory** padding, not a demonstrated wire-triggered vulnerability. BitVector deserialization already rejects non-zero trailing bits. Mainnet-preset `committee_bits` is `BitVector[64]`, so it has no padding; the minimal preset uses four bits.

The earlier claim that this fixes the nogroup Electra/Fulu attestation crash was not established and is withdrawn. Deployment of this change does not establish the incident's root cause or resolution; that requires separate evidence.

This PR targets `feat/beacon-node`, where `BitList.getSingleTrueBit` already masks its tail and has a regression from f87edc14275fb747e5966ccbac6d4a35954c56bc. Current `main` still has the unmasked BitList implementation; that is a separate backport scope, not a missing change on this PR's base.

### DIFF
--- a/src/ssz/type/bit_vector.zig
+++ b/src/ssz/type/bit_vector.zig
@@ -40,6 +40,15 @@ pub fn BitVector(comptime _length: comptime_int) type {
             }
         }
 
+        fn maskedByte(self: *const @This(), i_byte: usize) u8 {
+            const remainder_bits = length % 8;
+            if (remainder_bits == 0 or i_byte + 1 < byte_len) {
+                return self.data[i_byte];
+            }
+            const tail_mask: u8 = (@as(u8, 1) << @intCast(remainder_bits)) - 1;
+            return self.data[i_byte] & tail_mask;
+        }
+
         pub fn getTrueBitIndexes(self: *const @This(), out: []usize) !usize {
             if (out.len < length) {
                 return error.InvalidSize;
@@ -47,7 +56,7 @@ pub fn BitVector(comptime _length: comptime_int) type {
             var true_bit_count: usize = 0;
 
             for (0..byte_len) |i_byte| {
-                var b = self.data[i_byte];
+                var b = self.maskedByte(i_byte);
 
                 while (b != 0) {
                     const lsb: usize = @as(u8, @ctz(b));
@@ -65,7 +74,7 @@ pub fn BitVector(comptime _length: comptime_int) type {
             var found_index: ?usize = null;
 
             for (0..byte_len) |i_byte| {
-                var b = self.data[i_byte];
+                var b = self.maskedByte(i_byte);
 
                 while (b != 0) {
                     if (found_index != null) {
@@ -132,10 +141,10 @@ pub fn BitVector(comptime _length: comptime_int) type {
             allocator: std.mem.Allocator,
             values: *const [length]T,
         ) !std.array_list.AlignedManaged(T, null) {
-            var indices = try std.array_list.AlignedManaged(T, null).initCapacity(allocator, byte_len * 8);
+            var indices = try std.array_list.AlignedManaged(T, null).initCapacity(allocator, length);
 
             for (0..byte_len) |i_byte| {
-                var b = self.data[i_byte];
+                var b = self.maskedByte(i_byte);
                 // Kernighan's algorithm to count the set bits instead of going through 0..8 for every byte
                 while (b != 0) {
                     const lsb: usize = @as(u8, @ctz(b)); // Get the index of least significant bit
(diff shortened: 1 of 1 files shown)


wemeetagain:
I think maskedByte should only be needed for the final byte

lodekeeper-z:
Updated in fe4768a3: `maskedFinalByte()` is now only used for the final byte when the BitVector length is not byte-aligned; full bytes read `self.data[i_byte]` directly.

Verification:
- `zig build test:ssz -Dssz.filters='BitVectorType helpers ignore padding bits'`
- `zig build test:ssz`

nazarhussain (changes requested):
Three things I'd like to resolve before/around merging.

**1. The severity is higher than the description suggests**

It's an out-of-bounds write. Reproduced on `main` with a `Bitvector[4]` whose padding bits are set, using an `out` buffer sized exactly `length`:

```
panic: index out of bounds: index 4, len 4
```

`getTrueBitIndexes` only guards `if (out.len < length)`, but it can emit up to `byte_len * 8` indexes (8 for a 4-bit vector). Both call sites pass a buffer sized exactly `[preset.MAX_COMMITTEES_PER_SLOT]usize`:

- `src/state_transition/block/process_attestation_phase0.zig:101`
- `src/state_transition/cache/epoch_cache.zig:792`

So a *correctly sized* caller overflows. ReleaseSafe (what we ship) turns that into the panic above. 

Worth restating in the PR body, since it changes how urgently this wants backporting.

**2. The stated root cause doesn't hold — the real trigger may still be live**

Two independent reasons the described path can't produce this crash:

- **On mainnet, `MAX_COMMITTEES_PER_SLOT = 64`**, so `committee_bits` is `Bitvector[64]`,  byte-aligned, no padding bits exist, and this PR's masking is compiled out entirely (`length % 8 != 0` is comptime-false). Only the *minimal* preset makes it `Bitvector[4]`.
- **Dirty padding can't survive deserialization.** `BitVector.serialized.validate` has rejected non-zero trailing bits since cdd7a832 (2025-06-10), ~11 months before this PR, and `ContainerType.deserializeFromBytes` calls each field's validating deserializer. So a gossip/wire attestation cannot carry dirty `committee_bits`.


**3. `BitList.getSingleTrueBit` has the same bug**

`bit_list.zig` already masks its tail in `getTrueBitIndexes` and `intersectValues` — so this PR is bringing `BitVector` to parity, which is the right call. But `BitList.getSingleTrueBit` (~line 95) iterates all of `self.data.items` with no tail mask.

Only test callers today, so it's latent, but `BitList` is the type whose `bit_len` is runtime-variable and almost never byte-aligned (`aggregation_bits`), so if we're fixing this class of bug it'd be good to close it there too.

lodekeeper-z:
@nazarhussain Correct on the buffer bound and on the unsupported production attribution. I updated the body to explicitly describe the out-of-bounds write for dirty in-memory padding, distinguish checked-build panics from unchecked writes, and withdraw the claim that this establishes the nogroup crash's cause or resolution. The existing deserializer rejects trailing bits, and the mainnet preset uses a byte-aligned 64-bit committee vector.

There is an important base-branch distinction for point 3: this PR targets `feat/beacon-node`, not `main`. Its head already contains the BitList tail mask **and** `BitListType - getSingleTrueBit ignores tail bits beyond bit_len`, introduced by [f87edc14](https://github.com/ChainSafe/lodestar-z/commit/f87edc14275fb747e5966ccbac6d4a35954c56bc). I verified that commit is also an ancestor of the current target branch. Your observation is correct for current `main` (d4cc6b9e), which still has the unmasked implementation; a main-targeted backport would need that fix too. No duplicate BitList implementation change is needed on this PR.

Verified at the unchanged PR head fe4768a3 with Zig 0.16.0:
- `zig build test:ssz -Dssz.filters='BitVectorType helpers ignore padding bits'` — passed; this regression already uses an output buffer sized exactly to the four logical bits.
- `zig build test:ssz -Dssz.filters='getSingleTrueBit ignores tail bits beyond bit_len'` — passed.
- `zig build test:ssz` — passed.

The historical production incident remains unproven by this patch; I have not treated it as resolved or changed the deployment.

nazarhussain (approved):

wemeetagain:
I'm just going to close this and we can reconsider this later. Imo likely not necessary.

PR #330, opened 2026-04-28
feat: migrate BeaconStateView to use zapi js dsl

nazarhussain:
Use new Zapi DSL for `BeaconStateView` bindings.

### DIFF
--- a/bindings/napi/BeaconStateView.zig
+++ b/bindings/napi/BeaconStateView.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const napi = @import("zapi:zapi").napi;
+const js = @import("zapi:zapi").js;
 const c = @import("config");
 const fork_types = @import("fork_types");
 const st = @import("state_transition");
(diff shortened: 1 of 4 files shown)


bindings/napi/BeaconStateView.zig
```diff
@@ -10,62 +11,90 @@ const AnySignedBeaconBlock = fork_types.AnySignedBeaconBlock;
...
-        cb.this(),
-        CachedBeaconState,
-        cached_state,
-        BeaconStateView_finalize,
-        null,
-        null,
-    );
-
-    return cb.this();
+pub const js_meta = js.class(.{ .properties = .{
```
spiral-ladder, line 25: is there some way to comptime assert that this has been declared for classes?

also, would be nice if we could have a default `spec` declared and reused for the properties rather than have to read all of the same in the below lines (probably easier for debugging things at a glance too)

An adjacent question: why is `spec` `anytype`?

bindings/napi/BeaconStateView.zig
```diff
@@ -10,62 +11,90 @@ const AnySignedBeaconBlock = fork_types.AnySignedBeaconBlock;
...
+    .activeValidatorCount = js.prop(.{ .get = true, .set = false }),
+    .isMergeTransitionComplete = js.prop(.{ .get = true, .set = false }),
+    .isExecutionStateType = js.prop(.{ .get = true, .set = false }),
+    .proposerRewards = js.prop(.{ .get = true, .set = false }),
+    .clonedCount = js.prop(.{ .get = true, .set = false }),
+    .clonedCountWithTransferCache = js.prop(.{ .get = true, .set = false }),
+    .createdWithTransferCache = js.prop(.{ .get = true, .set = false }),
+} });
+
+cached_state: ?*CachedBeaconState = null,
```
spiral-ladder, line 68: feel like this should belong with `gpa` rather than below `js_meta`

bindings/napi/BeaconStateView.zig
```diff
@@ -10,62 +11,90 @@ const AnySignedBeaconBlock = fork_types.AnySignedBeaconBlock;
...
+const Self = @This();
```
spiral-ladder:
```suggestion
const BeaconStateView = @This();
```

think this should be `BeaconStateView` and at the top of the file

bindings/napi/BeaconStateView.zig
```diff
@@ -969,137 +921,56 @@ pub fn BeaconStateView_hashTreeRoot(env: napi.Env, cb: napi.CallbackInfo(0)) !na
...
+}
 
-    const ctor = try cb.this().getNamedProperty("constructor");
-    const new_state_value = try env.newInstance(ctor, .{});
-    const dummy_state = try env.unwrap(CachedBeaconState, new_state_value);
-    dummy_state.* = post_state.*;
-    allocator.destroy(post_state);
+fn requireState(self: *const Self) !*CachedBeaconState {
+    return self.cached_state orelse error.InvalidState;
+}
```
spiral-ladder: i'm not sure of the value of hiding this one-liner behind a `requireState` function rather than just doing it inline, seems unnecessary imo. good thing about doing it inline is we can see (without jumping to another fn) that it might return `error.InvalidState`

bindings/napi/BeaconStateView.zig
```diff
@@ -78,390 +107,371 @@ pub fn BeaconStateView_createFromBytes(env: napi.Env, cb: napi.CallbackInfo(1))
...
     };
 
-    var bytes: [*]u8 = undefined;
-    const buf = try env.createArrayBuffer(size, &bytes);
-    _ = pending_consolidations.serializeIntoBytes(bytes[0..size]) catch {
-        try env.throwError("STATE_ERROR", "Failed to serialize pendingConsolidations");
-        return env.getNull();
+    const result = try js.Uint8Array.alloc(size);
+    _ = pending_consolidations.serializeIntoBytes(try result.toSlice()) catch {
+        return throwNullAs(js.Uint8Array, "STATE_ERROR", "Failed to serialize pendingConsolidations");
```
spiral-ladder, line 341: should we define consts for `code`s like `STATE_ERROR` which are reused? Also, what is the real diff between `code` and `msg`? Can we make `code` more useful on the native layer by using error sets and then mapping it to `[:0] const u8` on the boundary?

bindings/napi/BeaconStateView.zig
```diff
@@ -969,137 +921,56 @@ pub fn BeaconStateView_hashTreeRoot(env: napi.Env, cb: napi.CallbackInfo(0)) !na
...
-    dummy_state.* = post_state.*;
-    allocator.destroy(post_state);
+fn requireState(self: *const Self) !*CachedBeaconState {
+    return self.cached_state orelse error.InvalidState;
+}
 
-    return new_state_value;
+fn asJsValue(value: napi.Value) js.Value {
+    return .{ .val = value };
 }
```
spiral-ladder: should we just use `wrap(js.Value, value)` over this? Seems like it's the same just without `T`

bindings/napi/BeaconStateView.zig
```diff
@@ -10,62 +11,90 @@ const AnySignedBeaconBlock = fork_types.AnySignedBeaconBlock;
...
+const js_types = @import("./js_types.zig");
 const sszValueToNapiValue = @import("./to_napi_value.zig").sszValueToNapiValue;
 const numberSliceToNapiValue = @import("./to_napi_value.zig").numberSliceToNapiValue;
-
-const getter = @import("napi_property_descriptor.zig").getter;
-const method = @import("napi_property_descriptor.zig").method;
+const napi_io = @import("./io.zig");
 
 /// Allocator used for all BeaconStateView instances.
 var gpa: std.heap.DebugAllocator(.{}) = .init;
```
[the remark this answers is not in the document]
nazarhussain: it was already in used, would like to change it in separate PR if needed.

bindings/napi/BeaconStateView.zig
nazarhussain: All `js.*` namespaces are `comptime` layer. We assert all the properties declared here for the struct it was declared for. Declaring a `pub` interface for a file in zig make it a by default a struct. So here the decleration of `js.class` will be asserted against `BeaconStateView.zig` file. 

I didn't understood other question related to `spec`, you mean the struct given for `.properties`, if that's the question then we don't know what properties will be declared for any class so it's type is `anytype`.

bindings/napi/BeaconStateView.zig
nazarhussain: The identifier name does not matter, using `Self` is a common convention referring to `@This()`

bindings/napi/BeaconStateView.zig
nazarhussain: It is module level property, not the `js-class` decleration.

bindings/napi/BeaconStateView.zig
nazarhussain: Having `code` and `msg` for error objects is convention very common in JS, though an `Error` can have any properties. 

Mostly consumers do logical conditioning based on `code` and use `msg` for logging purpose. 

Yes we can define `consts` for the `error codes` or we can use the error unions as well. But would like to do it in separate PR.

bindings/napi/BeaconStateView.zig
nazarhussain: Well while reading code you read the `requireState` once, and then on all occurrences you know what `requireState` means. Benefit of having safe guard function is for debugging/logging purpose. As we are not doing any logging here anyway so we can remove that as well. No hard opinion on it, if you feel strong about it I can change that.

bindings/napi/BeaconStateView.zig
spiral-ladder: As of now if you remove the `js_meta` var all you get is an unobvious error that doesn't clearly say that `js_meta` has to declared:

```
zig-pkg/zapi-2.0.0-rIqzUcc3BABKRuJlzpYNtiMVCOFybuVSGOLmk0KOCeel/src/js/wrap_function.zig:216:5: error: convertReturn: unsupported return type BeaconStateView
    @compileError("convertReturn: unsupported return type " ++ @typeName(T));
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

So imo it'd be nicer UX if there's some way to assert the existence of this var on that layer (not sure if it's possible?)

As for `spec` question, I was asking about the `prop` fn on the zapi layer but your reply answered it :)

bindings/napi/BeaconStateView.zig
spiral-ladder: it matters in the sense of consistency across codebase, our CamelCase structs (eg. `ChainConfig.zig`, `BeaconConfig.zig`) uses their name for `@This()`. I guess it doesn't matter for the scope of this PR but eventually we might want to use one variant consistently throughout

bindings/napi/BeaconStateView.zig
spiral-ladder: @nazarhussain let's keep it, no strong opinion too, perhaps a bit too pedantic of a comment

bindings/napi/BeaconStateView.zig
spiral-ladder: sounds good, let's do it in future

spiral-ladder (dismissed):
thanks for answering my queries! Think this LGTM

bindings/napi/BeaconStateView.zig
nazarhussain: Well the `js_meta` is not required if you just export pure functions from module. But if your functions are using or returning class objects then it expect to `js_meta`. 

I created an issue to improve that error message https://github.com/ChainSafe/zapi/issues/24

spiral-ladder (approved):

PR #333, opened 2026-04-29
refactor: use dontTransferCache on the bindings layer

spiral-ladder:
despite the double negation being confusing we're still on `dontTransferCache` on lodestar, so this change is just to reduce the diff surface when integrating bindings there.

We should switch back to `transferCache` for clearer naming once  native state transition is integrated.


spiral-ladder:
not sure why `getEffectiveBalanceIncrementsZeroInactive` is failing on an unrelated change...

nazarhussain (approved):

PR #338, opened 2026-05-02
fix(stf): post state not being committed when verify state root is true

spiral-ladder:
We need to commit the post state regardless of whether we verify state root or not

### DIFF
--- a/src/state_transition/state_transition.zig
+++ b/src/state_transition/state_transition.zig
@@ -233,12 +233,7 @@ pub fn stateTransition(
     }
     metrics.state_transition.process_block.observe(time.durationSeconds(time.since(io, timer)));
 
-    //
-    // TODO(bing): commit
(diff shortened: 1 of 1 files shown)


src/state_transition/state_transition.zig
```diff
@@ -233,12 +233,7 @@ pub fn stateTransition(
     }
     metrics.state_transition.process_block.observe(time.durationSeconds(time.since(io, timer)));
 
-    //
-    // TODO(bing): commit
-    //  const processBlockCommitTimer = metrics?.processBlockCommitTime.startTimer();
-    //  postState.commit();
-    //  processBlockCommitTimer?.();
```
spiral-ladder, line 240: also remove these outdated stuff that we're already doing

spiral-ladder:
`hashTreeRoot` commits so this is wrong

PR #341, opened 2026-05-03
test(spec): walk both presets so mainnet-only cases are emitted

GrapeBaBa:
## Summary

The EF spec test pack ships **asymmetric vector packs**: ~97 case dirs only exist under `mainnet/tests/mainnet/.../operations/...` (e.g. `random_*_with_duplicates` for `sync_aggregate`), while hundreds only exist under `minimal/tests/minimal/...`. The current generator walks just one preset (`minimal`), so any mainnet-only case is silently dropped at codegen time and never reaches the spec runner.

This PR makes `write_spec_tests.zig` walk **both** presets and dedup the union by case key, so every case the EF pack ships gets a generated test entry.

### Concrete impact (operations shard, mainnet)

| Before | After |
|--------|-------|
| 1567 / 1734 (167 skip) | 1664 / 1831 (167 skip) |

`+97` newly emitted mainnet-only cases. Skip count is unchanged because those 167 are runner-handled skips, not codegen omissions.

### DIFF
--- a/test/spec/write_spec_tests.zig
+++ b/test/spec/write_spec_tests.zig
@@ -25,6 +25,15 @@ const supported_test_runners = [_]RunnerKind{
     .finality,
 };
 
+// EF ships asymmetric vector packs — for example ~97 case dirs only exist
+// under mainnet (e.g. `*/operations/sync_aggregate/random_*_with_duplicates`)
+// while hundreds only exist under minimal. Walking just one preset silently
+// drops the others; the union is dedup'd by the case key the writer expects.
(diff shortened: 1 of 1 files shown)


GrapeBaBa:
@codex review

GrapeBaBa:
@claude review

GrapeBaBa:
@codex review

GrapeBaBa:
`gemini review`

wemeetagain (approved):
Good catch!
TIL that there are mainnet-only cases

PR #349, opened 2026-05-05
refactor(bindings): use `toU32()` where it makes sense instead of `toI64()`

spiral-ladder:
Prefer unsigned 32-bit integers for 2 reasons:

1) Realistically some of these values will not be bigger than a u32
2) u32 autowidens to u64 (which some params are the type of)


spiral-ladder:
on second thought this doesn't really make sense since it compromises correctness for some relatively unimportant ergonomics, closing

PR #353, opened 2026-05-07
feat: asyncAggregateWithRandomness

spiral-ladder:
This PR reintroduces pippenger based `asyncAggregateWithRandomness`, heavily modelled after the [rust bindings](https://github.com/supranational/blst/blob/dece82ea537b422890888bacde4034ca5b5a44d8/bindings/rust/src/pippenger.rs ).

We use a comptime `parallelMSM` fn which accepts a `Curve`. The underlying algorithm is the same, we just need to choose the usage (at comptime) between `G1`, which is the pubkey, or `G2`, the signature.

We don't necessarily have to do this but I thought it looks cleaner rather than repeating the algo twice. Of course this means we have to do some ugly things like defining a `CurveDescriptor` struct and use a bunch of pointer casts to anyopaque functions.

## Implementation details

Note that there are a few key differences:

- instead of using `blst`'s internal threadpool, we use our own `ThreadPool`
- Rust version uses pipelining to accumulate work. Key difference here is that the rows can come out of order because the top rows can start being summed while bottom rows are still computing; we use `submitAndWait` which assembles work sequentially from top to bottom. This has the benefit of allowing us to drop one of the outer loops that Rust does to handle out-of-order rows from workers. We have the benefit of simpler code here.

## Benchmarks

Looks pretty comparable on M2 Mac:

```
bindings/perf/blst.test.ts
  asyncAggregateWithRandomness
    ✔ asyncAggregateWithRandomness lodestar-z  1 sets                     5456.162 ops/s    183.2790 us/op        -        713 runs  0.996 s
    ✔ asyncAggregateWithRandomness @chainsafe/blst  1 sets                4940.736 ops/s    202.3990 us/op        -        524 runs  0.891 s
    ✔ asyncAggregateWithRandomness lodestar-z  8 sets                     1274.951 ops/s    784.3440 us/op        -         82 runs   1.82 s
    ✔ asyncAggregateWithRandomness @chainsafe/blst  8 sets                1295.293 ops/s    772.0260 us/op        -        138 runs   2.02 s
    ✔ asyncAggregateWithRandomness lodestar-z  32 sets                    408.2397 ops/s    2.449541 ms/op        -         44 runs   2.16 s
    ✔ asyncAggregateWithRandomness @chainsafe/blst  32 sets               413.3244 ops/s    2.419407 ms/op        -         31 runs   1.96 s
    ✔ asyncAggregateWithRandomness lodestar-z  64 sets                    221.6369 ops/s    4.511884 ms/op        -          9 runs   1.80 s
    ✔ asyncAggregateWithRandomness @chainsafe/blst  64 sets               218.6257 ops/s    4.574028 ms/op        -         13 runs   1.90 s
    ✔ asyncAggregateWithRandomness lodestar-z  128 sets                   111.7445 ops/s    8.948983 ms/op        -          5 runs   1.88 s
    ✔ asyncAggregateWithRandomness @chainsafe/blst  128 sets              113.1346 ops/s    8.839029 ms/op        -          7 runs   1.99 s

```

EDIT: Something to note is that this was mostly claude; but i did audit the code against the existing rust bindings and it's more or less doing the same stuff, would appreciate a separate pair of eyes to see if i missed something though.

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -798,11 +798,7 @@ pub fn asyncAggregateWithRandomness(env: napi.Env, cb: napi.CallbackInfo(1)) !na
     data.err = null;
     data.deferred = undefined;
     data.work = undefined;
-
-    var seed_bytes: [8]u8 = undefined;
-    napi_io.get().random(&seed_bytes);
-    var prng = std.Random.DefaultPrng.init(std.mem.readInt(u64, &seed_bytes, .little));
-    prng.random().bytes(data.randomness);
+    napi_io.get().random(data.randomness);
 
     for (0..n) |i| {
         const set_value = try sets.getElement(@intCast(i));
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -305,7 +305,6 @@ pub const SecretKey = struct {
         return formatHex(bytes[0..]);
     }
 };
-
 /// Verifies a given `msg` against a `Signature` and a `PublicKey`.
 ///
 /// Returns `true` if signature is valid, `false` otherwise.
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -647,3 +647,192 @@ pub fn aggregateWithRandomness(sets: js.Array) !js.Value {
     try result.setNamedProperty("sig", sig_value);
     return .{ .val = result };
 }
+
+/// Heap-allocated context shared between the JS thread (which kicks off the work),
+/// the libuv worker thread (which calls `ThreadPool.aggregateWithRandomness`), and
+/// the JS thread again (which resolves/rejects the Promise).
+///
+/// All input data should be copied into this struct so the worker thread doesn't depend on
+/// any JS-managed memory staying alive.
+const AsyncAggRandData = struct {
+    pks: []PublicKey,
+    sigs: []Signature,
+    pk_ptrs: []*const PublicKey,
+    sig_ptrs: []*const Signature,
+    randomness: []u8,
+    pk_out: PublicKey,
+    sig_out: Signature,
+    err: ?anyerror,
+    deferred: napi.Deferred,
+    work: napi.c.napi_async_work,
+
+    fn destroy(self: *AsyncAggRandData) void {
+        allocator.free(self.pks);
+        allocator.free(self.sigs);
+        allocator.free(self.pk_ptrs);
+        allocator.free(self.sig_ptrs);
+        allocator.free(self.randomness);
+        allocator.destroy(self);
+    }
+};
+
+/// Execute `aggregateWithRandomness` on a libuv worker thread.
+///
+/// Assumes that:
+/// 1) pubkeys are already validated,
+/// 2) signatures are not group-checked on JS thread
+///
+/// Note: MUST NOT call any napi APIs.
+fn asyncAggRand_execute(_: napi.Env, data: *AsyncAggRandData) void {
+    const pool = thread_pool orelse {
+        data.err = error.PoolNotInitialized;
+        return;
+    };
+    pool.aggregateWithRandomness(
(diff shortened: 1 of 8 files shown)


bindings/napi/blst.zig
```diff
@@ -954,6 +969,195 @@ pub fn blst_aggregateWithRandomness(env: napi.Env, cb: napi.CallbackInfo(1)) !na
...
+    const n = try sets.getArrayLength();
+
+    if (n == 0) return error.EmptyArray;
+    if (n > MAX_AGGREGATE_PER_JOB) return error.TooManySets;
+    if (thread_pool == null) return error.PoolNotInitialized;
+
+    const data = try allocator.create(AsyncAggRandData);
+    errdefer allocator.destroy(data);
+
+    data.pks = try allocator.alloc(PublicKey, n);
```
twoeths, line 785: since we cap MAX_AGGREGATE_PER_JOB
we can use stack allocation instead? same to below

src/bls/ThreadPool.zig
```diff
@@ -459,6 +462,47 @@ fn mergeAndVerify(
...
+    sigs_groupcheck: bool,
+    pk_out: *PublicKey,
+    sig_out: *Signature,
+) (BlstError || PoolError || std.Io.Cancelable || std.mem.Allocator.Error)!void {
+    if (pks.len == 0 or pks.len != sigs.len) return BlstError.AggrTypeMismatch;
+    if (pks.len > blst.MAX_AGGREGATE_PER_JOB) return BlstError.AggrTypeMismatch;
+    if (randomness.len < pks.len * 32) return BlstError.AggrTypeMismatch;
+
+    if (pks_validate) for (pks) |pk| try pk.validate();
+    if (sigs_groupcheck) for (sigs) |sig| try sig.validate(false);
```
twoeths:
```suggestion
    if (sigs_groupcheck) for (sigs) |sig| try sig.validate(true);
```
see https://github.com/ChainSafe/blst-ts/blob/86c49590d37d4e1dd44b3b5ba604132f3b51d99d/src/lib.rs#L427

need to double check with `fn aggregateVerify()` too

spiral-ladder: Good catch, thanks, fixed in [4fdf956](https://github.com/ChainSafe/lodestar-z/pull/353/commits/4fdf956b3a6aae340c2d1c5736b3bbf8155bf0bd)

bindings/napi/blst.zig
spiral-ladder: i think this wouldn't work because of the lifetime of the workload since its dispatching to async work

bindings/napi/blst.zig
```diff
@@ -954,6 +954,195 @@ pub fn blst_aggregateWithRandomness(env: napi.Env, cb: napi.CallbackInfo(1)) !na
...
+    data.pk_out = .{};
+    data.sig_out = .{};
+    data.err = null;
+    data.deferred = undefined;
+    data.work = undefined;
+
+    var seed_bytes: [8]u8 = undefined;
+    napi_io.get().random(&seed_bytes);
+    var prng = std.Random.DefaultPrng.init(std.mem.readInt(u64, &seed_bytes, .little));
+    prng.random().bytes(data.randomness);
```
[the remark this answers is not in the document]
spiral-ladder: fixed in [230f6cc](https://github.com/ChainSafe/lodestar-z/pull/353/commits/230f6ccc7e208248fd8a111f0b866f90972690d4), more sounder suggestion overall

twoeths (approved):
lgtm

bindings/napi/blst.zig
twoeths: I see, for async api we have to use heap allocation because the work is scheduled via libuv callback

wemeetagain (approved):

PR #354, opened 2026-05-08
feat(clock): add clock module for slot/epoch timing

GrapeBaBa:
## Motivation

Lodestar (TS) needs a beacon clock that emits slot/epoch events and lets callers wait on specific slots. This adds the equivalent in Zig as a self-contained module under `src/clock/`, ported from [ChainSafe/clock-zig](https://github.com/ChainSafe/clock-zig) and adapted to Zig 0.16's `std.Io` async model.

## Description

Three-layer architecture:

- **Layer 0** (`slot_math`) — pure arithmetic, comptime-compatible.
- **Layer 1** (`SlotClock`) — stateful clock with pluggable `TimeSource`.
- **Layer 2** (`EventClock`) — async event loop with listeners and waiters; built on `std.Io`.


GrapeBaBa:
@codex review

GrapeBaBa:
@codex review

GrapeBaBa:
@codex review again

GrapeBaBa:
@codex review

src/clock/EventClock.zig
```diff
@@ -0,0 +1,830 @@
...
+        .waiters = WaiterQueue.initContext({}),
+    };
+    self.clock = SlotClock.init(config, .{ .real = .{ .io = io_handle } }) catch return error.InvalidConfig;
+}
+
+/// Start the auto-advance loop.  Idempotent; second call is a no-op.
+pub fn start(self: *EventClock) Error!void {
+    if (self.loop_future != null) return;
+    self.loop_future = std.Io.concurrent(self.io, EventClock.runAutoLoop, .{self}) catch
+        return error.ConcurrencyUnavailable;
```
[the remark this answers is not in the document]
GrapeBaBa: this logic is same with lodestar ts

GrapeBaBa:
@codex review

src/clock/slot_math.zig
```diff
@@ -0,0 +1,386 @@
...
+        @compileError("too many slot duration transitions");
+    }
+    var arr: DurationTransitions = .{};
+    inline for (list) |t| arr.push(t);
+    return arr;
+}
+
+/// `duration_transitions` entries must be sorted strictly ascending by
+/// `from_slot`, with non-zero `new_duration_ms` and `from_slot != 0` (validated).
+pub const Config = struct {
```
nazarhussain: Suggest to rename it to `ClockConfig` or `ClockParams` and extract it out from this file as it's not related to slot math.

src/clock/time_source.zig
```diff
@@ -0,0 +1,76 @@
...
+//!
+//! Tagged union with two variants:
+//!   `.io`   – production: reads wall-clock time from `std.Io`
+//!   `.fake` – testing: reads from a mutable `FakeTime` struct
+
+const std = @import("std");
+const slot_math = @import("slot_math.zig");
+
+/// Production clock backed by std.Io wall-clock time.
+pub const RealClock = struct {
```
nazarhussain: Pick one noun `RealClock` <> `FakeClock` or `RealTime` <> `FakeTime`

src/clock/time_source.zig
```diff
@@ -0,0 +1,76 @@
...
+
+    pub fn nowMs(self: RealClock) u64 {
+        const ms = std.Io.Clock.real.now(self.io).toMilliseconds();
+        std.debug.assert(ms >= 0);
+        return @intCast(ms);
+    }
+};
+
+/// Controllable time source for deterministic testing.
+pub const FakeTime = struct {
```
nazarhussain: As above, pick one one noun for both structs.

src/clock/time_source.zig
```diff
@@ -0,0 +1,76 @@
...
+    real: RealClock,
+    fake: *FakeTime,
+
+    pub fn nowMs(self: TimeSource) u64 {
+        return switch (self) {
+            .real => |c| c.nowMs(),
+            .fake => |f| f.ms,
+        };
+    }
+};
```
nazarhussain: Not to change request, just asking for a style opinion. Why not a have one open interface where implementation for `RealClock` just throws error that you can not advance the clock. 

```zig
pub const TimeSource = struct {                                                                                                                              
      context: *anyopaque,                                                                                                                                   
      nowMsFn: *const fn (*anyopaque) u64,
      advanceMsFn: *const fn (*anyopaque) void,
                                                                                                                                                               
      pub fn nowMs(self: TimeSource) u64 {
          return self.nowMsFn(self.context);                                                                                                                   
      }                                                                                                                                                      
      
      pub fn advanceMs(self: TimeSource) void {
           return self.advanceMsFn(self.context);
      }
  };     
````


Now we can have two `pub fn realTimeSource(io) TimeSource` and `pub fn fakeTimeSource(io) TimeSource` with their own implementation. 

Why would we prefer the tagged union over this open set style.

File: src/clock/SlotClock.zig [hunk 6]
```diff
@@ -0,0 +1,327 @@
...
+}
+
+pub fn msFromSlot(self: *const SlotClock, slot: Slot, to_ms: ?u64) ?i64 {
+    const from_ms = slot_math.slotStartMs(self.config, slot) orelse return null;
+    const end_ms = to_ms orelse self.time.nowMs();
+    const diff = @as(i128, @intCast(end_ms)) - @as(i128, @intCast(from_ms));
+    return std.math.cast(i64, diff);
+}
+
+pub fn advanceTo(self: *SlotClock, target: Slot) AdvanceIterator {
```
nazarhussain, line 180:
```suggestion
/// Advances the clock toward `target` one event at a time.  The caller may
/// drop the iterator mid-walk; the clock is then left at the last slot the                                                                                  
/// iterator returned (i.e. partial advancement is observable).
pub fn advanceTo(self: *SlotClock, target: Slot) AdvanceIterator {
```

src/clock/EventClock.zig
```diff
@@ -0,0 +1,1285 @@
...
+    self.catchUp();
+    return self.clock.currentSlotWithGossipDisparity();
+}
+
+pub fn isCurrentSlotGivenGossipDisparity(self: *EventClock, slot: Slot) bool {
+    self.catchUp();
+    return self.clock.isCurrentSlotGivenGossipDisparity(slot);
+}
+
+pub fn slotWithFutureTolerance(self: *EventClock, tolerance_ms: u64) ?Slot {
```
nazarhussain:
```suggestion
pub fn slotWithFutureToleranceMs(self: *EventClock, tolerance_ms: u64) ?Slot {
```

src/clock/EventClock.zig
```diff
@@ -0,0 +1,1285 @@
...
+    return self.clock.slotWithFutureTolerance(tolerance_ms);
+}
+
+pub fn slotWithPastTolerance(self: *EventClock, tolerance_ms: u64) ?Slot {
```
nazarhussain:
```suggestion
pub fn slotWithPastToleranceMs(self: *EventClock, tolerance_ms: u64) ?Slot {
```

nazarhussain (dismissed):
Overall it is good. just some design observations. 


1. The separation between `SlotClock` and `EventClock` is not  easy to reason about and create more confusion for understanding the code. 
2. I recommend to have one public `Clock.zig` which mostly be the `EventClock`. 
3. Internalize the `SlotClock` as it's not consumed directly. 

Image

src/clock/time_source.zig
GrapeBaBa: Good question, I used vtable implementation in the previous commit, while I reviewed it I thought `TimeSource` not need open to extend which should only two implementation, I switched to tagged union approach.

GrapeBaBa:
@codex review

GrapeBaBa:
@codex review

GrapeBaBa:
@codex review

GrapeBaBa:
@codex review

twoeths (commented):
seems too much to have a separate `src/clock` module, maybe have it under `src/utils` instead?
some upcoming modules to be added are quite big, like `src/beacon_node`, `src/validator`

wemeetagain:
Imo its fine to have small modules if / when it makes sense. (For example we have a hex module that is just a single file!) I also don't really like "utils" as a top-level concept, I think that kind of misc grab-bag is lazy organization and should be avoided if possible.

GrapeBaBa:
@nazarhussain could you review again, addressed your comments and make time_source generic

src/clock/time_source.zig
```diff
@@ -0,0 +1,64 @@
+//! Pluggable time source so tests can inject deterministic time.
```
wemeetagain: This seems a bit strange to me because std.Io is already an interface which might have a mocked time implementation.

spiral-ladder:
> seems too much to have a separate `src/clock` module, maybe have it under `src/utils` instead? some upcoming modules to be added are quite big, like `src/beacon_node`, `src/validator`

i'm also against dumping more things into `/utils` 😬 every fn, every module is technically a utility if we think about things from a util world. Hence https://github.com/ChainSafe/lodestar-z/issues/169

there were some efforts to move things out (eg. https://github.com/ChainSafe/lodestar-z/pull/322) but dont think we have reached some conclusion yet as a team, but refactoring is not too urgent, we just want to be careful to not add more things into the utils black hole

src/clock/time_source.zig
GrapeBaBa: Change to use a fake clock io for testing, it looks like more cleaner

src/clock/slot_math.zig
```diff
@@ -0,0 +1,349 @@
+//! Layer 0 – Pure slot/epoch arithmetic.
+//!
+//! No state, no allocation, no I/O.  Every function is comptime-compatible.
+//! All overflow paths return `null` (`?T`) instead of panicking.
```
wemeetagain: What's the motivation here returning nulls instead if error unions?

GrapeBaBa: Almost all the overflow here is effectively unreachable, so I thought it was not worth to pollute a lot of function signatures to return error union. But it is probably we should just use `+/-` to instead of `math.add/math.sub`

wemeetagain (approved):

nazarhussain:
One of my major design concern on the design was the separation between `EventClock` and `SlotClock` which makes it difficult to reason about the `clock` which we used to have. The `SlotClock` mostly about mathematics can be moved to slot math, so we end of up with one user facing object of `Clock`. 

Created one issue for it. https://github.com/ChainSafe/lodestar-z/issues/457

PR #356, opened 2026-05-12
feat(bindings/bls): use external buffers for blst operations

spiral-ladder:
This is one of possible likely causes for increased GC pressure on experiments to swap out blst-ts for lodestar-z/bls, as observed on feat2 and feat3 deployments in [this PR](https://github.com/ChainSafe/lodestar/pull/9342/).

With external array buffers, V8 is only aware of the pointer to the backing memory, instead of having to track both the pointer and the backing memory. This means that during marking phase the GC does not have to walk the backing memory to mark it as 'live' - the frequency of the GC firing off is still the same, but each cycle does less work.

This of course comes with a tradeoff, we need a **finalizer** to let V8 know how much external memory is in native heap  so that the GC tells the native impl to free the useless memory.

Though, regardless of the effect, we should still probably do this anyway, since [napi-rs does the same](https://github.com/napi-rs/napi-rs/blob/159395b365c583a6642ad481edc5708d9f36a24b/crates/napi/src/bindgen_runtime/js_values/arraybuffer.rs#L175), and only defaults to V8 managed array buffers if it is disallowed (like in Electron).

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -137,6 +137,41 @@ fn coerceToBool(boolish: napi.Value) napi.status.NapiError!bool {
     return b.getValueBool();
 }
 
+/// Wraps `bytes` as a JS `Uint8Array` backed by an *external* (native-heap) ArrayBuffer.
+///
+/// V8 holds the pointer to manage the buffer's lifetime, but the buffer must still
+/// be freed natively via a finalizer.
+///
+/// ## Details
+///
+/// Heap-allocation rate & GC frequency increases when using `napi_create_arraybuffer`
+/// since V8 owns both the pointer and the payload. We rely on external buffers here
+/// to lighten the pressure on V8's GC engine.
+fn createExternalUint8Array(env: napi.Env, bytes: []const u8) !napi.Value {
+    const buf = try allocator.alloc(u8, bytes.len);
+    @memcpy(buf, bytes);
+
+    const len_hint: ?*anyopaque = @ptrFromInt(bytes.len);
+    const arraybuffer = env.createExternalArrayBuffer(buf, externalUint8Finalize, len_hint) catch |err| {
+        allocator.free(buf);
+        return err;
+    };
+
+    _ = try env.adjustExternalMemory(@intCast(bytes.len));
+    return try env.createTypedarray(.uint8, bytes.len, arraybuffer, 0);
+}
+
+fn externalUint8Finalize(env: napi.c.napi_env, data: ?*anyopaque, hint: ?*anyopaque) callconv(.c) void {
+    const len: usize = @intFromPtr(hint);
+    if (data) |d| {
+        const ptr: [*]u8 = @ptrCast(d);
+        allocator.free(ptr[0..len]);
+    }
+    var result: i64 = undefined;
+
+    _ = napi.status.check(napi.c.napi_adjust_external_memory(env, -@as(i64, @intCast(len)), &result)) catch return {};
+}
+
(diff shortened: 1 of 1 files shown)


bindings/napi/blst.zig
```diff
@@ -137,6 +137,41 @@ fn coerceToBool(boolish: napi.Value) napi.status.NapiError!bool {
...
+/// be freed natively via a finalizer.
+///
+/// ## Details
+///
+/// Heap-allocation rate & GC frequency increases when using `napi_create_arraybuffer`
+/// since V8 owns both the pointer and the payload. We rely on external buffers here
+/// to lighten the pressure on V8's GC engine.
+fn createExternalUint8Array(env: napi.Env, bytes: []const u8) !napi.Value {
+    const buf = try allocator.alloc(u8, bytes.len);
+    @memcpy(buf, bytes);
```
wemeetagain, line 152: can use allocator.dupe as a shorthand

bindings/napi/blst.zig
```diff
@@ -137,6 +137,41 @@ fn coerceToBool(boolish: napi.Value) napi.status.NapiError!bool {
     return b.getValueBool();
 }
 
+/// Wraps `bytes` as a JS `Uint8Array` backed by an *external* (native-heap) ArrayBuffer.
```
wemeetagain, line 140: I'd change the comment here to mentioning duping, copying, etc. Current comment sounds like `bytes` is being captured.

bindings/napi/blst.zig
```diff
@@ -137,6 +137,41 @@ fn coerceToBool(boolish: napi.Value) napi.status.NapiError!bool {
...
+/// Heap-allocation rate & GC frequency increases when using `napi_create_arraybuffer`
+/// since V8 owns both the pointer and the payload. We rely on external buffers here
+/// to lighten the pressure on V8's GC engine.
+fn createExternalUint8Array(env: napi.Env, bytes: []const u8) !napi.Value {
+    const buf = try allocator.alloc(u8, bytes.len);
+    @memcpy(buf, bytes);
+
+    const len_hint: ?*anyopaque = @ptrFromInt(bytes.len);
+    const arraybuffer = env.createExternalArrayBuffer(buf, externalUint8Finalize, len_hint) catch |err| {
+        allocator.free(buf);
```
wemeetagain, line 156: you should use an `errdefer allocator.free(buf)` above, and call `env.createExternalArrayBuffer` with `try`.

bindings/napi/blst.zig
```diff
@@ -137,6 +137,41 @@ fn coerceToBool(boolish: napi.Value) napi.status.NapiError!bool {
...
+
+fn externalUint8Finalize(env: napi.c.napi_env, data: ?*anyopaque, hint: ?*anyopaque) callconv(.c) void {
+    const len: usize = @intFromPtr(hint);
+    if (data) |d| {
+        const ptr: [*]u8 = @ptrCast(d);
+        allocator.free(ptr[0..len]);
+    }
+    var result: i64 = undefined;
+
+    _ = napi.status.check(napi.c.napi_adjust_external_memory(env, -@as(i64, @intCast(len)), &result)) catch return {};
```
wemeetagain, line 172: don't need `result` or `napi.status.check` since neither are intended to be used.

wemeetagain (changes requested):

spiral-ladder:
@wemeetagain thanks for the review! Though this is largely irrelevant now since the layer that handles creating of array buffers is moved to zapi, will open a PR there instead

PR #358, opened 2026-05-13
feat(blst): use external buffers for blst operations

spiral-ladder:
Depends on ChainSafe/zapi#30 (we need to release zapi and update the dep)

This is one of possible likely causes for increased GC pressure on experiments to swap out blst-ts for lodestar-z/bls, as observed on feat2 and feat3 deployments in [this
PR](https://github.com/ChainSafe/lodestar/pull/9342/).

With external array buffers, V8 is only aware of the pointer to the backing memory, instead of having to track both the pointer and the backing memory. This means that during marking phase the GC does not have to walk the backing memory to mark it as 'live' - the frequency of the GC firing off is still the same, but each cycle does less work.

This of course comes with a tradeoff, we need a
**finalizer** to let V8 know how much external memory is in native heap  so that the GC tells the native impl to free the useless memory.

Though, regardless of the effect, we should still probably do this anyway, since [napi-rs does the same](https://github.com/napi-rs/napi-rs/blob/159395b365c583a6642ad481edc5708d9f36a24b/crates/napi/src/bindgen_runtime/js_values/arraybuffer.rs#L175), and only defaults to V8 managed array buffers if it is disallowed (like in Electron).

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -136,10 +136,10 @@ pub const PublicKey = struct {
     pub fn toBytes(self: *const PublicKey, compress: ?js.Boolean) !js.Uint8Array {
         if (try boolOrDefault(compress, true)) {
             const bytes = self.raw.compress();
-            return js.Uint8Array.from(bytes[0..]);
+            return js.Uint8Array.fromExternal(bytes[0..]);
         }
         const bytes = self.raw.serialize();
-        return js.Uint8Array.from(bytes[0..]);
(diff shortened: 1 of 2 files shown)


spiral-ladder:
The gemini reviews are inaccurate, since the data is indeed copied (but hidden in the [zapi API](https://github.com/ChainSafe/zapi/pull/30/changes#diff-53d58f314219749e5378259c5d7845bb5bc8134162fdc0727d5183acc65232f1R62))

PR #364, opened 2026-05-16
chore: support dev workflow

wemeetagain:
- Motivation: https://github.com/ChainSafe/lodestar-z/pull/360#issuecomment-4462314964

Afaik, zig-out/lib/* must be in package.json `files` in order for the library to be part of the packed package when installing lodestar-z as a git dependency.
Unfortunately, that also allows it to be part of the package when _published_!! which breaks cross-platform usage of the library, short circuiting the correct library from being loaded.

To that end, this PR attempts to get the intended behavior for both cases:
- keep zig-out/lib in `files` so git-dependency-installed lodestar-z packs a freshly-built library
- but crucially suppress the prepare script from being run during publish via `--ignore-scripts` so npm-dependency-installed lodestar-z uses a platform-specific published library

### DIFF
--- a/.github/workflows/publish-bindings.yml
+++ b/.github/workflows/publish-bindings.yml
@@ -62,5 +62,5 @@ jobs:
         run: |
           pnpm zapi build-artifacts --optimize ReleaseSafe
           pnpm zapi prepublish
-          pnpm zapi publish --access public -- --provenance ${{ steps.detect_dist_tag.outputs.tag_flag }}
+          pnpm zapi publish --access public -- --ignore-scripts --provenance ${{ steps.detect_dist_tag.outputs.tag_flag }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-bindings.yml
```diff
@@ -62,5 +62,5 @@ jobs:
         run: |
           pnpm zapi build-artifacts --optimize ReleaseSafe
           pnpm zapi prepublish
```
wemeetagain, line 64:
```suggestion
          pnpm zapi prepublish
          rm -rf zig-out/lib
```

.github/workflows/publish-bindings.yml
```diff
@@ -62,5 +62,6 @@ jobs:
...
-          pnpm zapi publish --access public -- --provenance ${{ steps.detect_dist_tag.outputs.tag_flag }}
+          rm -rf zig-out/lib
```
spiral-ladder, line 65:
```suggestion
          # We do this because: https://github.com/ChainSafe/lodestar-z/pull/364
          rm -rf zig-out/lib
```

spiral-ladder (approved):
would be nice to have context for future readers in the workflow file itself, otherwise lgtm

PR #371, opened 2026-05-22
fix: export asyncAggregateWithRandomness through napi binding

nazarhussain:
### Summary

`asyncAggregateWithRandomness` was declared as a `pub fn` in `bindings/napi/blst.zig` and re-exported from `bindings/src/blst.js`, but at runtime the JS-side binding was `undefined`. This PR rewrites the function to use the zapi DSL so it is actually registered as a JS export.

### Root cause

zapi's `exportModule` auto-registration (`export_module.zig`, `wrap_function.zig:isDslOrOptionalDsl`) only picks up a `pub fn` when **all** of its parameters are DSL types (`napi.Value`, DSL wrappers like `js.Array`, or class types). The previous signature

  ```zig
  pub fn asyncAggregateWithRandomness(env: napi.Env, cb: napi.CallbackInfo(1)) !napi.Value
```

uses raw NAPI types, so the function was silently skipped at registration. Because Zig is lazy about analyzing unreferenced functions, this also hid several internal bugs in the body that would otherwise have failed to compile.

### Test plan

- `pnpm test` 
- `zig build build-lib:bindings -Doptimize=ReleaseSafe` builds cleanly.

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -710,23 +710,23 @@ fn asyncAggRand_execute(_: napi.Env, data: *AsyncAggRandData) void {
 /// Ran on the JS thread once the worker has finished. Always settles the
 /// promise — `settle` does the resolve/reject; if it errors we fall back to a
 /// bare reject so callers never see a dangling Promise.
-fn asyncAggRand_complete(env: napi.Env, stat: napi.status.Status, data: *AsyncAggRandData) void {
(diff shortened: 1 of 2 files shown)


spiral-ladder (dismissed):
LGTM, gemini reviews look pretty pedantic to me

spiral-ladder (approved):

Issue #357, opened 2026-05-12
Memory Safety Review


twoeths:
## Summary

A targeted memory-safety review of `state-transition-z` covering the SSZ type layer, tree-view layer, persistent merkle tree, and state transition. Findings below are bugs that are **not caught by current tests** — most require allocator-fault injection, malformed-peer input, or specific call-sequences that no test exercises.

Findings are grouped by severity. Every section links to the file and lines.

---

### C1. `errdefer pool.free(&nodes)` over an `undefined` `Node.Id` array

Files / lines:
- `src/ssz/type/container.zig:151-160`, `:168-180`, `:192-200`, `:600-654`
- `src/ssz/type/vector.zig:158-227`, `:430-465`

```zig
var nodes: [chunk_count]Node.Id = undefined;
errdefer pool.free(&nodes);                       // iterates UNDEFINED ids
inline for (fields, 0..) |field, i| {
    nodes[i] = try field.type.tree.deserializeFromBytes(pool, ...);
}
```

If any field's `deserializeFromBytes` / `fromValue` / `default` errors at iteration K, indices `K..chunk_count` are uninitialized. `pool.free` calls `unref` on each. In Debug/Safe builds, `undefined` is `0xAA…` (caught by the bounds check in `unref`). **In `ReleaseFast`, `undefined` is whatever stack garbage is there — easily a valid live `Node.Id`** — silently decrementing an unrelated node's refcount and producing a delayed UAF / double-free.

**Fix:** track `var filled: usize = 0` and `errdefer pool.free(nodes[0..filled])`, or zero-fill before the loop (zero nodes are no-ops in `unref`).

---

### C2. `Pool.alloc` permanently leaks slots on `preheat` OOM (acknowledged TODO)

`src/persistent_merkle_tree/Node.zig:254-280`

```zig
if (@intFromEnum(self.next_free_node) == self.nodes.len) {
    const remaining = out.len - i;
    try self.preheat(@intCast(remaining));
    // TODO how to handle failing to resize here
    // errdefer self.free(out[0..i]);
```

On `preheat` OOM, slots in `out[0..i]` are popped from the free list and marked `branch_lazy(refcount=0)`, but the caller never receives them and they are never freed. Each failed `alloc` permanently shrinks the pool. Long-running nodes under intermittent memory pressure bleed capacity until exhaustion.

[...]

PR #377, opened 2026-05-25
fix: harden memory safety across PMT, SSZ tree views, and state transition

GrapeBaBa:
## Summary

Addresses the #357 memory-safety review. The findings are error-path bugs that
the happy path never exercises: `errdefer`/`defer` cleanup running over
`undefined` or already-moved memory, OOM leaving aliased refcounted pointers,
orphaned pool nodes on rollback, and an out-of-bounds proposer index from
untrusted input. All fixes are allocate/validate-before-commit so a failure
can't leave shared state corrupt.

## Changes

**persistent_merkle_tree**
- Zero-fill node-id buffers at `errdefer` sites so a mid-build error's
  `unref`/`free` no-ops on the unbuilt tail instead of unref-ing stack garbage (C1)
- `Pool.alloc` rolls back already-popped slots on a preheat OOM (C2)
- `createBranch` rolls back an applied child ref on `RefCountOverflow` (M2)
- `FillWithContentsIterator` reclaims both orphaned `left` and `carry` on a
  `createBranch` OOM, guarding the all-default aliased-node case (M10)
- `setNodes*` start `unfinalized_parents` all-null so the unref loop can't read
  an undefined `?Id` (H1); empty-`indices` no-op guarded before the ascending assert
- `View.destroy` asserts no live children + poisons the recycled slot (H2)

**ssz tree views**
- `commit`/`set` reserve capacity up front so the stores are infallible (H4/M5)
- `setValue` no longer double-frees the child view on `set`'s OOM; documented the
  borrowed-pointer invalidation contract on `get`/`set`/`getReadonly`/`clone` (C5)
- `sliceTo` unrefs its intermediate orphan roots (H6)
- `ReadonlyIterator.nextValue` initializes `out` before `toValue` (H5)

**state_transition**
- Epoch-cache shuffling rotation and effective-balance/pubkey updates allocate
  before mutating shared `Rc`/maps, so OOM can't leave an aliased double-unref (C4)
- `epoch_transition_cache` errdefers for its temporary lists (M8)
- Proposer signature set bounds-checks the proposer index (`verifyProposerSignature`
  runs before `processBlockHeader` validates it) (C3)
- `effectiveBalanceIncrementsSet` allocates the buffer with the same allocator the
  owning `Rc` frees it with
- Signature-set `out` params changed from by-value `ArrayList` to `*ArrayList`


GrapeBaBa:
@codex review

GrapeBaBa:
@claude review

GrapeBaBa:
@codex review

src/persistent_merkle_tree/Node.zig
```diff
@@ -611,6 +630,11 @@ pub const Id = enum(u32) {
         if (indices.len == 0) {
             return root_node;
         }
+        // Callers must pass strictly-ascending indices; unsorted or duplicate input silently
+        // corrupts the tree. Check it in safe builds, after the empty guard (`for (1..0)` panics).
+        if (builtin.mode == .Debug or builtin.mode == .ReleaseSafe) {
```
wemeetagain: I think the mode check is not necessary (a noop in a loop should get optimized away)

wemeetagain (dismissed):

wemeetagain (approved):

PR #383, opened 2026-05-28
refactor(bindings): wake only required number of workers

spiral-ladder:
introduces

- `sleeping_workers`: mark itself as asleep when no work is available, mark itself as not a `sleeping_worker` (i.e. `sleeping_workers -= 1` when it sees work
- use `signal` instead of `broadcast`

I suspect contention on threads causing regression on bls workload. This at least metrics wise on `feat3-sas` has some good effects (white line is redeployment):

Screenshot 2026-05-28 at 9 30 20 PM

Job wait time is also trending down:

Screenshot 2026-05-28 at 9 32 18 PM

### DIFF
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -60,6 +60,7 @@ const JobQueue = struct {
     cond: std.Io.Condition = std.Io.Condition.init,
     head: ?*WorkItem = null,
     tail: ?*WorkItem = null,
+    sleeping_workers: usize = 0,
 
     /// Pushes a batch of `WorkItem`s to the `JobQueue`.
     ///
@@ -80,7 +81,7 @@ const JobQueue = struct {
             }
             self.tail = item;
         }
-        for (0..@min(items.len, pool.n_workers)) |_| {
+        for (0..@min(items.len, self.sleeping_workers)) |_| {
             self.cond.signal(io);
         }
         return true;
@@ -163,7 +164,9 @@ fn workerLoop(pool: *ThreadPool, io: std.Io) void {
             while (true) {
                 if (pool.queue.pop()) |wi| break :blk wi;
                 if (pool.shutdown.load(.acquire)) return;
+                pool.queue.sleeping_workers += 1;
                 pool.queue.cond.waitUncancelable(io, &pool.queue.mutex);
+                pool.queue.sleeping_workers -= 1;
             }
         };
 
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -80,7 +80,9 @@ const JobQueue = struct {
             }
             self.tail = item;
         }
-        self.cond.broadcast(io);
+        for (0..@min(items.len, pool.n_workers)) |_| {
+            self.cond.signal(io);
+        }
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


src/bls/ThreadPool.zig
```diff
@@ -80,7 +81,9 @@ const JobQueue = struct {
             }
             self.tail = item;
         }
-        self.cond.broadcast(io);
+        for (0..@min(items.len, self.sleeping_workers)) |_| {
+            self.cond.signal(io);
+        }
```
spiral-ladder, line 94: this is really the key change. I was doing something stupid with waking up all workers using `broadcast` instead of waking only the necessary num of workers. the sleep portion is just a further optimization

src/bls/ThreadPool.zig
```diff
@@ -60,6 +60,7 @@ const JobQueue = struct {
     cond: std.Io.Condition = std.Io.Condition.init,
     head: ?*WorkItem = null,
     tail: ?*WorkItem = null,
+    sleeping_workers: usize = 0,
```
wemeetagain, line 67: The `sleeping_workers` field carries a non-trivial concurrency invariant but ships without a doc comment, unlike its neighbors in this file (`shutdown` vs `shutting_down`, the `workerLoop` "pop before shutdown" note, the `submitAndWait` uncancelable note). Suggest documenting the invariant so the next reader doesn't have to reconstruct it:

```suggestion
    /// Count of workers currently blocked in `cond.wait`. Guarded by `mutex`
    /// (read in `pushBatch`, maintained in `workerLoop`), so it is exact at
    /// signal time. Lets `pushBatch` wake only as many workers as there is new
    /// work for, instead of broadcasting to all of them.
    sleeping_workers: usize = 0,
```

src/bls/ThreadPool.zig
wemeetagain, line 94: This replaced a one-line `broadcast` with a subtler counted-signal policy whose correctness argument is non-obvious: one wakeup already suffices (a running worker drains the whole queue), and the `@min` is what avoids the thundering herd. A short comment brings it up to the documentation bar set by the rest of this file:

```suggestion
        // Wake at most one worker per new item, and never more than are actually
        // asleep — avoids the broadcast thundering-herd where workers wake only
        // to lose the race for `job.counter` and immediately re-sleep. One wakeup
        // is enough for correctness (a running worker drains the whole queue);
        // the rest are for parallelism.
        for (0..@min(items.len, self.sleeping_workers)) |_| {
            self.cond.signal(io);
        }
```

wemeetagain (commented):
Two small documentation suggestions for the wake-only-required-workers change. The logic itself is correct and a solid improvement over `broadcast` — these just bring the new field and the counted-signal loop up to the concurrency-documentation standard already set elsewhere in this file (e.g. the `shutdown`/`shutting_down` ordering and the `submitAndWait` uncancelable notes).

src/bls/ThreadPool.zig
spiral-ladder: I rephrased this a bit bc work items don't necessarily use a counter as an invariant, rather workers loop back to `pop()` after each item

wemeetagain (approved):

PR #320, opened 2026-04-21
feat: use high level zapi dsl

nazarhussain:
Use the high level DSL introduced in zapi 1.0.0

Excluding transitioning of following modules from this PR. Will open separate PR for each for more clarity and understanding.  
- `pubkeys`
- `blst` 
- `BeaconStateView`

PR #386, opened 2026-06-05
fix(bindings): size BLS thread pool by cgroup-aware CPU count

GrapeBaBa:
## Motivation

`std.Thread.getCpuCount()` only reads the CPU affinity mask, so under a cgroup CPU quota (`docker --cpus=N` / k8s `limits.cpu`) it reports the host core count rather than the quota. That over-sizes the BLS verification thread pool at NAPI init, which can cause thread oversubscription and CFS throttling in CPU-limited containers.

## Changes

- Add `src/cpu_count.zig` — `getNumCpus(gpa, io)` returns `min(cgroup quota, affinity)`, locating the cpu controller via `/proc/self/{cgroup,mountinfo}` and reading `cpu.max` / `cpu.cfs_quota_us` (cgroup v1 + v2). Parsing logic ported from the `num_cpus` crate.
- The quota is the **minimum along the cgroup ancestor chain** (leaf up to the mount point), matching Rust `std::thread::available_parallelism`: the kernel enforces every level but each level's file reports only its own limit, so leaf-only reads miss LXC/Proxmox `cpulimit`, systemd `CPUQuota=` on a parent slice, and sub-cgroups inside a limited container.
- Hardened beyond `num_cpus`: cgroup paths containing `:` (containerd's `…slice:cri-containerd:<id>` cgroupfs naming) are kept intact, and `..` paths (process outside its cgroupns root, `cgroup_namespaces(7)`) fall back cleanly instead of erroring.
- Use it at NAPI thread-pool init (`bindings/napi/root.zig`) instead of `std.Thread.getCpuCount()`.
- Run `test:cpu_count` in CI (the build-test job enumerates module test steps explicitly).

Notes:
- `/proc` pseudo-files report size 0, so they are streamed to EOF (`readerStreaming` + `allocRemaining`), not read by stat size (`readFileAlloc` reads empty).
- Error policy: `getNumCpus` itself is fail-fast — a genuinely **absent** quota (non-Linux, no cpu controller, no cgroup mount, unresolvable path, unlimited) is `null` → affinity fallback, while a **broken read** of an existing resource (unreadable `/proc` or quota file, unopenable cgroup dir, malformed content) is an error, so a readable quota is never silently masked. The NAPI call site catches detection errors, logs a warning, and sizes by the affinity count — a sizing probe must not prevent the module from loading.
- `ceilDiv` is overflow-safe; the allocator is passed in by the caller (the module stays allocator-agnostic / libc-free).

## Testing

- 22 unit tests covering the parse-and-resolve pipeline and the ancestor walk; test vectors adapted from `seanmonstar/num_cpus` (MIT).
- Verified end-to-end in Linux containers (OrbStack): `--cpus=2` → 2, `--cpus=4` → 4, `--cpus=1.5` → 2 (ceil), unlimited → host count. The old `getCpuCount()` returned the host count in all cases.
- Ancestor walk verified end-to-end: with `--cpus=2` and the process moved into an unconstrained child cgroup (`0::/child`, no `cpu.max` of its own), the walk finds the limit at the cgroupns root and returns 2 — leaf-only reads return the host count here.
- Colon handling verified end-to-end: a process in `0::/x.slice:cri-containerd:y` resolves its cgroup dir and returns the quota.

---
🤖 This PR was developed with AI assistance (Claude Code).

### DIFF
--- a/bindings/napi/root.zig
+++ b/bindings/napi/root.zig
@@ -1,5 +1,4 @@
 const std = @import("std");
-const builtin = @import("builtin");
 const js = @import("zapi:zapi").js;
 pub const pool = @import("./pool.zig");
 pub const shuffle = @import("./shuffle.zig");
@@ -13,9 +12,6 @@ pub const pubkeys = @import("./pubkeys.zig");
 const options = @import("bls_options");
 const napi_io = @import("./io.zig");
 
-var gpa: std.heap.DebugAllocator(.{}) = .init;
-const allocator = if (builtin.mode == .Debug) gpa.allocator() else std.heap.c_allocator;
-
 fn init(old_ref_count: u32) !void {
     if (old_ref_count == 0) {
         // First environment — initialize shared state in your threadpool init.
@@ -24,8 +20,13 @@ fn init(old_ref_count: u32) !void {
 
         var cpu_count: u64 = options.thread_count;
         if (options.thread_count == 0) {
-            cpu_count = @max(try @import("cpu_count").getNumCpus(allocator, napi_io.get()), 2) - 1;
-            std.debug.print("Note: no -Dthread-count set, using cgroup-aware CPU count minus 1: {}\n", .{cpu_count});
+            const detected =
+                try @import("cpu_count").getNumCpus(std.heap.page_allocator, napi_io.get());
+            cpu_count = @max(detected, 2) - 1;
+            std.debug.print(
+                "Note: no -Dthread-count set, using cgroup-aware CPU count minus 1: {}\n",
+                .{cpu_count},
+            );
         }
 
         const n_workers = @min(cpu_count, @import("bls").ThreadPool.MAX_WORKERS);
@@ -44,7 +45,6 @@ fn cleanup(new_ref_count: u32) void {
         pubkeys.state.deinit();
         pool.state.deinit();
         metrics.deinit();
-        blst.deinitThreadPool();
         napi_io.deinit();
     }
 }
--- a/bindings/napi/root.zig
+++ b/bindings/napi/root.zig
@@ -24,7 +24,7 @@ fn init(old_ref_count: u32) !void {
 
         var cpu_count: u64 = options.thread_count;
         if (options.thread_count == 0) {
-            cpu_count = @max(@import("cpu_count").getNumCpus(allocator, napi_io.get()), 2) - 1;
+            cpu_count = @max(try @import("cpu_count").getNumCpus(allocator, napi_io.get()), 2) - 1;
             std.debug.print("Note: no -Dthread-count set, using cgroup-aware CPU count minus 1: {}\n", .{cpu_count});
         }
 
--- a/bindings/napi/root.zig
+++ b/bindings/napi/root.zig
@@ -21,7 +21,7 @@ fn init(old_ref_count: u32) !void {
         var cpu_count: u64 = options.thread_count;
         if (options.thread_count == 0) {
             const detected =
-                try @import("cpu_count").getNumCpus(std.heap.page_allocator, napi_io.get());
+                try @import("cpu_count").getNumCpus(std.heap.c_allocator, napi_io.get());
             cpu_count = @max(detected, 2) - 1;
             std.debug.print(
                 "Note: no -Dthread-count set, using cgroup-aware CPU count minus 1: {}\n",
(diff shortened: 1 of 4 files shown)


GrapeBaBa:
@codex review

GrapeBaBa:
@codex review

bindings/napi/root.zig
```diff
@@ -20,8 +24,8 @@ fn init(old_ref_count: u32) !void {
 
         var cpu_count: u64 = options.thread_count;
         if (options.thread_count == 0) {
-            cpu_count = @max((try std.Thread.getCpuCount()) - 1, 1);
-            std.debug.print("Note: no -Dthread-count set, will use runtime CPU count minus 1: {}\n", .{cpu_count});
+            cpu_count = @max(try @import("cpu_count").getNumCpus(allocator, napi_io.get()), 2) - 1;
```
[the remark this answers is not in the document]
GrapeBaBa: we already fall back to `std.Thread.getCpuCount()` inside `getNumCpus` @codex

bindings/napi/root.zig
```diff
@@ -40,7 +48,6 @@ fn cleanup(new_ref_count: u32) void {
         pubkeys.state.deinit();
         pool.state.deinit();
         metrics.deinit();
-        blst.deinitThreadPool();
```
spiral-ladder, line 43: is this intentional?

src/cpu_count.zig
```diff
@@ -0,0 +1,723 @@
+//! cgroup-aware logical CPU count for Linux.
+//!
+//! `std.Thread.getCpuCount()` only reads the CPU affinity mask and is blind to
+//! the cgroup CFS quota (`cpu.max` / `cpu.cfs_quota_us`), so under
+//! `docker --cpus=N` or k8s `limits.cpu` it reports the host core count. This
+//! returns `min(quota, affinity)` instead, where the quota is the smallest one
+//! along the cgroup ancestor chain — the kernel enforces every level, but each
+//! level's quota file reports only its own limit.
```
spiral-ladder, line 8: It would be nice to also include where we are consuming this:

```suggestion
//! cgroup-aware logical CPU count for Linux.
//!
//! `std.Thread.getCpuCount()` only reads the CPU affinity mask and is blind to
//! the cgroup CFS quota (`cpu.max` / `cpu.cfs_quota_us`), so under
//! `docker --cpus=N` or k8s `limits.cpu` it reports the host core count. This
//! returns `min(quota, affinity)` instead, where the quota is the smallest one
//! along the cgroup ancestor chain — the kernel enforces every level, but each
//! level's quota file reports only its own limit.
//!
//! This module is important when consumed by docker deployed applications using
//! multithreaded operations within `lodestar-z`, eg.
//! our threadpool implementation used for bls operations in `bls/ThreadPool.zig`.
```

src/cpu_count.zig
```diff
@@ -0,0 +1,723 @@
...
+    return result;
+}
+
+/// Logical CPU count from the affinity mask. Propagates the error rather than
+/// silently degrading — a failed affinity count would size the pool to 1 worker.
+fn logicalCpus() !usize {
+    const n = try std.Thread.getCpuCount();
+    assert(n >= 1);
+    return n;
+}
```
spiral-ladder: we can do this inline and assert `n >= 1` in `getNumCpus` and also get rid of the test, this is just doing an unnecessary abstraction on `std.Thread.getCpuCount()` imo

spiral-ladder (commented):
looks good, some comments

bindings/napi/root.zig
GrapeBaBa: Yes — intentional. Both calls came in together with the #320 zapi rewrite (merge artifact), and the first position is the correct one to keep: `ThreadPool.deinit` drains the queue and joins all workers, which must happen *before* the `pubkeys`/`pool` caches (which in-flight jobs read) are freed, and before `napi_io.deinit()` (pool teardown uses the io). The removed second call was a guaranteed idempotent no-op.

src/cpu_count.zig
GrapeBaBa: Added in 62b14945 (condensed wording).

src/cpu_count.zig
GrapeBaBa: Done in 62b14945 — inlined with the `n >= 1` assert moved into `getNumCpus`, and dropped the "never exceeds logical" test since the `result <= logical` postcondition assert now checks it on every call (including from the remaining smoke test).

spiral-ladder (approved):
In general looks good but this should really be a library on its own, but we can leave that for a followup cleanup, let's merge this so we can cut a release and deploy bls updates

PR #396, opened 2026-06-09
fix: node init pool

spiral-ladder:


spiral-ladder:
superceded by https://github.com/ChainSafe/lodestar-z/pull/394

PR #397, opened 2026-06-09
feat(bindings): aggregate cached public keys by validator index

spiral-ladder:
Instead of getting the pk from cache and then passing it back through napi, we just pass the already known indices into the native layer and let the cache do the work.

On `feat3-super`, this halved pk aggregation time (8s to 4s):

Screenshot 2026-06-09 at 5 26 16 PM

Notably, sig set verification time per set went up (530us -> 600us), but I suspect this is due to it becoming the bottleneck now. Job wait time has gone down (37-38ms to about 31-32ms), and we're doing less verifications in batches (88% to 87% roughly)

Screenshot 2026-06-09 at 5 27 09 PM

block processing looks better with this:

Screenshot 2026-06-09 at 5 29 44 PM

verification metrics across the board trending down with this:

Screenshot 2026-06-09 at 5 30 39 PM

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -11,6 +11,7 @@ const napi_io = @import("./io.zig");
 const allocator = std.heap.page_allocator;
 
 const default_initial_capacity: u32 = 0;
+const max_stack_aggregate_pubkeys = 512;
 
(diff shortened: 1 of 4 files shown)


GrapeBaBa (approved):
LGTM

PR #438, opened 2026-06-22
fix: do not set start index for deposit request in non-electra forks

spiral-ladder:
This was causing a spec test failure

See: https://github.com/ChainSafe/lodestar/blob/f6b2af6879db8092cdaf0dad41b3a098ff391a4a/packages/state-transition/src/block/processDepositRequest.ts#L133


spiral-ladder:
for some reason native spec tests pass despite this but via bindings it fails. So something's up with lodestar-z

spiral-ladder:
false positive on a test

PR #441, opened 2026-06-22
chore: update spec-test-version: v1.6.0-beta.2 -> v1.7.0-alpha.10

spiral-ladder:
we're integrating stf into lodestar soon, so let's keep the versions in sync.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -105,7 +105,7 @@
             },
             .spec_test_version = .{
                 .type = .string,
-                .default = "v1.6.0-beta.2",
+                .default = "v1.7.0-alpha.10",
                 .description = "Consensus spec tests version tag.",
             },
             .spec_test_out_dir = .{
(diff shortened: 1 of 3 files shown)


ensi321 (commented):
I think alpha.11 is the latest version. Should target that instead. Also I think there is couple of refactoring changes to pre-gloas specs between the two versions that it is good to have lodestar-z aligned with the spec even though we are passing the spec tests.

spiral-ladder:
> I think alpha.11 is the latest version. Should target that instead. Also I think there is couple of refactoring changes to pre-gloas specs between the two versions that it is good to have lodestar-z aligned with the spec even though we are passing the spec tests.

definitely, but I think we should wait on https://github.com/ChainSafe/lodestar/pull/9541 for that, otherwise the mismatch might possibly cause failures on chainsafe/lodestar#9516

Issue #405, opened 2026-06-15
Implement BeaconEngine


twoeths:
Given an `IBeaconEngine` interface to be implemented in the current lodestar https://github.com/ChainSafe/lodestar/issues/9508, we should start building a native BeaconEngine following that interface. Once this is ready, we can switch the BeaconEngine part of lodestar to use this native implementation.

Tracked issues: TBD
- https://github.com/ChainSafe/lodestar-z/issues/406
- https://github.com/ChainSafe/lodestar-z/issues/407
- https://github.com/ChainSafe/lodestar-z/issues/408
- https://github.com/ChainSafe/lodestar-z/issues/409
- https://github.com/ChainSafe/lodestar-z/issues/410
- https://github.com/ChainSafe/lodestar-z/issues/411
- https://github.com/ChainSafe/lodestar-z/issues/412
- https://github.com/ChainSafe/lodestar-z/issues/413
- https://github.com/ChainSafe/lodestar-z/issues/414
- https://github.com/ChainSafe/lodestar-z/issues/415
- https://github.com/ChainSafe/lodestar-z/issues/416
- https://github.com/ChainSafe/lodestar-z/issues/417
- https://github.com/ChainSafe/lodestar-z/issues/418
- https://github.com/ChainSafe/lodestar-z/issues/419
- https://github.com/ChainSafe/lodestar-z/issues/420
- https://github.com/ChainSafe/lodestar-z/issues/421
- https://github.com/ChainSafe/lodestar-z/issues/422
- https://github.com/ChainSafe/lodestar-z/issues/423
- https://github.com/ChainSafe/lodestar-z/issues/424
- https://github.com/ChainSafe/lodestar-z/issues/425
- https://github.com/ChainSafe/lodestar-z/issues/426
- https://github.com/ChainSafe/lodestar-z/issues/427
- https://github.com/ChainSafe/lodestar-z/issues/428
- https://github.com/ChainSafe/lodestar-z/issues/429
- https://github.com/ChainSafe/lodestar-z/issues/430

PR #446, opened 2026-06-24
feat: implement logging module

### LINKED ISSUE #413: Implement logging module
As part of the BeaconEngine in #405, we need a logging module
The log level will be configured in the current BeaconChain, and be specified when we create this BeaconEngine

markolazic01:
### Description

Native Zig logging module for the BeaconEngine (#405), wired in as a log module the consensus modules can import. Built standalone rather than forwarding from the JS side (used metrics module as a reference). Used the Lodestar logger as a behavioral reference.

Closes #413

### Included

- stderr output with per-level gating
- 6 log levels matching Lodestar's `LogLevel` (err/warn/info/verbose/debug/trace) — kept identical since the level is injected from the JS side
- string-based module tags via a `ModuleLogger` view — chose strings over an enum for open extensibility and to map onto Lodestar's string-keyed levelModule
- monotonic since-start timestamps
- ANSI color on the level

### Deferred to follow-up PRs
- file output + daily rotation
- wall-clock timestamps — deferred because it touches the shared time.zig, so it belongs in its own PR
- dynamic / per-module level changes
- TTY-gated color (currently colors unconditionally)
- context appended as key=value pairs

### Open question

The logger calls `defaultIo()` internally rather than taking an `Io` at construction (matches Zeam log module). Happy to thread an `Io` through init instead if you'd prefer the engine to own it.

If anything else needs to be changed or you want some of the deferred topics to be included in here please let me know.

### Testing

`$ zig build test:log` covers level gating, line formatting (with and without a module tag), and a construction smoke test.

### AI

Used Claude to find proper patterns and audit code, present code including the tests is authored manually.


spiral-ladder:
@markolazic01 please also add a job to `.github/CI.yml` testing this module. It fails with

```
➜ zig build test:log
test:log
└─ run test log
   └─ compile test log Debug native 1 errors
src/log.zig:81:26: error: root source file struct 'time' has no member named 'timestampNow'
            .start = time.timestampNow(defaultIo()),
                     ~~~~^~~~~~~~~~~~~
src/time.zig:1:1: note: struct declared here
const std = @import("std");
^~~~~
referenced by:
    test.log smoke: src/log.zig:152:32
error: 1 compilation errors
```

at the moment

markolazic01:
Hey @spiral-ladder, I adapted `log` to `time.zig` changes and added `zig build test:log` to CI.

CI passes but I get this error on local, haven't yet determined the exact cause but upon experimenting it seems related to stderr:
```
~/lodestar-z feat/logger* > zig build test:log    
test:log
└─ run test log w
[0.000s] [test-module] [info] visible 1
failed command: ./.zig-cache/o/b163e826bfe4adf813cf2ac735f25404/log --cache-dir=./.zig-cache --seed=0x43867136 --listen=-
```
Are you getting it as well?

src/log.zig
```diff
@@ -0,0 +1,155 @@
...
+
+fn writeStderr(bytes: []const u8) void {
+    const io = defaultIo();
+    _ = std.debug.lockStderr(&.{});
+    defer std.debug.unlockStderr();
+    const stdErr = std.Io.File.stderr();
+
+    var stdErrWriteBuffer: [4096]u8 = undefined;
+    var stdErrWriter = stdErr.writer(io, &stdErrWriteBuffer);
+    nosuspend stdErrWriter.interface.writeAll(bytes) catch return;
```
nazarhussain: This will silently discard the whole log entry. We have two options; 

1. Truncate it to a max buffer, so user can at least see some log entries 
2. Log error instead of that entry, so user get aware what happens

src/log.zig
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+fn writeStderr(bytes: []const u8) void {
+    const io = defaultIo();
+    _ = std.debug.lockStderr(&.{});
+    defer std.debug.unlockStderr();
+    const stdErr = std.Io.File.stderr();
+
+    var stdErrWriteBuffer: [4096]u8 = undefined;
```
nazarhussain: We use convention to have camel_case names for the variables.

src/log.zig
```diff
@@ -0,0 +1,155 @@
...
+    if (module) |m| {
+        return std.fmt.bufPrint(buf, "[{d:.3}s] [{s}] {s}[{s}]{s} " ++ fmt ++ "\n", .{ secs, m, level.color(), level.asText(), Colors.reset } ++ args);
+    } else {
+        return std.fmt.bufPrint(buf, "[{d:.3}s] {s}[{s}]{s} " ++ fmt ++ "\n", .{ secs, level.color(), level.asText(), Colors.reset } ++ args);
+    }
+}
+
+fn writeStderr(bytes: []const u8) void {
+    const io = defaultIo();
+    _ = std.debug.lockStderr(&.{});
```
nazarhussain: Don't get this line. The `std.debug.lockStderr` returns [Io](https://ziglang.org/documentation/master/std/#std.Io).[LockedStderr](https://ziglang.org/documentation/master/std/#std.Io.LockedStderr) why would we discard it instead of using? 

The whole function can be written as which seems more concise. 

```zig
fn writeStderr(bytes: []const u8) void {
    // lockStderr acquires the shared stderr mutex (also held by std.log,
    // std.debug.print, and panic output) and returns a writer bound to `buf`.
    // Auto-flushed by unlockStderr — no manual flush, no separate File.Writer.
    var buf: [4096]u8 = undefined;
    const stderr = std.debug.lockStderr(&buf);
    defer std.debug.unlockStderr();
    stderr.file_writer.interface.writeAll(bytes) catch {};
}
```

nazarhussain (changes requested):
There are two divergence form existing logger which is not mentioned in the PR. 

1. Log level in current implementation is  formatted as `info: (colorized, padded trailing colon, no brackets)` in this PR it is as `[info] (bracketed)`
2. Structured context logging is different. In current implementation we handle context by `appends {k=v} context + error`  but in this PR it is formatted as `printf-style comptime fmt + args`

nazarhussain (commented):
Also instead of `log.zig` we should name it `logger.zig`

markolazic01:
thanks @nazarhussain, will make all the requested changes

src/log.zig
markolazic01: I carried this over from Zeam's logger as the reference, didn't realize lockStderr returns a usable writer, will make the requested adaptation 👍

src/log.zig
markolazic01: will go with truncation here, it provides a better ux than just logging an error

GrapeBaBa:
Based on previous discussion in discord, we probably prefer just use the `std.log` for simplify. The remain problem is can we set debug level under release mode. Tigerbeetle seems support it with https://github.com/tigerbeetle/tigerbeetle/blob/main/src/tigerbeetle/main.zig#L61 and https://github.com/tigerbeetle/tigerbeetle/blob/main/src/tigerbeetle/main.zig#L92. Do we need this runtime debug log and just follow the tigerbeetle approach for our logging? @spiral-ladder @wemeetagain @twoeths

markolazic01:
I also thought about just using the std.log but due to a higher number of log levels in current lodestar setup I went this way. Anyways I am willing to make any changes needed.

markolazic01:
Hey @nazarhussain, I've made the following changes:
- log truncation when overflowing the buffer
- update log level formatting: colored, dynamically padded, no brackets
- took your rewrite for stderr handling
- rename: `log` -> `logger` / apply camel case

I wanted to ask if we should defer `{k=v}` to another PR or do you want it here as well?

markolazic01:
Merged to resolve a conflict.

nazarhussain:
@markolazic01 Still see it as draft, is this PR ready for review?

markolazic01:
@nazarhussain you can review it, just please check the changelog in the previous message.
I left it as a draft in case you want to request more changes.

nazarhussain (changes requested):
One remaining divergence from the Lodestar logger that isn't covered in the "Included" or "Deferred" lists: structured context. Lodestar's call convention is `logger.info("Block processed", {slot, root})`, with context appended as `key=value` pairs (and a similar mechanism for errors). This PR is printf-only `logger.info("slot {d}", .{slot})`, which is a different calling convention rather than just a formatting gap.

It's not a Zig limitation — an anonymous-struct context param `.{ .slot = slot }` with an inline for over `@typeInfo` fields would give the same msg `slot=12345` output with zero allocation and compile-time-checked keys.

Totally fine to keep it out of this PR, but could you add it to the deferred list so it's an explicit decision? It affects every future call site, so cheaper to settle the API direction before usage spreads.

src/logger.zig
```diff
@@ -0,0 +1,168 @@
...
+
+    const red = "\x1b[31m";
+    const yellow = "\x1b[33m";
+    const green = "\x1b[32m";
+    const cyan = "\x1b[36m";
+    const white = "\x1b[37m";
+    const gray = "\x1b[90m";
+};
+
+fn formatLine(buf: []u8, module: ?[]const u8, secs: f64, level: LogLevel, comptime fmt: []const u8, args: anytype) ![]const u8 {
```
nazarhussain: The return type is `![]const u8` but I don't see in the code anywhere it can return error. All such cases have error handled.

src/logger.zig
```diff
@@ -0,0 +1,168 @@
...
+    const padding_length: usize = if (used >= padding_between_info) 0 else padding_between_info - used;
+    var padding_storage: [padding_between_info]u8 = undefined;
+    const padding = padding_storage[0..padding_length];
+    @memset(padding, ' ');
+
+    w.print("[{d:.3}s] [{s}] {s}{s}{s}:{s} " ++ fmt ++ "\n", .{ secs, m, padding, level.color(), level.asText(), Colors.reset } ++ args) catch |err| {
+        std.debug.assert(err == error.WriteFailed);
+        @memcpy(buf[w.end..][0..marker.len], marker);
+        return buf[0 .. w.end + marker.len];
+    };
```
nazarhussain, line 69: This code path may result in color bleed. If truncated before the `Color.reset`, then next output will be using earlier color.

nazarhussain (changes requested):

src/logger.zig
markolazic01: removed `!` and adapted the affected lines to the type change

src/logger.zig
markolazic01: Added `Color.reset` as a truncation prefix.

markolazic01:
@nazarhussain I made the necessary changes and added the 'context appended as `key=value` pairs' to the list of deferred features in the description. Do you want me to create issues for the listed features?

Also, a reminder, this error still occurs: https://github.com/ChainSafe/lodestar-z/pull/446#issuecomment-4847143407

wemeetagain:
> Based on previous discussion in discord, we probably prefer just use the `std.log` for simplify. The remain problem is can we set debug level under release mode. Tigerbeetle seems support it with https://github.com/tigerbeetle/tigerbeetle/blob/main/src/tigerbeetle/main.zig#L61 and https://github.com/tigerbeetle/tigerbeetle/blob/main/src/tigerbeetle/main.zig#L92. Do we need this runtime debug log and just follow the tigerbeetle approach for our logging? @spiral-ladder @wemeetagain @twoeths

Yes I'm generally hesitant to enshrine any additional logger over std.log at this stage in our development.
@GrapeBaBa Yes this is exactly what I was thinking. We likely want runtime configuration of log level, like we do in lodestar.

spiral-ladder:
> Yes I'm generally hesitant to enshrine any additional logger over std.log at this stage in our development. @GrapeBaBa Yes this is exactly what I was thinking. We likely want runtime configuration of log level, like we do in lodestar.

+1 as well. I think we can look to add features if we eventually want a more fully featured logger but for now it seems overkill. Let's defer bloating our codebase until we need things

markolazic01:
makes sense, should I close it or keep it as a draft for later?

wemeetagain:
Lets close and reopen later as necessary

Issue #406, opened 2026-06-15
Implement state caches


twoeths:
Implement Block State Cache + Checkpoint State Cache following https://github.com/ChainSafe/lodestar/tree/unstable/packages/beacon-node/src/chain/stateCache

This is part of #405

PR #452, opened 2026-06-28
feat(beacon-node): add block state cache and checkpoint datastore

GrapeBaBa:
## Summary

Part 1 for #406 

- **block_state_cache** — `FIFOBlockStateCache` port: a head-pinned FIFO over recent post-states keyed by state root. Keeps two structures — an insertion-ordered map (so `getSeedState` returns a *stable* reload anchor that doesn't churn when the head moves) and a head-pinned FIFO key order (so eviction takes the oldest tail and never the current head). Per-`add` pruning, with the just-added key protected so `max_states == 1` transiently holds one extra. Read-tracking metrics (size / reads / seconds-since-last-read).
- **cp_datastore** — the checkpoint-state byte datastore the persistent checkpoint cache persists to. Two backends behind a vtable: `File` (the file backend fans batch removal out over `io.async`) and an in-memory `Map` fake for tests. `removeMany` stays in the interface so each backend picks its own strategy (concurrent fan-out for I/O, sequential for the non-thread-safe map).
- **key** — checkpoint ⇄ datastore-key (`0x`-hex filename) encoding.
- **metrics** — block state-cache metric definitions.


GrapeBaBa:
@codex review

GrapeBaBa:
@codex review

src/beacon_node/chain/state_cache/metrics.zig
```diff
@@ -0,0 +1,173 @@
...
+
+    const Count = m.Counter(u64);
+    const Gauge = m.Gauge(u64);
+    const F64Gauge = m.Gauge(f64);
+    const ClonedHistogram = m.Histogram(u64, &.{ 1, 2, 5, 10, 50, 250 });
+};
+
+/// `initializeNoop` default: metrics always emit (no `enabled()` gate), so the cache is safe to use
+/// whether or not `init` is called.
+pub var cache_metrics = m.initializeNoop(BlockStateCacheMetrics);
```
twoeths:
```suggestion
pub var block_cache_metrics = m.initializeNoop(BlockStateCacheMetrics);
```

src/beacon_node/chain/state_cache/block_state_cache.zig
```diff
@@ -0,0 +1,912 @@
...
+    }
+
+    /// Add a state to this cache.
+    /// `is_head` if true, move it to the head of the list. Otherwise add to the 2nd position. In
+    /// importBlock() steps, normally it'll call add() with `is_head` false first, then call
+    /// setHeadState() to set the head.
+    ///
+    /// Returns the canonical resident state — `state` on a fresh insert, the resident on the
+    /// duplicate path (the incoming duplicate is deinit'd). Use the return value afterwards, not
+    /// the passed pointer. Takes `state` only on success; on `hashTreeRoot` failure the caller
```
twoeths: >  Takes `state` only on success;

does this mean ownership is transferred? prefer this term in the documentation instead, this makes it consistent to other places I guess

twoeths (commented):
looks great, I dropped minor comments

src/beacon_node/chain/state_cache/block_state_cache.zig
GrapeBaBa: Reworded in 3e5a3e5 to say ownership of `state` transfers to the cache on success and stays with the caller on `hashTreeRoot` failure — using the ownership vocabulary for consistency. Thanks.

src/beacon_node/chain/state_cache/cp_datastore.zig
```diff
@@ -0,0 +1,623 @@
...
+/// Short or foreign bytes fail the slot reads (treated as unsafe) rather than asserting.
+fn isSafeCheckpointState(state_bytes: []const u8, epoch: Epoch) bool {
+    const state_slot = getStateSlotFromBytes(state_bytes) catch return false;
+    const last_processed = getLastProcessedSlotFromStateBytes(state_bytes) catch return false;
+    // not CRCS or PRCS, skip. Guard the subtract so a malformed slot-0 state cannot underflow.
+    const is_crcs = last_processed == state_slot;
+    const is_prcs = state_slot > 0 and last_processed == state_slot - 1;
+    if (!is_crcs and !is_prcs) return false;
+    // at epoch boundary (subsumes the slot % SLOTS_PER_EPOCH == 0 check).
+    return state_slot == computeStartSlotAtEpoch(epoch);
```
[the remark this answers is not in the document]
GrapeBaBa: Already guarded: `readLatestSafe` computes `computeEpochAtSlot(state_slot)` (division) and returns `false` when it doesn't equal the filename's `epoch` **before** any `computeStartSlotAtEpoch(epoch)` multiply, so a corrupt oversized epoch is skipped rather than overflowing. See cp_datastore.zig — the `Division first` guard.

twoeths (dismissed):

src/beacon_node/chain/state_cache/metrics.zig
GrapeBaBa: Went with your `block_cache_metrics` in 547bdee

GrapeBaBa:
Heads-up on a small functional change pushed mid-review: the lookup map switched from `AutoHashMapUnmanaged` to `AutoArrayHashMapUnmanaged` + `swapRemove` (fe5521d). Rationale: this cache holds its size flat (every add evicts once full), so `HashMap.grow()` — the only point that clears tombstones — never fires, and tombstone-based deletion degrades probes permanently under the add/evict churn (ziglang/zig#17851; the std `rehash()` doc names exactly this pattern). ArrayHashMap's index deletes by backward shift and is immune; map order is unused (both orderings live in the intrusive lists). Sub-noise either way in production — this removes a structural degradation, not a visible slowdown.

wemeetagain (approved):

PR #455, opened 2026-06-30
feat(metrics): metrics bindings

spiral-ladder:
Summary:

- extract `lodestar_` from the names, we can choose to do `lodestar_native_` if we want to that way
- add bindings for scrapeMetrics and init
- add declarations to package.json

### DIFF
--- a/src/state_transition/state_transition.zig
+++ b/src/state_transition/state_transition.zig
@@ -239,7 +239,6 @@ pub fn stateTransition(
     try post_state.commit();
     metrics.state_transition.process_block_commit.observe(time.durationSeconds(time.since(io, timer)));
 
-
     try metrics.state_transition.onPostState(post_cached_state);
 
     // Verify state root
--- a/src/state_transition/state_transition.zig
+++ b/src/state_transition/state_transition.zig
@@ -233,11 +233,9 @@ pub fn stateTransition(
     }
     metrics.state_transition.process_block.observe(time.durationSeconds(time.since(io, timer)));
 
-    //
-    // TODO(bing): commit
-    //  const processBlockCommitTimer = metrics?.processBlockCommitTime.startTimer();
-    //  postState.commit();
-    //  processBlockCommitTimer?.();
+    timer = time.start(io);
+    try post_state.commit();
+    metrics.state_transition.process_block_commit.observe(time.durationSeconds(time.since(io, timer)));
 
     try metrics.state_transition.onPostState(post_cached_state);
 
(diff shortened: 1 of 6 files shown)


src/state_transition/state_transition.zig
```diff
@@ -233,27 +233,22 @@ pub fn stateTransition(
...
+    metrics.state_transition.process_block_commit.observe(time.durationSeconds(time.since(io, timer)));
 
     try metrics.state_transition.onPostState(post_cached_state);
 
     // Verify state root
     if (opts.verify_state_root) {
         timer = time.start(io);
         const post_state_root = try post_state.hashTreeRoot();
-        try metrics.state_transition.state_hash_tree_root.observe(.{ .source = .block_transition }, time.durationSeconds(time.since(io, timer)));
+        try metrics.state_transition.state_hash_tree_root.observe(.{ .source = .state_transition }, time.durationSeconds(time.since(io, timer)));
```
spiral-ladder, line 248: in a previous PR review @twoeths mentioned that it might be more accurate to rename this `block_transition`, but I suppose we should leave it like so so we can be the same as upstream consumer at this point

nazarhussain (commented):
Based on earlier discussion, didn't we identified that metrics will stay in native layer. If we need metrics for behavior in JS which is using some binding code, we can still use the metrics in JS directly for it.

nazarhussain (dismissed):

spiral-ladder:
@nazarhussain based on our private discussion lets not merge this for now

nazarhussain:
Our discussion help to have a metrics setup for future, but for current changes we can proceed with this PR.

GrapeBaBa:
There is a missing piece in zig metrics: no lazy collect hook, so these gauges have to be pull-refreshed

The Zig metrics library doesn't have an equivalent of TS's `addCollect` — i.e. a way to register a lazy collect callback that's evaluated at scrape time. On the TS side, derived/aggregate gauges (cache size, per-epoch counts, read stats, …) register a collect function that the registry invokes on demand when it's scraped, so their values are computed lazily — only when actually needed.

The Zig module has no such hook: a gauge just holds a value, and the exposition serializes whatever it currently holds. So there's no way to compute these lazily at collect time. Should we add this feature in the metrics lib or instead we fall back to a pull model — the cache has to eagerly scan its own state and write the values into the gauges right before write()/scrape?

spiral-ladder:
@grapebaba you're right, but I think it's ok for now, at least in the only place tht we replace with zig (`/packages/state-transition`) we don't have lazy collection happening (checked with `rg \.addCollect` in that package), so I think we can leave it for a followup, unless i'm missing something

GrapeBaBa:
> @GrapeBaBa you're right, but I think it's ok for now, at least in the only place tht we replace with zig (`/packages/state-transition`) we don't have lazy collection happening (checked with `rg \.addCollect` in that package), so I think we can leave it for a followup, unless i'm missing something

Yes, but I need that in forkchoice and state cache, open an issue for that.

PR #456, opened 2026-06-30
fix: state transition binding exports

spiral-ladder:
- use `stateTransition` as instance method instead
- expose `deinitStateTransition`


bindings/napi/stateTransition.zig
```diff
@@ -0,0 +1,6 @@
+const napi_io = @import("./io.zig");
+const st = @import("state_transition");
+
+pub fn deinitStateTransition() void {
+    st.deinitStateTransition(napi_io.get());
+}
```
nazarhussain: Should not we simply have a `deinit` in binding to cleanup automatically? Or we this is required to be done via adhoc user invocation?

nazarhussain (dismissed):

bindings/napi/stateTransition.zig
spiral-ladder: The only cleanup this does is deinit the `ReusedEpochTransitionCache`, which should never be deinit'd anyway during a node's uptime. The usage is just for spec tests. It appeared like not clearing up the cache caused failures when vitest runs tests in parallel. currently i'm using it like this: https://github.com/ChainSafe/lodestar/blob/aafa01409bb4c118c7b8c3278d081f4f80fa90db/packages/beacon-node/test/spec/utils/specTestIterator.ts#L66-L72

see chainsafe/lodestar#9516

spiral-ladder: I think maybe deinitStateTransition is too unclear of a name. Really it's just clearing up the epoch transition cache. We probably should not layer names, let me change it.

wemeetagain (approved):

PR #448, opened 2026-06-26
fix: remove `slashValidator` gating on active status

spiral-ladder:
Causing failures on deployment throwing with:

```
2026-06-26 05:53:29.875	
Error: InCorrectCurrentTargetUnslashedBalance
2026-06-26 05:53:29.875	
    at BeaconChain.verifyBlocksInEpoch (file:///usr/src/lodestar/packages/beacon-node/src/chain/blocks/verifyBlock.ts:176:7)
2026-06-26 05:53:29.875	
    at JobItemQueue.runJob (file:///usr/src/lodestar/packages/beacon-node/src/util/queue/itemQueue.ts:149:22)
2026-06-26 05:53:29.875	
    at NativeBeaconStateView.stateTransition (file:///usr/src/lodestar/packages/state-transition/src/stateView/nativeBeaconStateView.ts:661:54)
```

See: https://github.com/ChainSafe/lodestar/blob/848b01419b92832ea4d0d6b32595641e3cf5cd75/packages/state-transition/src/block/slashValidator.ts#L87-L95

PR #458, opened 2026-07-01
fix: slashing never committing child state tree

spiral-ladder:
We slash the validator in the child view, but never end up committing
the slashing, resulting in inconsistent state view when later
reconstructing the epoch transition cache

### DIFF
--- a/src/state_transition/block/slash_validator.zig
+++ b/src/state_transition/block/slash_validator.zig
@@ -1,3 +1,4 @@
+const std = @import("std");
 const BeaconConfig = @import("config").BeaconConfig;
 const ForkSeq = @import("config").ForkSeq;
 const types = @import("consensus_types");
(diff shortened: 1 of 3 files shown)


src/state_transition/block/slash_validator.zig
```diff
@@ -108,3 +111,67 @@ pub fn slashValidator(
         }
     }
 }
+
+test "slashValidator keeps current target unslashed balance consistent" {
```
spiral-ladder, line 115: this is a bit of much of a setup, honestly not a fan but it helps sanity check our impl

twoeths (commented):
the commit() is not designed to do in the middle of the flow
the whole point of it is to flush changes after processing a block (stateTransition()) or epoch/slot (processSlots())
what's the flow leading to the bug @spiral-ladder

spiral-ladder:
The issue was the slashings not being updated correctly resulting in the same `InCorrectCurrentTargetUnslashedBalance` error (as seen in #448). But you're right that this shouldn't be here. I think the fixes with #454 was the correct fix, because we weren't committing in `postState` after block processing. Closing this

PR #459, opened 2026-07-01
fix: missing deinits in loadOtherState

spiral-ladder:


bindings/napi/BeaconStateView.zig
GrapeBaBa, line 1093: Why this function implemented in binding layer, while I developed state cache, I need exactly this function and implement it in native side too

twoeths: agree that the logic of cached BeaconState needs to be moved to the core state-transition
only binding logic needs to stay here
feel free to refactor that in your PR @GrapeBaBa

twoeths: `validators_view` has borrowed semantic so should not `deinit()` it
looks like the leak happen at this line instead, need to free the result here
same to `balances_view`

spiral-ladder: ya i just needed this for stf so i implemented it the quickest way possible, feel free to refactor

spiral-ladder: it should really just be plain `free`s, fixed in latest commit

wemeetagain (approved):

bindings/napi/BeaconStateView.zig
```diff
@@ -1150,16 +1150,18 @@ pub fn loadOtherState(
...
+            const validators_view = try new_cached_state.state.validators();
+            const validators = validators_view.getAllReadonlyValues(allocator) catch |err| {
                 try js.env().throwError("STATE_ERROR", "Failed to preload validators");
                 return err;
             };
-            var balances_view = try new_cached_state.state.balances();
-            _ = balances_view.getAll(allocator) catch |err| {
+            allocator.free(validators);
+            const balances_view = try new_cached_state.state.balances();
+            const balances = balances_view.getAll(allocator) catch |err| {
```
GrapeBaBa, line 1160: This is useless since currently balances use `chunked-leaf`, there are less nodes than ts side for walk, and it don't populate nodes and not need in current access pattern after bench

bindings/napi/BeaconStateView.zig
```diff
@@ -1150,16 +1150,18 @@ pub fn loadOtherState(
             // This doesn't matter for typescript lodestar because GC clears it anyway,
             // but we're losing some savings here. Consider implementating something like
             // a `prefetchAll` that only does `populateAllNodes` that returns void
-            var validators_view = try new_cached_state.state.validators();
-            _ = validators_view.getAllReadonlyValues(allocator) catch |err| {
+            const validators_view = try new_cached_state.state.validators();
+            const validators = validators_view.getAllReadonlyValues(allocator) catch |err| {
```
GrapeBaBa, line 1154: This is also useless since in the stf right now most used `validatorsPtrSlice` after bench, and try some bench not benefit for `prefetch`

PR #303, opened 2026-04-02
feat(STF): implement Gloas Fork 

guha-rahul:
This pr implements 
- The Gloas fork including EIP-7732
- upgrades to use test vectors `v1.7.0-alpha.4`

PR #462, opened 2026-07-02
feat(STF): implement Gloas Fork  

guha-rahul:
This pr implements 
- The Gloas fork including EIP-7732
- upgrades to use test vectors `v1.7.0-alpha.11`
supersedes https://github.com/ChainSafe/lodestar-z/pull/303


src/constants/root.zig
```diff
@@ -32,21 +32,23 @@ pub const DOMAIN_BLS_TO_EXECUTION_CHANGE = [_]u8{ 10, 0, 0, 0 };
...
+pub const DOMAIN_BUILDER_DEPOSIT = [_]u8{ 14, 0, 0, 0 }; // 0x0E000000
 
 // Application specific domains
 
 pub const DOMAIN_APPLICATION_MASK = [_]u8{ 0, 0, 0, 1 };
 pub const DOMAIN_APPLICATION_BUILDER = [_]u8{ 0, 0, 0, 1 };
 
 // need to be updated when new domain is added
-pub const ALL_DOMAINS = [_][4]u8{ DOMAIN_BEACON_PROPOSER, DOMAIN_BEACON_ATTESTER, DOMAIN_RANDAO, DOMAIN_DEPOSIT, DOMAIN_VOLUNTARY_EXIT, DOMAIN_SELECTION_PROOF, DOMAIN_AGGREGATE_AND_PROOF, DOMAIN_SYNC_COMMITTEE, DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF, DOMAIN_CONTRIBUTION_AND_PROOF, DOMAIN_BLS_TO_EXECUTION_CHANGE, DOMAIN_BEACON_BUILDER, DOMAIN_PTC_ATTESTER, DOMAIN_PROPOSER_PREFERENCES, DOMAIN_APPLICATION_MASK, DOMAIN_APPLICATION_BUILDER };
+pub const ALL_DOMAINS = [_][4]u8{ DOMAIN_BEACON_PROPOSER, DOMAIN_BEACON_ATTESTER, DOMAIN_RANDAO, DOMAIN_DEPOSIT, DOMAIN_VOLUNTARY_EXIT, DOMAIN_SELECTION_PROOF, DOMAIN_AGGREGATE_AND_PROOF, DOMAIN_SYNC_COMMITTEE, DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF, DOMAIN_CONTRIBUTION_AND_PROOF, DOMAIN_BLS_TO_EXECUTION_CHANGE, DOMAIN_BEACON_BUILDER, DOMAIN_PTC_ATTESTER, DOMAIN_PROPOSER_PREFERENCES, DOMAIN_BUILDER_DEPOSIT, DOMAIN_APPLICATION_MASK, DOMAIN_APPLICATION_BUILDER };
```
spiral-ladder:
```suggestion
pub const ALL_DOMAINS = [_][4]u8{ DOMAIN_BEACON_PROPOSER, DOMAIN_BEACON_ATTESTER, DOMAIN_RANDAO, DOMAIN_DEPOSIT, DOMAIN_VOLUNTARY_EXIT, DOMAIN_SELECTION_PROOF, DOMAIN_AGGREGATE_AND_PROOF, DOMAIN_SYNC_COMMITTEE, DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF, DOMAIN_CONTRIBUTION_AND_PROOF, DOMAIN_BLS_TO_EXECUTION_CHANGE, DOMAIN_BEACON_BUILDER, DOMAIN_PTC_ATTESTER, DOMAIN_PROPOSER_PREFERENCES, DOMAIN_BUILDER_DEPOSIT, DOMAIN_APPLICATION_MASK, DOMAIN_APPLICATION_BUILDER, };
```

this got a bit long, let's put a comma at the end and `zig fmt` it

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+
+fn hasBuilderIndexFlag(index: u64) bool {
+    return (index & c.BUILDER_INDEX_FLAG) != 0;
+}
+
+pub fn isBuilderIndex(validator_index: u64) bool {
+    return hasBuilderIndexFlag(validator_index);
+}
+
+pub fn convertBuilderIndexToValidatorIndex(builder_index: u64) u64 {
```
spiral-ladder, line 54: would be nice if we can preserve [doc comments](https://github.com/ChainSafe/lodestar/blob/2e1d834bb9bdfbfc050628171153f6110931d2ca/packages/state-transition/src/util/gloas.ts#L56-L59) from lodestar-ts, same comment for other functions

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+}
+
+pub fn isGasLimitTargetCompatible(parent_gas_limit: u64, gas_limit: u64, target_gas_limit: u64) bool {
+    return gas_limit == getExpectedGasLimit(parent_gas_limit, target_gas_limit);
+}
+
+pub fn getPendingBalanceToWithdrawForBuilder(allocator: Allocator, state: *BeaconState(.gloas), builder_index: u64) !u64 {
+    var pending_balance: u64 = 0;
+
+    var withdrawals = try state.inner.get("builder_pending_withdrawals");
```
spiral-ladder:
```suggestion
    var withdrawals = try state.inner.getReadonly("builder_pending_withdrawals");
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+        if (p.withdrawal.builder_index == builder_index) {
+            pending_balance += p.withdrawal.amount;
+        }
+    }
+
+    return pending_balance;
+}
+
+pub fn canBuilderCoverBid(allocator: Allocator, state: *BeaconState(.gloas), builder_index: u64, bid_amount: u64) !bool {
+    var builders = try state.inner.get("builders");
```
spiral-ladder:
```suggestion
    var builders = try state.inner.getReadonly("builders");
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+    const is_matching = std.mem.eql(u8, &data.beacon_block_root, block_root);
+
+    const prev_block_root = try root_cache.getBlockRootAtSlot(data.slot - 1);
+    const is_current = !std.mem.eql(u8, &data.beacon_block_root, prev_block_root);
+
+    return is_matching and is_current;
+}
+
+pub fn isParentBlockFull(state: *BeaconState(.gloas)) !bool {
+    var bid = try state.inner.get("latest_execution_payload_bid");
```
spiral-ladder:
```suggestion
    var bid = try state.inner.getReadonly("latest_execution_payload_bid");
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+}
+
+pub fn initializePtcWindow(
+    comptime fork: ForkSeq,
+    allocator: Allocator,
+    epoch_cache: *const EpochCache,
+    state: *BeaconState(fork),
+) ![ct.gloas.PtcWindow.length][ct.gloas.PtcWindow.Element.length]ValidatorIndex {
+    const PtcSize = ct.gloas.PtcWindow.Element.length;
+    const PtcWindowLen = ct.gloas.PtcWindow.length;
```
spiral-ladder:
```suggestion
    const ptc_size = ct.gloas.PtcWindow.Element.length;
    const ptc_window_len = ct.gloas.PtcWindow.length;
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+
+    const empty_previous_epoch = [_][PtcSize]ValidatorIndex{
+        [_]ValidatorIndex{0} ** PtcSize,
+    } ** preset.SLOTS_PER_EPOCH;
+
+    const current_epoch = computeEpochAtSlot(try state.slot());
+    var current_and_lookahead: [(1 + preset.MIN_SEED_LOOKAHEAD) * preset.SLOTS_PER_EPOCH][PtcSize]ValidatorIndex = undefined;
```
spiral-ladder:
```suggestion
    var ptcs: [(1 + preset.MIN_SEED_LOOKAHEAD) * preset.SLOTS_PER_EPOCH][PtcSize]ValidatorIndex = undefined;
```

nit: Could we rename this `ptcs` to align with [spec](https://github.com/ethereum/consensus-specs/blob/edfba7e031162fcae7c52198e99df114c1e750f4/tests/core/pyspec/eth_consensus_specs/test/helpers/gloas/state.py#L8)?

src/state_transition/utils/pending_deposits_lookup.zig
```diff
@@ -0,0 +1,92 @@
+const std = @import("std");
```
spiral-ladder, line 13: Let's preserve upstream [doc comment](https://github.com/ChainSafe/lodestar/blob/2e1d834bb9bdfbfc050628171153f6110931d2ca/packages/state-transition/src/util/pendingDepositsLookup.ts#L12-L24)

```
//! Mutable lookup for the pending-deposit sequence used by builder-routing logic.
//!
//! This is to implement the spec's `is_pending_validator(pending_deposits, pubkey)` lazily:
//! deposits are grouped by pubkey without verifying signatures, and BLS verification is
//! deferred until a builder deposit needs to know whether the same pubkey already has a
//! valid pending validator deposit.
//!
//! Call `add()` whenever a deposit is appended to the represented sequence. A cached `true`
//! result short-circuits all subsequent checks for that pubkey; a cached `false` records
//! how many deposits were already verified, so appending a new deposit only verifies the
//! newly-appended tail rather than re-running BLS on previously-invalid entries.

```

As well as any other doc comment found in lodestar-ts

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,311 @@
...
+
+pub fn isGasLimitTargetCompatible(parent_gas_limit: u64, gas_limit: u64, target_gas_limit: u64) bool {
+    return gas_limit == getExpectedGasLimit(parent_gas_limit, target_gas_limit);
+}
+
+pub fn getPendingBalanceToWithdrawForBuilder(allocator: Allocator, state: *BeaconState(.gloas), builder_index: u64) !u64 {
+    var pending_balance: u64 = 0;
+
+    var withdrawals = try state.inner.get("builder_pending_withdrawals");
+    try withdrawals.commit();
```
spiral-ladder:
```suggestion
    var withdrawals = try state.inner.getReadonly("builder_pending_withdrawals");
```
we don't do anything with withdrawals so we can do a readonly get here, and we don't need to commit

src/state_transition/utils/pending_deposits_lookup.zig
```diff
@@ -0,0 +1,92 @@
...
+
+const PendingDepositEntry = struct {
+    deposits: std.ArrayList(PendingDeposit) = .empty,
+    validated_count: usize = 0,
+    has_valid_signature: bool = false,
+
+    fn deinit(self: *PendingDepositEntry, allocator: Allocator) void {
+        self.deposits.deinit(allocator);
+    }
+};
```
spiral-ladder: this is never used outside this file, if we could self contain this, we have the benefit of turning this into a top-level struct + implicitly telling the reader that `PendingDepositEntry` is only relevant here.

Something like:

```zig
pub const PendingDepositsLookup = struct {
    allocator: Allocator,
    deposits_by_pubkey: std.AutoHashMap(BLSPubkey, PendingDeposits),

    const PendingDeposits = struct {
        deposits: std.ArrayList(PendingDeposit) = .empty,
        validated_count: usize = 0,
        has_valid_signature: bool = false,

        fn deinit(self: *PendingDeposits, allocator: Allocator) void {
            self.deposits.deinit(allocator);
        }
    };
```

src/state_transition/utils/pending_deposits_lookup.zig
```diff
@@ -0,0 +1,92 @@
...
+pub const PendingDepositsLookup = struct {
+    allocator: Allocator,
+    deposits_by_pubkey: std.AutoHashMap(BLSPubkey, PendingDepositEntry),
+
+    pub fn buildEmpty(allocator: Allocator) PendingDepositsLookup {
+        return .{
+            .allocator = allocator,
+            .deposits_by_pubkey = std.AutoHashMap(BLSPubkey, PendingDepositEntry).init(allocator),
+        };
+    }
```
spiral-ladder: 2 things here:

because we already own an allocator here + this hashmap is not really doing complicated stuff, we can make this a dumb hashmap by making it unmanaged

Then our init can be simpler too (and we can name it `init` to be zig idiomatic)

```suggestion
    deposits_by_pubkey: std.AutoHashMapUnmanaged(BLSPubkey, PendingDepositEntry),

    pub fn init(allocator: Allocator) PendingDepositsLookup {
        return .{
            .allocator = allocator,
            .deposits_by_pubkey = .empty,
        };
    }
```

Of course this means that the hashmap operations have to pass in `self.allocator` (but i think that makes sense since this struct already owns one

src/state_transition/utils/pending_deposits_lookup.zig
```diff
@@ -0,0 +1,92 @@
...
+
+    pub fn build(comptime fork: ForkSeq, allocator: Allocator, state: *BeaconState(fork)) !PendingDepositsLookup {
```
spiral-ladder: We can be more explicit with naming, eg. `initFromPendingDeposits` or `initFromState` (pick whihcever one u think is more appropriate)

src/state_transition/utils/pending_deposits_lookup.zig
```diff
@@ -0,0 +1,92 @@
...
+        }
+        try result.value_ptr.deposits.append(self.allocator, pending_deposit.*);
+    }
+
+    pub fn hasPendingValidator(
+        self: *PendingDepositsLookup,
+        config: *const BeaconConfig,
+        pubkey: *const BLSPubkey,
+    ) !bool {
+        const entry = self.deposits_by_pubkey.getPtr(pubkey.*) orelse return false;
```
spiral-ladder, line 100: we're missing the validationCache here, see https://github.com/ChainSafe/lodestar/blob/2e1d834bb9bdfbfc050628171153f6110931d2ca/packages/state-transition/src/util/pendingDepositsLookup.ts#L68

src/state_transition/cache/epoch_cache.zig
```diff
@@ -48,6 +48,30 @@ const ForkSeq = @import("config").ForkSeq;
...
+    current: PtcCommittees,
+    next: PtcCommittees,
+};
+
+fn getPtcWindowEpochCacheData(state: *AnyBeaconState) !PtcWindowEpochCacheData {
+    return getPtcWindowEpochCacheDataFromState(try state.tryCastToFork(.gloas));
+}
+
+fn getPtcWindowEpochCacheDataFromState(state: *BeaconState(.gloas)) !PtcWindowEpochCacheData {
+    var ptc_window = try state.inner.get("ptc_window");
```
spiral-ladder:
```suggestion
    var ptc_window = try state.inner.getReadonly("ptc_window");
```

src/state_transition/cache/epoch_cache.zig
```diff
@@ -48,6 +48,30 @@ const ForkSeq = @import("config").ForkSeq;
...
+    var data: PtcWindowEpochCacheData = undefined;
+
+    for (0..preset.SLOTS_PER_EPOCH) |i| {
+        try ptc_window.getValue(undefined, i, &data.previous[i]);
```
spiral-ladder, line 65: out of scope of this PR, but would be nice if we had `getReadonlyByRange` for this usecase

src/state_transition/cache/epoch_cache.zig
```diff
@@ -136,6 +160,15 @@ pub const EpochCache = struct {
 
     epoch: Epoch,
 
+    /// PTC for current epoch, computed eagerly at epoch transition.
+    payload_timeliness_committees: ?[preset.SLOTS_PER_EPOCH][preset.PTC_SIZE]ValidatorIndex,
+
+    /// PTC for previous epoch, required for slot N block validating slot N-1 attestations.
+    previous_payload_timeliness_committees: ?[preset.SLOTS_PER_EPOCH][preset.PTC_SIZE]ValidatorIndex,
```
spiral-ladder: nit: order


```suggestion
    /// PTC for previous epoch, required for slot N block validating slot N-1 attestations.
    previous_payload_timeliness_committees: ?[preset.SLOTS_PER_EPOCH][preset.PTC_SIZE]ValidatorIndex,
    
    /// PTC for current epoch, computed eagerly at epoch transition.
    payload_timeliness_committees: ?[preset.SLOTS_PER_EPOCH][preset.PTC_SIZE]ValidatorIndex,
```

spiral-ladder (commented):
first pass, mostly on `utils/gloas.zig`

spiral-ladder:
One overarching thought that is out of scope of this PR is that I still kinda dislike the whole dumping of everything into `utils/gloas.zig`. We should probably separate the functions meaningfully by name. With that said that's not required here, in fact the current form makes it easier to review, but I'd like to have a followup PR to rearrange stuff eventually

spiral-ladder:
#464 might've broken `nextValue` for this PR - refactored it to make use of some comptime facts to be more explicit, see PR for details

src/state_transition/utils/pending_deposits_lookup.zig
guha-rahul: should i add the doc comments for pre gloas funcs too?

GrapeBaBa:
I suggest to add some OOM/Double free fault injection test using the `test_allocators` and also run process_epoch/process_block bench before and after this change to see if any regression

guha-rahul:
> I suggest to add some OOM/Double free fault injection test using the `test_allocators` and also run process_epoch/process_block bench before and after this change to see if any regression

I did the benchmarking and found near to no regresssions.

src/state_transition/block/process_withdrawals.zig
```diff
@@ -13,13 +13,26 @@ const ExecutionAddress = types.primitive.ExecutionAddress.Type;
...
+const isParentBlockFull = gloas_utils.isParentBlockFull;
 const Node = @import("persistent_merkle_tree").Node;
+const Withdrawal = types.capella.Withdrawal.Type;
+const Epoch = types.primitive.Epoch.Type;
 
 pub const WithdrawalsResult = struct {
     withdrawals: Withdrawals,
     sampled_validators: usize = 0,
     processed_partial_withdrawals_count: usize = 0,
+    // processedBuilderWithdrawalsCount is withdrawals coming from builder payment since EIP-7732
```
spiral-ladder, line 32: nit: make this a doc comment

```suggestion
    /// Tracks number of withdrawals from builder payments since EIP-7732
```

src/state_transition/block/process_withdrawals.zig
```diff
@@ -13,13 +13,26 @@ const ExecutionAddress = types.primitive.ExecutionAddress.Type;
...
+    processed_builder_withdrawals_count: usize = 0,
+    // processedBuildersSweepCount is withdrawals from builder sweep since EIP-7732
```
spiral-ladder, line 34:
```suggestion
    /// Tracks number of withdrawals from builder sweep since EIP-7732
```

src/state_transition/block/process_withdrawals.zig
```diff
@@ -33,45 +46,72 @@ pub fn processWithdrawals(
     expected_withdrawals_result: WithdrawalsResult,
     payload_withdrawals_root: Root,
 ) !void {
-    // processedPartialWithdrawalsCount is withdrawals coming from EL since electra (EIP-7002)
-    const processed_partial_withdrawals_count = expected_withdrawals_result.processed_partial_withdrawals_count;
-    const expected_withdrawals = expected_withdrawals_result.withdrawals.items;
-    const num_withdrawals = expected_withdrawals.len;
+    // [New in EIP-7732] Return early if parent block is empty
```
spiral-ladder, line 49: I think we can do without the [new in ...] comments


```suggestion
```

src/state_transition/block/process_withdrawals.zig
```diff
@@ -33,45 +46,72 @@ pub fn processWithdrawals(
...
         }
     }
 
     // Update the nextWithdrawalIndex
-    if (expected_withdrawals.len > 0) {
-        const latest_withdrawal = expected_withdrawals[expected_withdrawals.len - 1];
-        try state.setNextWithdrawalIndex(latest_withdrawal.index + 1);
+    const latest_withdrawal = if (expected_withdrawals.len > 0) expected_withdrawals[expected_withdrawals.len - 1] else null;
+    if (latest_withdrawal) |lw| {
+        try state.setNextWithdrawalIndex(lw.index + 1);
```
spiral-ladder, line 105: We can probably skip the optional here

```suggestion
    if (expected_withdrawals.len > 0) {
        try state.setNextWithdrawalIndex(expected_withdrawals[expected_withdrawals.len - 1].index + 1);
    }
```

later below `expected_withdrawals.len == preset.MAX_WITHDRAWALS_PER_PAYLOAD` check necessarily means `expected_withdrawals.len > 0`

src/state_transition/block/process_withdrawals.zig
```diff
@@ -82,143 +122,425 @@ pub fn processWithdrawals(
     }
 }
 
-/// Called by the block proposer to find a list of withdrawals to include in the block.
-///
-/// This list is assumed to be bounded by `preset.MAX_WITHDRAWALS_PER_PAYLOAD`.
-///
-/// Caller should deinit `withdrawal_balances` with .deinit() after use.
+// Consumer should deinit WithdrawalsResult with .deinit() after use
```
spiral-ladder, line 125: why was this removed? it's also not a doc comment (triple slash instead of double slash)

src/state_transition/block/process_withdrawals.zig
```diff
@@ -13,13 +13,26 @@ const ExecutionAddress = types.primitive.ExecutionAddress.Type;
...
+const Epoch = types.primitive.Epoch.Type;
 
 pub const WithdrawalsResult = struct {
     withdrawals: Withdrawals,
     sampled_validators: usize = 0,
     processed_partial_withdrawals_count: usize = 0,
+    // processedBuilderWithdrawalsCount is withdrawals coming from builder payment since EIP-7732
+    processed_builder_withdrawals_count: usize = 0,
+    // processedBuildersSweepCount is withdrawals from builder sweep since EIP-7732
+    processed_builders_sweep_count: usize = 0,
```
spiral-ladder, line 35:
```suggestion
    processed_builders_sweep_count: u64 = 0,
```

think we can just use u64 for this and avoid the intCast later

src/state_transition/block/process_withdrawals.zig
```diff
@@ -82,143 +122,425 @@ pub fn processWithdrawals(
...
+
+    var builder_pending_withdrawals = try state.inner.get("builder_pending_withdrawals");
+    try builder_pending_withdrawals.commit();
+    const builder_pending_withdrawals_len = try builder_pending_withdrawals.length();
+    var bw_it = builder_pending_withdrawals.iteratorReadonly(0);
+
+    var processed_count: usize = 0;
+    for (0..builder_pending_withdrawals_len) |_| {
+        // Check combined length against limit
+        const all_withdrawals = prior_withdrawals_len + builder_withdrawals.items.len;
```
spiral-ladder, line 423: naming is confusing. `all_withdrawals` imply this contains all `Withdrawal` objects. prefer a different name, eg. `all_withdrawals_len`

We might even want to just bound the for loop between min of builder_pending_withdrawals_len and  `MAX_WITHDRAWALS_PER_PAYLOAD - 1`. then we don't need this check

src/state_transition/block/process_withdrawals.zig
```diff
@@ -82,143 +122,425 @@ pub fn processWithdrawals(
...
 
-        // capella full withdrawal
-        if (withdrawable_epoch <= epoch) {
+        if (isFullyWithdrawableValidator(fork, withdrawal_credentials, balance, withdrawable_epoch, epoch)) {
             var execution_address: ExecutionAddress = undefined;
             @memcpy(&execution_address, withdrawal_credentials[12..]);
-            withdrawals_result.withdrawals.appendAssumeCapacity(.{
-                .index = withdrawal_index,
+            try sweep_withdrawals.append(allocator, .{
+                .index = withdrawal_index.*,
```
spiral-ladder, line 373: I think we can still bound this by `preset.MAX_WITHDRAWALS_PER_PAYLOAD - num_prior_withdrawals`?

src/state_transition/block/process_withdrawals.zig
```diff
@@ -82,143 +122,425 @@ pub fn processWithdrawals(
...
+            &builder_balance_after_withdrawals,
+        );
+        defer result.withdrawals.deinit(allocator);
+        for (result.withdrawals.items) |w| {
+            try withdrawals_result.withdrawals.append(allocator, w);
+        }
+        processed_builders_sweep_count = result.processed_count;
+    }
+
+    // get_validators_sweep_withdrawals
```
spiral-ladder, line 203:
```suggestion
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,417 @@
...
+/// Check if a validator index represents a builder (has the builder flag set).
+/// Spec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-is_builder_index
+pub fn isBuilderIndex(validator_index: u64) bool {
+    return hasBuilderIndexFlag(validator_index);
+}
+
+/// Convert a builder index to a flagged validator index for use in Withdrawal containers.
+/// Spec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-convert_builder_index_to_validator_index
+pub fn convertBuilderIndexToValidatorIndex(builder_index: u64) u64 {
+    return if (hasBuilderIndexFlag(builder_index)) builder_index else builder_index | c.BUILDER_INDEX_FLAG;
```
spiral-ladder, line 55: this is idempotent so we don't need the if check


```suggestion
    return builder_index | c.BUILDER_INDEX_FLAG;
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,417 @@
...
+}
+
+/// Convert a flagged validator index back to a builder index.
+/// Spec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-convert_validator_index_to_builder_index
+pub fn convertValidatorIndexToBuilderIndex(validator_index: u64) u64 {
+    return if (hasBuilderIndexFlag(validator_index)) validator_index & ~c.BUILDER_INDEX_FLAG else validator_index;
```
spiral-ladder, line 61:
```suggestion
    return validator_index & ~c.BUILDER_INDEX_FLAG;
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,417 @@
...
+    return quorum / c.BUILDER_PAYMENT_THRESHOLD_DENOMINATOR;
+}
+
+fn hasBuilderIndexFlag(index: u64) bool {
+    return (index & c.BUILDER_INDEX_FLAG) != 0;
+}
+
+/// Check if a validator index represents a builder (has the builder flag set).
+/// Spec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-is_builder_index
+pub fn isBuilderIndex(validator_index: u64) bool {
```
spiral-ladder, line 48: this is just wrapping `hasBuilderIndexFlag`? We should move that here instead to align with spec

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,417 @@
...
+    deposit_signature: BLSSignature,
+) bool {
+    const deposit_message = ct.phase0.DepositMessage.Type{
+        .pubkey = pubkey.*,
+        .withdrawal_credentials = withdrawal_credentials.*,
+        .amount = amount,
+    };
+
+    var domain: ct.primitive.Domain.Type = undefined;
+    computeDomain(c.DOMAIN_BUILDER_DEPOSIT, config.chain.GENESIS_FORK_VERSION, c.ZERO_HASH, &domain) catch return false;
```
spiral-ladder, line 227: a bit long, add a trailing comma and fmt
```suggestion
    computeDomain(c.DOMAIN_BUILDER_DEPOSIT, config.chain.GENESIS_FORK_VERSION, c.ZERO_HASH, &domain,) catch return false;
```

src/state_transition/utils/gloas.zig
```diff
@@ -0,0 +1,417 @@
...
+    const block_root = try getBlockRootAtSlot(.gloas, state, data.slot);
+    const is_matching = std.mem.eql(u8, &data.beacon_block_root, block_root);
+
+    const prev_block_root = try getBlockRootAtSlot(.gloas, state, data.slot - 1);
+    const is_current = !std.mem.eql(u8, &data.beacon_block_root, prev_block_root);
+
+    return is_matching and is_current;
+}
+
+pub fn isAttestationSameSlotRootCache(root_cache: *RootCache(.gloas), data: *const ct.phase0.AttestationData.Type) !bool {
```
spiral-ladder, line 252: this should maybe even be in `root_cache.zig`

```suggestion
/// Use cached block roots to avoid repeated state root lookups while matching the spec's is_attestation_same_slot behavior.
pub fn isAttestationSameSlotRootCache(root_cache: *RootCache(.gloas), data: *const ct.phase0.AttestationData.Type) !bool {
```

spiral-ladder (changes requested):
reviewed `process_withdrawals.zig` and `utils/gloas.zig`

spiral-ladder:
We should also keep this aligned with the spec test version in https://github.com/ChainSafe/lodestar/blob/62b30e7337e3bbc1098037b0e856ef6b2f106466/spec-tests-version.json

edit: but this will need #99

src/state_transition/block/process_withdrawals.zig
spiral-ladder: lets' ignore this and fix it later, i think keeping it same as upstream is ok

src/state_transition/block/process_withdrawals.zig
spiral-ladder: resolving, https://github.com/ChainSafe/lodestar-z/pull/462#discussion_r3839341593

Issue #457, opened 2026-07-01
refactor(clock): consolidate SlotClock/EventClock into a single user-facing Clock


nazarhussain:
## Context

Follow-up on design feedback from PR #354 (*feat(clock): add clock module for slot/epoch timing*, merged in #354). The PR merged before this architectural concern was resolved. My review raising it was dismissed, and a [post-merge comment](https://github.com/ChainSafe/lodestar-z/pull/354#issuecomment-3282555888) restating it received no follow-up. Opening this to track it.

## Problem

The clock module currently exposes a three-layer design with **two** public stateful types:

- `slot_math` — pure arithmetic (layer 0)
- `SlotClock` — stateful slot clock (layer 1)
- `EventClock` — async event loop with listeners/waiters (layer 2)

`root.zig` exports both `SlotClock` and `EventClock` as public API:

```zig
pub const SlotClock = @import("SlotClock.zig");
pub const EventClock = @import("EventClock.zig");
```

The separation between `SlotClock` and `EventClock` is hard to reason about for consumers. There is no single user-facing `Clock` object like we conceptually had before; callers must understand which of the two layers to reach for, and `SlotClock` is not actually consumed directly by external users.

## Proposal

1. Provide **one** public, user-facing `Clock` (essentially today's `EventClock`).
2. Fold the pure-math portions of `SlotClock` into `slot_math`.
3. Internalize `SlotClock` (make it a private implementation detail rather than a public export), since it isn't consumed directly.

This preserves the internal layering while collapsing the public surface to a single, easy-to-reason-about type.

## References

- PR #354 — https://github.com/ChainSafe/lodestar-z/pull/354
- Original (dismissed) design review and the post-merge restatement of the concern on that PR.

GrapeBaBa:
Thanks for raising this.

The reason both `SlotClock` and `EventClock` are exposed is separation of concerns:

- `SlotClock` is the query layer — current slot/epoch, `secFromSlot`/`msFromSlot`, tolerances, etc. It's pure and side-effect-free, and it's what the vast majority of callers should depend on.
- `EventClock` is just the ticker (the name could probably be better). It runs the slot/epoch event loop and handles catch-up internally within that loop. You should only hold it to subscribe to slot/epoch events or to `waitForSlot`.

So the split itself is intentional and I still think it's right: most places use `SlotClock`, and only the event-driven ones touch `EventClock`.

[...]

PR #463, opened 2026-07-03
refactor(clock): single public Clock; internalize SlotClock

GrapeBaBa:
Follow-up to #354 (which added the clock module) and the design feedback in #457.

Consolidates the clock module to a single public `Clock`: `EventClock` is renamed to `Clock`, `SlotClock` is dissolved into it, and the pure slot reads move into `slot_math`.


wemeetagain (dismissed):
I think this is much more readable, lgtm

src/clock/Clock.zig
```diff
@@ -357,42 +375,36 @@ pub fn waitForSlot(self: *EventClock, target: Slot) Error!WaitForSlotResult {
...
-/// No-op if already caught up or pre-genesis (currentSlot() returns null).
-fn catchUp(self: *EventClock) void {
-    if (self.clock.currentSlot()) |wall_slot| {
+/// Advance to wall-clock time, emitting any pending slot/epoch events, and
+/// return the reading used. Emits nothing if already caught up or pre-genesis.
+///
+/// Accessors must derive from this reading, not a fresh clock read: a slow
+/// callback can cross a slot boundary mid-dispatch, so a second read could
+/// name a slot the just-emitted events haven't reached.
+fn catchUp(self: *Clock) u64 {
```
nazarhussain: Most of the usage of this function result in recomputing the `slot` from the given `now_ms`. Why not return the both from this function to avoid  recomputation on multiple places?

nazarhussain: Possible stack overflow when an `onSlot` callback calls `clock.currentSlot()` during a backlog.

`currentSlot()` calls `catchUp()`, and `catchUp()` is also the thing that runs listener callbacks. So while my callback is being called by `catchUp` for slot N, my call to currentSlot() starts a second catchUp inside it....

With a normal backlog of 1 slot this is harmless. But after the process is paused for a while (laptop sleep, SIGSTOP, VM pause), the backlog can be thousands of slots.

Suggested fix, never dispatch from a nested call. Add a dispatching flag: if currentSlot() is called from inside a callback, catchUp just records the newest wall slot in pending_target and returns.

```zig
fn catchUp(self: *Clock) u64 {
    const now_ms = time.nowMs(self.io);
    const target = slot_math.slotAtMs(self.config, now_ms) orelse return now_ms;

    if (self.dispatching) {
        // Called from inside a callback: just note the target, don't recurse.
        self.pending_target = @max(self.pending_target orelse target, target);
        return now_ms;
    }

    self.dispatching = true;
    defer self.dispatching = false;
    self.pending_target = target;
    defer self.pending_target = null;
    while (!self.stopped) {
        const t = self.pending_target orelse break;
        self.pending_target = null;
        self.advanceAndDispatch(t);
    }
    return now_ms;
}
```

Nice side effect, listeners now always see events in order (1, 2, 3 instead of 2, 3, 1 when a callback triggers a mid-backlog catch-up), so the "out of order but exactly once" caveat in the docs and tests goes away.

nazarhussain (changes requested):

src/clock/slot_math.zig
```diff
@@ -1,9 +1,12 @@
-//! Layer 0 – Pure slot/epoch arithmetic.
+//! Pure slot/epoch arithmetic and policy helpers.
```
spiral-ladder: what are policy helpers?

src/clock/slot_math.zig
```diff
@@ -313,3 +399,105 @@ test "fork-aware: two transitions" {
...
+}
+
+test "gossip disparity: just after slot boundary" {
+    try testing.expect(
+        isCurrentSlotGivenGossipDisparity(test_cfg, 0, 112_300),
+    );
+}
+
+test "gossip disparity: exact threshold (500ms) applies inclusively" {
+    // 111_500 is 500 ms before slot 1's 112_000 start — inclusive edge; 111_499 is 1 ms past.
```
spiral-ladder: imo we need more similar comments or var names similar to this for all the other tests so it's less like magic numbers, otherwise the next reader (me) might have a hard time understanding why these numbers were used

src/clock/slot_math.zig
```diff
@@ -105,6 +108,89 @@ pub fn msUntilNextSlot(config: ClockConfig, now_ms: u64) u64 {
...
+/// disagree — a fresh read could sit past the boundary being compared.
+///
+/// Per phase0/p2p-interface.md, gossip validation rejects future messages with
+/// strict `<` (`current_time + MAXIMUM_GOSSIP_CLOCK_DISPARITY < message_time`),
+/// so the boundary case (exactly equal) is accepted — hence `<=` here.
+///
+/// Assumes the disparity window reaches at most the adjacent slot — true
+/// for every real config (500 ms disparity vs seconds-long slots). A
+/// config where disparity approaches or exceeds the slot duration is not
+/// supported.
```
spiral-ladder: this can really be a lot more concise with the same meaning


```suggestion
/// Returns the slot the network may be advancing to, accounting for gossip
/// clock disparity, or null pre-genesis when no slot is current yet.
///
/// Per phase0/p2p-interface.md, gossip validation rejects future messages with
/// strict `<`, hence `<=` here.
///
/// Assumes the disparity window reaches at most the adjacent slot — true
/// for every real config (500 ms disparity vs seconds-long slots).
```

src/clock/Clock.zig
```diff
@@ -402,19 +414,75 @@ fn dispatchWaiters(self: *EventClock, current_slot: ?Slot) void {
...
-fn advanceAndDispatch(self: *EventClock, target: Slot) void {
-    var iter = self.clock.advanceTo(target);
+const Event = union(enum) {
+    slot: Slot,
+    epoch: Epoch,
+};
+
+// Holds only what advancing needs — config and the slot cursor — so the
+// iterator cannot dispatch (no listeners, waiters, or io in reach).
+const AdvanceIterator = struct {
```
spiral-ladder: iterators are not meant to dispatch anyway, so i think this comment is kind of redundant

spiral-ladder (commented):
i'm concerned with the amount of doc comment and test noise that is generated, a lot of it is verbose and using terms that are not really obvious at first glance, we should strive to review the generated doc comments as much as possible to optimize for human readability

i'm being a bit of a stickler about this because it's easy to generate tons of comments and for others to give it an approval since technically it doesn't affect code quality, but imo after a few more of such PRs it makes it hard to maintain and understand a codebase

GrapeBaBa:
> i'm concerned with the amount of doc comment and test noise that is generated, a lot of it is verbose and using terms that are not really obvious at first glance, we should strive to review the generated doc comments as much as possible to optimize for human readability
> 
> i'm being a bit of a stickler about this because it's easy to generate tons of comments and for others to give it an approval since technically it doesn't affect code quality, but imo after a few more of such PRs it makes it hard to maintain and understand a codebase

Yes I agree that, actually I had a specific skill and step to handle code comments in each loop. But probably still need edit by hand

src/clock/slot_math.zig
GrapeBaBa:
```
/// Per phase0/p2p-interface.md, gossip validation rejects future messages with
/// strict `<` (`current_time + MAXIMUM_GOSSIP_CLOCK_DISPARITY < message_time`),
/// so the boundary case (exactly equal) is accepted — hence `<=` here.
```
This is generated by @lodekeeper 😂 in ts side

lodekeeper: Guilty 😄 — that's ported from the TS `slotWithGossipDisparity` doc comment. @spiral-ladder's trim above is the right call: I over-explained the `<=`/boundary and adjacent-slot caveat on the TS side, and the port shouldn't inherit that verbosity. Please trim it freely.

lodekeeper:
Owning this one — a lot of the verbose doc comments and test scaffolding here were ported from my TS side, and terms like "policy helpers" or the bare-number tests read okay with TS context but are opaque standing alone. Please don't carry the TS verbosity into the Zig port: trim aggressively, rename unclear terms, and add named constants where tests use magic numbers. I'll keep the TS-side comments leaner too. Happy to propose concise rewrites for any specific comment if that helps.

spiral-ladder:
@lodekeeper <3

> Owning this one — a lot of the verbose doc comments and test scaffolding here were ported from my TS side, and terms like "policy helpers" or the bare-number tests read okay with TS context but are opaque standing alone. Please don't carry the TS verbosity into the Zig port: trim aggressively, rename unclear terms, and add named constants where tests use magic numbers. I'll keep the TS-side comments leaner too. Happy to propose concise rewrites for any specific comment if that helps.

src/clock/slot_math.zig
nflaig: @lodekeeper please open a PR to trim this on lodestar ts side

lodekeeper: Done — ChainSafe/lodestar#9631 trims both TS-side comments (`currentSlotWithGossipDisparity` + `isCurrentSlotGivenGossipDisparity`) down to the spec pointer.

src/clock/Clock.zig
GrapeBaBa: Cool, nice refactoring suggestion fixed both stack overflow and make the events in order

nazarhussain (dismissed):

GrapeBaBa:
@nazarhussain @spiral-ladder eliminate the magic number in the tests and remove some useless tests

nazarhussain (approved):

PR #464, opened 2026-07-03
feat: split nextValue fn

spiral-ladder:
`nextValue` always required an `allocator` as a param but it was not used if the `ST.Element` was a fixed type element. This goes against zig's ideal of being as explicit as possible.

Since we already know if the `Element` contained with the composite list is a fixed type or not, so we use that fact to delegate the fn to non public functions `nextValueFixed` or `nextValueAlloc` depending on the inner type.

Also gets rid of the ugly `undefined` usage in consumers.


spiral-ladder:
We might be able to clean the `ssz` module up since this pattern is repeated quite a bit throughout

GrapeBaBa (approved):
Good refactor

PR #469, opened 2026-07-07
refactor(bindings): use zapi js.io() instead of local io module

nazarhussain:
## Motivation

zapi v3.0.0 added a lifecycle-managed `js.io()` to the JS DSL (ChainSafe/zapi#26): `js.exportModule` retains a shared `std.Io.Threaded` before the module's `init` hook and releases it after the `cleanup` hook, refcounted across N-API environments. This is the same lifecycle our hand-rolled `bindings/napi/io.zig` implemented, so the local module can be deleted.

## Summary

- bump zapi v2.2.0 → v3.0.0
- delete `bindings/napi/io.zig`; replace all `napi_io.get()` call sites with `js.io()` (`blst`, `pubkeys`, `metrics`, `BeaconStateView`, `root`)
- drop manual io init/deinit from the `root.zig` lifecycle hooks — ordering is preserved upstream (`retain()` runs before `init`, `release()` after `cleanup`), so `js.io()` is valid during CPU detection at init and thread pool teardown at cleanup

### zapi v3 breaking change

v3.0.0 rejects non-DSL `pub fn`s in `exportModule` (ChainSafe/zapi#38) instead of silently skipping them. Internal decls that the exporter walked are now hidden:

- `pool.State`, `config.State`, `pubkeys.State`, `config.chainConfigFromObject`: de-pubbed (cross-file access still works through the pub `state` vars; the exporter skips pub vars)
- `pool.PoolRc`: de-pubbed; `BeaconStateView.pool_rc` now uses `@TypeOf(pool.state.pool_rc)`
- `blst.initThreadPool`/`deinitThreadPool`: moved behind a pub `lifecycle` var so they stay native-only and are no longer silently exported to JS (`deinitThreadPool`, zero-arg, previously *was* reachable from JS; nothing on the TS side used it)

## Validation

- `zig build build-lib:bindings` (Debug and ReleaseSafe)
- `pnpm test`: all 233 tests pass, including `teardown.test.ts` (clean process exit through the new env-cleanup → io release path, no panic)
- `zig fmt --check` on changed files


bindings/napi/blst.zig
```diff
@@ -39,25 +38,31 @@ const BATCH_VERIFY_SIZE = 32;
 /// Initialized lazily on first use, torn down via `deinitThreadPool`.
 var thread_pool: ?*ThreadPool = null;
 
-pub fn initThreadPool(n_workers: u16) !void {
```
nazarhussain: We could have kept the function as it is by using following signature. 

```zig
pub fn initThreadPool(n_workers: js.Number) !void
```

But then it would have been expose to JS surface, which we don't want, as it's an internal life cycle function only used at root with the number of cpu count.

bindings/napi/blst.zig
```diff
@@ -39,25 +38,31 @@ const BATCH_VERIFY_SIZE = 32;
...
-    if (thread_pool != null) return error.PoolExists;
-    thread_pool = try ThreadPool.init(std.heap.page_allocator, napi_io.get(), .{ .n_workers = n_workers });
-}
+/// Native-only thread pool lifecycle, reached from `root.zig` through the
+/// pub `lifecycle` var so it is not part of the JS module surface.
+const Lifecycle = struct {
```
nazarhussain: This pattern is safety hatch for above mentioned problem. Now we can access the lifecycle functions in zig but not in the js surface.

spiral-ladder: this seems reasonable and cleaner than https://github.com/ChainSafe/lodestar-z/pull/483 ,  but thinking about future modules we expose, might get confusing with mixed use in the same file, but this might be overthinking, we can reconsider again when/if we encounter this issue because the blast radius here is not big

nit: is `LifeCycle` a good name though, seems like `blst.lifecycle...` is not immediately obvious

spiral-ladder (dismissed):
LGTM with comment

nazarhussain:
@spiral-ladder 

> nit: is LifeCycle a good name though

Suggest any other name, which you feel much aligned. 

> might get confusing with mixed use in the same file,

Ideally we should not have any interface in a binding file that we actually want to hide from the JS runtime.

spiral-ladder:
@nazarhussain i don't have a better name yet :) hence approving now, we can come up with one if someone other than me finds it confusnig

wemeetagain (commented):
Well another option is to use the standard of the other bindings modules: `State` / `state`.

eg: `blst.state.deinit`

spiral-ladder (dismissed):

nazarhussain:
> Well another option is to use the standard of the other bindings modules: `State` / `state`.
> 
> eg: `blst.state.deinit`

@wemeetagain Yes I agree to it and opened a separate PR to address this change https://github.com/ChainSafe/lodestar-z/pull/516 to not mix with the `js.io()` change.

spiral-ladder (approved):

GrapeBaBa:
I have a question about why zapi should expose an IO implementation, it looks like it create a separate thread pool and not use the application's thread pool, I thought it is better to let application managing the unified thread pool and concurrency strategy, more control for tuning

nazarhussain:
@GrapeBaBa 

Short answer to your question, the pool isn't actually created up front, nothing forces you to use it, and in a Node addon there is no "application thread pool" to unify with.


Key points are: 

1. `js.io()` is an opt-in, not a strict policy of the Zapi DSL. No DSL internals consume `js.io()`, the `exportModule` only manage it's life cycle with it. So the only caller of `js.io()` is the user code (in examples here or actual addons). An addon that wants the full control can simply construct it's own `std.Io` an use it. The `hell_world` example does this with `Io.Threaded.global_single_threaded.io().`. 

2. "It creates a separate thread pool" is latent, not eager.  Threaded.init spawns zero threads — it just fills in a struct. 

3. There is no application `main` to thread an Io. A Node add on has no Zi gentry point that owns the process; every entry is an N-API callback invoked by Node, and "the application" is the JS program which has no ZIg thread pool to unify with. So the realistic choice is "every addon author hand-rolls a global" vs. "the library provides a correctly managed default" — same rationale as js.allocator().

4. The part that genuinely belongs in the library is lifecycle, not concurrency strategy. The shared Io is refcounted against active N-API environments (`SharedResource` in `src/js/lifecycle.zig`) — initialized on the first env, torn down after the last — so worker exit neither leaks the pool nor leaves callbacks holding a freed Io. A hand-rolled application global has to reimplement exactly that refcounting to be worker-safe. That subtlety is the actual value of `js.io()`.

--- 

One point that shared instance is hardcoded to `hardcoded to std.Io.Threaded.init(page_allocator, .{}) ` is correct. A reasonable enhancement would be an `exportModule` option to configure the shared instance's `InitOptions`, or to let the application supply its own `Io` that the DSL merely lifecycles.

GrapeBaBa:
> A reasonable enhancement would be an exportModule option to configure the shared instance's InitOptions, or to let the application supply its own Io that the DSL merely lifecycles.

Yes, this is my thought.

nazarhussain:
@GrapeBaBa I feel we should do that enhancement as separate PR in the future. What do you think?

GrapeBaBa:
> @GrapeBaBa I feel we should do that enhancement as separate PR in the future. What do you think?

Sure

PR #471, opened 2026-07-07
fix(bindings): stop leaking the state-preload scratch slices

GrapeBaBa:
`preloadValidatorsAndBalances` discarded the slices returned by `getAllReadonlyValues`/`getAll` without freeing them — every call leaked **~240 MB** of decoded validator values plus ~16 MB of balances from the global allocator (the TODO's "GC clears it anyway" reasoning does not apply on the Zig side).

- Free the scratch immediately after the warm-up.
- Warm the validator view via `getAllReadonly` (pointers) instead of decoding every validator into values — the cache warming is identical (both routes go through `getReadonly`) and the scratch drops from ~240 MB to ~16 MB.

Follow-up: once a void `prefetchAll` lands on the list views (#466), this site can drop the scratch allocation entirely.


GrapeBaBa:
Superseded by #459, which already frees both preload scratch slices. Closing.

PR #475, opened 2026-07-09
feat: partial validator monitor

spiral-ladder:
Implements the missing validator monitor metrics from lodestar-ts.

Note that there are some metrics intentionally omitted because they probably require even bigger changes on the zig side, this is just the minimal change required to track validator perf. 

The following metrics are excluded:

- `validator_monitor_prev_epoch_on_chain_attester_hit_total`
- `validator_monitor_prev_epoch_on_chain_attester_miss_total`
- `validator_monitor_prev_epoch_on_chain_attester_correct_head_total`
- `validator_monitor_prev_epoch_on_chain_attester_incorrect_head_total`
- `validator_monitor_prev_epoch_on_chain_inclusion_distance`

These will be updated on the TS side to run on [`onceEveryEndOfEpoch`](https://github.com/ChainSafe/lodestar/blob/33dbc1859acac60a4f6878259c4e5c830c4ff506/packages/beacon-node/src/chain/validatorMonitor.ts#L105) rather than inside `registerValidatorStatuses`, which is now zig native


spiral-ladder:
> These will be updated on the TS side to run on [onceEveryEndOfEpoch](https://github.com/ChainSafe/lodestar/blob/33dbc1859acac60a4f6878259c4e5c830c4ff506/packages/beacon-node/src/chain/validatorMonitor.ts#L105) rather than inside registerValidatorStatuses, which is now zig native

This is now done in this commit https://github.com/ChainSafe/lodestar/pull/9632/changes/19747b9db42d5ec429ddc82d8091990c3facc966

bindings/napi/metrics.zig
```diff
@@ -28,7 +49,8 @@ pub fn scrapeMetrics() !js.String {
 }
 
 pub fn deinit() void {
+    validator_monitor.deinit();
```
wemeetagain: needs to be moved down

wemeetagain:
more conflicts

src/state_transition/ValidatorMonitor.zig
```diff
@@ -0,0 +1,118 @@
...
+const types = @import("consensus_types");
+const metrics = @import("metrics.zig");
+const attester_status = @import("utils/attester_status.zig");
+const hasMarkers = attester_status.hasMarkers;
+
+const Allocator = std.mem.Allocator;
+const Epoch = types.primitive.Epoch.Type;
+const ValidatorIndex = types.primitive.ValidatorIndex.Type;
+
+pub const ValidatorMonitor = @This();
```
wemeetagain: I suppose this doesn't need to be marked pub?

```zig
const ValidatorMonitor = @import("./ValidatorMonitor.zig");
ValidatorMonitor == ValidatorMonitor.ValidatorMonitor
```

spiral-ladder: fixed in [5b38ad6](https://github.com/ChainSafe/lodestar-z/pull/475/commits/5b38ad64ef215f172f8216153ea0ad27e2448727)

bindings/napi/validator_monitor.zig
```diff
@@ -0,0 +1,31 @@
...
+var gpa: std.heap.DebugAllocator(.{}) = .init;
+const allocator = if (builtin.mode == .Debug)
+    gpa.allocator()
+else
+    std.heap.c_allocator;
+
+/// Fed by `BeaconStateView.processSlots`/`stateTransition` on every epoch
+/// transition. Only records metrics for validators registered via
+/// `metrics.registerLocalValidator()`.
+var monitor = state_transition.ValidatorMonitor.init(allocator);
```
GrapeBaBa, line 21: Is this possible accessed by multi threads?

wemeetagain: in practice it will only be accessed from the main thread / control thread -- as helper functions that toggle validator monitor behavior which feeds the metrics, which are themselves also only accessed from the main thread / control thread..

spiral-ladder: historical regen also tracks metrics but under a different namespace

bindings/napi/BeaconStateView.zig
```diff
@@ -1338,7 +1346,14 @@ pub fn stateTransition(
...
     defer signed_block.deinit(allocator);
 
-    const post_state = try st.stateTransition(allocator, js.io(), cached_state, signed_block, opts);
+    const post_state = try st.stateTransition(
+        allocator,
+        js.io(),
+        cached_state,
+        signed_block,
+        opts,
+        validator_monitor.get(),
```
GrapeBaBa, line 1355: Lodestar seems not pass validator_monitor for regen https://github.com/ChainSafe/lodestar/blob/e22a22a11d2bc01ca481a1f4d74164ddbbff0264/packages/beacon-node/src/chain/archiveStore/historicalState/getHistoricalState.ts#L65

src/state_transition/ValidatorMonitor.zig
```diff
@@ -0,0 +1,118 @@
...
+
+const Allocator = std.mem.Allocator;
+const Epoch = types.primitive.Epoch.Type;
+const ValidatorIndex = types.primitive.ValidatorIndex.Type;
+
+const ValidatorMonitor = @This();
+
+allocator: Allocator,
+/// Unordered list of validators that require additional monitoring.
+validators: std.AutoArrayHashMapUnmanaged(ValidatorIndex, void),
```
GrapeBaBa: For new code, may prefer not use deprecated unmanaged version?

src/state_transition/ValidatorMonitor.zig
```diff
@@ -0,0 +1,118 @@
...
+    var total_balance: u64 = 0;
+
+    for (self.validators.keys()) |index| {
+        // The monitored validator may not be in the state yet.
+        if (index >= flags.len) continue;
+
+        const flag = flags[index];
+
+        if (hasMarkers(flag, attester_status.FLAG_PREV_SOURCE_ATTESTER)) {
+            vm.prev_epoch_on_chain_source_attester_hit.incr();
```
GrapeBaBa, line 91: Maybe a minor optimization not necessary, sum in loop first, only call `incr` once after loop.

wemeetagain (approved):
lgtm

Looking at this, we might rethink the singleton metrics approach and instead pass the instance thru as an arg. (I was a little surprised looking at ValidatorMonitor.zig and it took a sec to see that it just uses the singleton)

bindings/napi/BeaconStateView.zig
GrapeBaBa: @wemeetagain @spiral-ladder Is this intentional change, to use this monitor for regen in lodestar-z? The code I looked at in lodestar it seems not do that

wemeetagain: It looked like there is a short-circuit which prevents old epoch processing from contaminating the monitor, didn't seem like a big deal

GrapeBaBa: ok, I didn't see that short-circuit, I just noticed the current validator monitor can't access concurrent, lodestar pass null for regen seems more safe

Issue #467, opened 2026-07-07
Add readonly range access for SSZ list views


guha-rahul:
out of scope of this PR, but would be nice if we had `getReadonlyByRange` for this usecase

_Originally posted by @spiral-ladder in https://github.com/ChainSafe/lodestar-z/pull/462#discussion_r3529291103_

andreolf:
Picking this up. Draft PR: #586

PR #476, opened 2026-07-09
feat: add readonly range reads for SSZ composite vectors

guha-rahul:
This pr implements  
- adds `getReadonlyByRange` so callers can read a range from an SSZ composite vector in one go. 
- This is for cases like reading PTC windows see(https://github.com/ChainSafe/lodestar-z/issues/467 ), where we currently loop over the vector and call `getValue` one item at a time.


src/ssz/tree_view/array_composite.zig
```diff
@@ -148,6 +148,11 @@ pub fn ArrayCompositeTreeView(comptime ST: type) type {
             return self.chunks.getAllValues(allocator, length);
         }
 
+        pub fn getReadonlyByRange(self: *Self, allocator: Allocator, start_index: usize, out: []ST.Element.Type) ![]ST.Element.Type {
```
GrapeBaBa:
```suggestion
        pub fn getValuesByRangeInto(self: *Self, allocator: Allocator, start_index: usize, out: []ST.Element.Type) ![]ST.Element.Type {
```

src/ssz/tree_view/chunks.zig
```diff
@@ -519,17 +519,22 @@ pub fn CompositeChunks(
...
             const len = values.len;
             if (len == 0) return values;
 
             if (self.state.changed.count() != 0) {
                 return error.MustCommitBeforeBulkRead;
             }
 
-            const nodes = try allocator.alloc(Node.Id, len);
-            defer allocator.free(nodes);
+            const nodes = try self.state.allocator.alloc(Node.Id, len);
```
GrapeBaBa, line 547: here may need follow the [split change](https://github.com/ChainSafe/lodestar-z/pull/464/changes#diff-39141c4c37a0084c90d8e9f3a231537d2ad319c570564e0cd78a8d5bc364ce40R258) to make [test](https://github.com/ChainSafe/lodestar-z/pull/476/changes#diff-7687317f75037d644c499bd43f7f1e6932bbc12606c38d7ad29c572af82d8964R269) no need undefined

src/ssz/tree_view/chunks.zig
```diff
@@ -519,17 +519,35 @@ pub fn CompositeChunks(
...
         /// Fills `values` with all element values.
         pub fn getAllValuesInto(self: *Self, allocator: Allocator, values: []Value) ![]Value {
+            return self.getValuesByRangeIntoImpl(allocator, 0, values);
+        }
+
+        /// Fills `values` with element values starting at `start_index`.
+        pub const getValuesByRangeInto = if (isFixedType(ST.Element))
+            getValuesByRangeIntoFixed
+        else
+            getValuesByRangeIntoAlloc;
```
spiral-ladder, line 529: this split isn't necessary if both paths take an allocator anyway, we're still using `self.state.allocator` in the fixed path

spiral-ladder: to do a true split here we would need to split on the `getValuesByRangeIntoImpl` level so we can have one that takes the allocator and one that doesn't that works on fixed size types

spiral-ladder (changes requested):

PR #477, opened 2026-07-09
chore(release): v0.1.2-rc.10

spiral-ladder:
adds reset API to release for consumption, unblocks chainsafe/lodestar#8900


spiral-ladder:
I'm merging this to test, this is likely the cause of spec tests failing, would need a retroactive review

PR #479, opened 2026-07-09
ci: automate releases with release-please

### LINKED ISSUE #505: audit(release): restore verifiable source-to-package git trail

## Problem
The tracker says shipped code was based on a commit not present in the source tree: a deleted `v1.0.2` release reportedly copy/pasted main-branch state-transition code on top of `rc.10` plus an improperly committed tag. That creates provenance and possible OSS-license review risk.

Before Lodestar ships lodestar-z on mainnet, the source-to-package trail must be end-to-end verifiable.

## Acceptance Criteria
- Reconstruct the source commit/tag/package trail for the lodestar-z package Lodestar will consume.
- Ensure every published artifact maps to a reachable reviewed commit/tag.
- Document the cleanup of deleted/incorrect releases and tags.
- Confirm there is no unresolved license/provenance concern.
- Link final evidence here for Lodestar RC sign-off.

## References
- ChainSafe/lodestar#9647
- ChainSafe/lodestar-z#479

wemeetagain:
## Motivation

Our release process is manual and opaque: versions are hand-bumped in `package.json` via PR, tags and GitHub releases are created by hand, and publishing is a manual `workflow_dispatch` runnable on any ref. This has led to drift — right now `v0.1.2-rc.10` is on npm and GitHub releases, but its tag points to a commit that isn't on `main`, and `main`'s `package.json` still says `0.1.2-rc.9` (see #477 / #478).

Resolves #505


.release-please-manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+  ".": "0.1.2-rc.10"
```
wemeetagain:
```suggestion
  ".": "0.1.2"
```

build.zig.zon
```diff
@@ -1,6 +1,6 @@
 .{
     .name = .state_transition_z,
-    .version = "0.0.0",
+    .version = "0.1.2-rc.9", // x-release-please-version
```
wemeetagain:
```suggestion
    .version = "0.1.2", // x-release-please-version
```

.github/workflows/publish-bindings.yml
```diff
@@ -1,14 +1,35 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+      tag_name: ${{ steps.release.outputs.tag_name }}
+    steps:
+      - name: Run release-please
+        id: release
+        uses: googleapis/release-please-action@v4
```
spiral-ladder: we should consider sha tagging this like in #507 , but i'm also realizing we [dont sha tag this](https://github.com/search?q=org%3AChainSafe+release-please+language%3AYAML&type=code&p=3&l=YAML) across many repos

spiral-ladder (dismissed):

.github/workflows/publish-bindings.yml
wemeetagain: ah good call, will update here

spiral-ladder: I can open a pr for the rest of the deps. Seems worth a fix

spiral-ladder (dismissed):

release-please-config.json
```diff
@@ -0,0 +1,36 @@
...
+  "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+  "last-release-sha": "602480070524c0fe3e6099d7aae94ad5568d2d83",
+  "release-type": "node",
+  "include-component-in-tag": false,
+  "include-v-in-tag": true,
+  "versioning": "prerelease",
+  "prerelease": false,
+  "prerelease-type": "rc",
+  "bump-minor-pre-major": true,
+  "bump-patch-for-minor-pre-major": true,
```
nazarhussain, line 11: Why would we set this to `true`? 

A change with `feat:` will result `0.1.2` to `0.1.3` instead of `0.2.0`, is it really our expectations?

release-please-config.json
```diff
@@ -0,0 +1,36 @@
...
+    { "type": "perf", "section": "Performance Improvements" },
+    { "type": "refactor", "section": "Code Refactoring" },
+    { "type": "deps", "section": "Dependencies" },
+    { "type": "revert", "section": "Reverts" },
+    { "type": "build", "section": "Build System", "hidden": true },
+    { "type": "ci", "section": "Continuous Integration", "hidden": true },
+    { "type": "test", "section": "Tests", "hidden": true },
+    { "type": "style", "section": "Styles", "hidden": true },
+    { "type": "chore", "section": "Miscellaneous Chores", "hidden": true },
+    { "type": "docs", "section": "Documentation", "hidden": true }
```
nazarhussain: I feel these two are important enough to list down in the release notes and not hide.

.github/workflows/release.yml
```diff
@@ -1,14 +1,35 @@
...
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+      tag_name: ${{ steps.release.outputs.tag_name }}
+    steps:
+      - name: Run release-please
+        id: release
+        uses: googleapis/release-please-action@5c625bfb5d1ff62eadeeb3772007f7f66fdcf071 # v4.4.1
+        with:
+          config-file: release-please-config.json
+          manifest-file: .release-please-manifest.json
```
nazarhussain: I think it will load up default files, not need to list in workflow.

nazarhussain (commented):
One finding worth mentioning. 

**One cosmetic caveat (not a correctness problem):** the 0.1.2 side branch carried one real commit beyond the merge-base — e4f6cadc "bindings(pubkeys): add reset api". That exact commit never landed on main, but the same functionality arrived on main via b8ec2735 "feat(bindings): align BeaconStateView with IBeaconStateView (#347)", which sits inside the scan range. So the 0.1.3 changelog will list #347 even though part of its content (the pubkeys reset API) already shipped in 0.1.2.

release-please-config.json
wemeetagain: I'd be fine just deleting these two options and bumping to 1.0.
I don't think there's any reason to stay pre-1.0 anymore.

I also don't think the zig version (in build.zig.zon) should be pinned to the bindings version.

release-please-config.json
wemeetagain: I agree

matthewkeil (dismissed):

matthewkeil:
@wemeetagain and @nazarhussain I updated this a bit to help reduce publish failures. Added a build step to the CI to make sure that cross-compliation works as expected before we merge a PR.  This will prevent that failing the publish step when the release PR is merged.

matthewkeil (dismissed):

nazarhussain (dismissed):

.github/workflows/CI.yml
```diff
@@ -437,6 +437,34 @@ jobs:
...
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
+        with:
+          version: ${{ env.ZIG_VERSION }}
+          cache-key: ubuntu-latest-${{ env.ZIG_VERSION }}
+      - name: Install dependencies
+        run: |
+          pnpm install --frozen-lockfile
+      - name: Build release artifacts
+        run: |
+          pnpm zapi build-artifacts --optimize ReleaseSafe
```
spiral-ladder, line 466: I think we should reuse the output of this in the publish step, so we don't have to rebuild? see https://github.com/ChainSafe/lodestar-z/blob/ff6093117941934b4dff5058c35fa7422738a00c/.github/workflows/release.yml#L84

spiral-ladder: then that eliminates the need for #511

wemeetagain: Not a fan of the cross-workflow coupling, or of the amount of workflow refactoring for this.

Reusing this artifact doesn't materially address the security concern. `build.zig` still controls the exact binaries that the privileged job publishes. Separating jobs only prevents build-time code from directly exercising the narrowly-scoped npm publishing identity. So we prevent publishing multiple poisoned versions but it still can publish a poisoned version. That's not really a win.

And the build itself is already quite fast. So we're not really saving much time.

Imo duplication is cheaper than unnecessary orchestration here.

matthewkeil: I did consider the same thing @spiral-ladder but I tend to agree with @wemeetagain and decided to do the duplicated build for simplicity and long term maintainability

spiral-ladder: > Separating jobs only prevents build-time code from directly exercising the narrowly-scoped npm publishing identity

Yeah this is the main vector of attack splitting the jobs would prevent but it does seem extra paranoid and ur point about publishing poisoned versions is true too if `build.zig` was already compromised. so I think I concede that point, any hardening should probably be done at zbuild level then

matthewkeil (dismissed):

matthewkeil (approved):

spiral-ladder (approved):

PR #480, opened 2026-07-10
feat: grow pubkey cache by a calculated headroom

spiral-ladder:
Registry growth is capped at `MAX_PENDING_DEPOSITS_PER_EPOCH` new validators per epoch, so we reserve 3 months of worst-case growth, over a year at organic rates.

### DIFF
--- a/src/state_transition/cache/pubkey_cache.zig
+++ b/src/state_transition/cache/pubkey_cache.zig
@@ -12,7 +12,7 @@ pub const Index2PubkeyCache = std.ArrayList(bls.PublicKey);
 
 const headroom_epochs = (90 * 24 * 60 * 60) / (preset.SECONDS_PER_SLOT * preset.SLOTS_PER_EPOCH);
 
-// Reserve headroom so the native cache does not realloc on growth. Reallocs are unsafe
+/// Reserve headroom so the native cache does not realloc on growth. Reallocs are unsafe
 /// while the historical state worker reads the cache, until lodestar-z adds locking.
 /// Registry growth is capped at MAX_PENDING_DEPOSITS_PER_EPOCH new validators per epoch,
 /// reserve 3 months of worst-case growth, over a year at organic rates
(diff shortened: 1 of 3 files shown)


src/state_transition/cache/pubkey_cache.zig
```diff
@@ -9,6 +10,14 @@ pub const PubkeyIndexMap = std.AutoHashMap([48]u8, u64);
 /// Map from validator index to pubkey
 pub const Index2PubkeyCache = std.ArrayList(bls.PublicKey);
 
+const headroom_epochs = (90 * 24 * 60 * 60) / (preset.SECONDS_PER_SLOT * preset.SLOTS_PER_EPOCH);
+
+/// Reserve headroom so the native cache does not realloc on growth. Reallocs are unsafe
+/// while the historical state worker reads the cache, until lodestar-z adds locking.
+/// Registry growth is capped at MAX_PENDING_DEPOSITS_PER_EPOCH new validators per epoch,
+/// reserve 3 months of worst-case growth, over a year at organic rates
+pub const pubkey_cache_headroom = preset.MAX_PENDING_DEPOSITS_PER_EPOCH * headroom_epochs;
```
matthewkeil, line 19: Duplicating this in lodestar feels dirty and not DRY.  We should prob figure out a way to have this value in one place or the other and use the singular value across both uses

matthewkeil (approved):
LGTM!! 🚀 

Jus the nit about keeping that constant DRY but not a biggie for now. Can revisit this later

src/state_transition/cache/pubkey_cache.zig
nflaig: the value on the zig side is really irrelevant, this code path won't be executed ever, the worst case is ~3 months of a running node, which from my data on mainnet today, would affect 5-10 nodes. on top of that, this is the worst case `MAX_PENDING_DEPOSITS_PER_EPOCH` are mostly top-ups today as far as I know and will become more and more once `0x02` validators are more on mainnet over time. So roughly, the cache size we set on the ts side will hold ~1 year assuming 3/4 are top-ups, this is long enough

PR #481, opened 2026-07-10
feat(bindings): configurable pubkey cache growth step

nflaig:
## Motivation

Lodestar reserves pubkey cache capacity at startup with deliberately sized headroom (3 months of worst-case registry growth, see ChainSafe/lodestar#9640). When a `set()` outgrows that reservation the cache currently doubles its capacity, which at mainnet scale is a multi-million-slot jump (~450MB copied and twice that reserved) that is disproportionate to the sizing policy. Growth reallocations are also the unsafe window for concurrent readers on other threads, so they should stay small and predictable.

## Description

- Replace capacity doubling with a fixed growth step derived from the preset: `MAX_PENDING_DEPOSITS_PER_EPOCH` new validators per epoch for ~3 months at 12s slots (324k slots on mainnet preset). Each growth extends the reservation by the same worst-case window the initial headroom covers, at any network scale. No API change needed on the caller side.
- Expose `capacity` so callers can observe the reservation, e.g. to emit a remaining-headroom metric.
- Use `ensureTotalCapacityPrecise` for the index2pubkey array so growth follows the policy exactly instead of the ArrayList growth curve.
- Rename a local in `load()` that shadowed the new `capacity()` binding.

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -16,6 +16,8 @@ const max_stack_aggregate_pubkeys = 512;
 pub const State = struct {
     pubkey2index: PubkeyIndexMap = undefined,
     index2pubkey: Index2PubkeyCache = undefined,
+    /// Capacity added when a set() outgrows the current capacity, doubles when 0
+    growth_step: u32 = 0,
     initialized: bool = false,
 
     pub fn init(self: *State) !void {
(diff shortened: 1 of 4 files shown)


spiral-ladder (dismissed):
lgtm

wemeetagain (dismissed):
Looks good

nflaig (commented):
LGTM, self-review

spiral-ladder (approved):

PR #486, opened 2026-07-13
chore: silence debug log when used in release builds

spiral-ladder:
```console
➜ ./lodestar --help
Note: no -Dthread-count set, using cgroup-aware CPU count minus 1: 11
```

`debug.print` prints no matter what so even in upstream lodestar we get this message printed, but this should really be using `std.log` scoped behind proper optimize modes

### DIFF
--- a/bindings/napi/root.zig
+++ b/bindings/napi/root.zig
@@ -25,7 +25,7 @@ fn init(old_ref_count: u32) !void {
         var cpu_count: u64 = options.thread_count;
         if (options.thread_count == 0) {
             cpu_count = @max(try detectCpuCount(), 2) - 1;
-            std.debug.print(
+            std.log.debug(
                 "Note: no -Dthread-count set, using cgroup-aware CPU count minus 1: {}\n",
                 .{cpu_count},
             );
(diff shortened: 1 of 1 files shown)


bindings/napi/root.zig
```diff
@@ -25,7 +25,7 @@ fn init(old_ref_count: u32) !void {
         var cpu_count: u64 = options.thread_count;
         if (options.thread_count == 0) {
             cpu_count = @max(try detectCpuCount(), 2) - 1;
-            std.debug.print(
+            std.log.debug(
```
GrapeBaBa, line 28: this looks like a intentional print info, not a log?

spiral-ladder: then it's a question of if we should make this be configurable from lodestar. I think even previously with the rust bindings this thread count was silently set on the library side, which isn't ideal but that was the way it worked. Otherwise it comes up as a weird unconfigurable option that logs as noise to the user

GrapeBaBa: ok, got it

wemeetagain (approved):
This looks good.
Imo, ideally none of our prod code uses `std.debug.print`, rather uses `std.log` instead. That way we can easily configure all logs in a single way.

PR #487, opened 2026-07-14
fix(bindings): validate secret key hex length

GrapeBaBa:
## Summary

- Validate SecretKey.fromHex input length before copying into its fixed-size buffer.
- Require exactly 32 decoded bytes after stripping an optional 0x prefix.
- Add regression coverage for prefixed and unprefixed valid keys, short and long inputs, odd-length hex, and non-ASCII input.

## Why

Short hex strings previously reached a 32-byte slice operation without a length check, causing a ReleaseSafe bounds panic that aborted the Node process instead of throwing a JavaScript error. Checking the UTF-8 byte length before copying also prevents overlong strings from being silently truncated by N-API.


bindings/napi/blst.zig
```diff
@@ -272,8 +272,18 @@ pub const SecretKey = struct {
...
+        switch (try hex_string.len()) {
+            NativeSecretKey.serialize_size * 2,
+            NativeSecretKey.serialize_size * 2 + 2,
+            => {},
+            else => return error.InvalidSecretKeyLength,
+        }
+
         var hex_buf: [NativeSecretKey.serialize_size * 2 + 3]u8 = undefined;
         const hex = try hexFromString(hex_string, &hex_buf);
+        if (hex.len != NativeSecretKey.serialize_size * 2) {
```
wemeetagain, line 290: this second check is ensuring that a non-0x-prefixed hex string thats the wrong length doesn't slip through

wemeetagain (approved):

bindings/napi/blst.zig
```diff
@@ -278,8 +278,18 @@ pub const SecretKey = struct {
...
```
spiral-ladder, line 290: this is more suitable to be an assert, since if this check fails it means something is wrong with `hexFromString` and not the consumer input

program logic error = assert
input error = throw error

semes like a good practice to go by

spiral-ladder (changes requested):

bindings/napi/blst.zig
GrapeBaBa: - "0x" + 31 bytes has length 64, but becomes 62 after removing the prefix.
- An unprefixed 33-byte value has length 66 and remains 66.

Both pass the first check, so this should remain an input error rather than an assert.

spiral-ladder: you're right, doesn't that mean hexFromString is misleading though because it accepts any string and returns it unchanged if it's not 0x prefixed

spiral-ladder (approved):
lgtm, but an item to maybe leave for further consideration is the semantics behind `fromHexString`. Maybe we should consider assuming that string inputs would be either length 62/64 and asserting first 2 chars are 0x prefixed?

spiral-ladder:
seems like your original commits are not signed @GrapeBaBa

GrapeBaBa:
Superseded by #517, recreated from the latest `main` with a single verified signed commit.

PR #488, opened 2026-07-14
fix(bindings): check signature infinity by default

GrapeBaBa:
## Summary

- default Signature.fromHex sigInfcheck to true when signature validation is enabled
- keep the explicit sigInfcheck=false opt-out behavior
- add regression coverage using the compressed G2 point at infinity


spiral-ladder (approved):
good catch, this should [default to true](https://github.com/ChainSafe/blst-ts/blob/86c49590d37d4e1dd44b3b5ba604132f3b51d99d/src/lib.rs#L301)

spiral-ladder (changes requested):
Signed commit

GrapeBaBa:
Superseded by #509. The replacement contains the same code change on top of the latest `main`, with a GitHub-verified signed commit (`77c83436`). Closing this PR because its original change commit `1c525dcf` is unsigned.

PR #489, opened 2026-07-15
revert: refactor(clock): single public Clock; internalize SlotClock

GrapeBaBa:
Reverts ChainSafe/lodestar-z#463 since merged mistake which not using signed commit


twoeths (approved):

wemeetagain:
My feeling is that we don't need to revert commits that aren't signed.
Rather, moving forward, we should only allow signed commits.

GrapeBaBa:
> My feeling is that we don't need to revert commits that aren't signed. Rather, moving forward, we should only allow signed commits.

ok, so I can just close this PR?

PR #490, opened 2026-07-15
fix(bls): clean up partial thread pool initialization

GrapeBaBa:
## Summary

- track only successfully spawned workers during thread pool initialization
- shut down and join already-started workers when a later spawn fails
- release the pool allocation instead of leaking workers and shared state

### DIFF
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -121,14 +121,20 @@ pub fn init(allocator_: Allocator, io: std.Io, opts: Opts) (Allocator.Error || s
     std.debug.assert(opts.n_workers <= MAX_WORKERS);
 
     const pool = try allocator_.create(ThreadPool);
+    errdefer pool.deinit(io);
+
     pool.* = .{
         .allocator = allocator_,
-        .n_workers = opts.n_workers,
+        .n_workers = 0,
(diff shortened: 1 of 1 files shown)


src/bls/ThreadPool.zig
```diff
@@ -121,14 +121,20 @@ pub fn init(allocator_: Allocator, io: std.Io, opts: Opts) (Allocator.Error || s
...
         .allocator = allocator_,
-        .n_workers = opts.n_workers,
+        .n_workers = 0,
         .queue = .{},
     };
-    for (0..pool.n_workers) |i| {
+
+    for (0..opts.n_workers) |i| {
         pool.threads[i] = try std.Thread.spawn(.{}, workerLoop, .{ pool, io });
+        pool.n_workers += 1;
```
wemeetagain, line 134: how come this one-at-a-time is needed?
If there is an error part-way through, the `pool` will be destroyed? (Or does it need a `allocator_.destroy(pool)`?)

GrapeBaBa: yes, any error will `deinit` the pool right now

spiral-ladder (approved):

wemeetagain (approved):

PR #507, opened 2026-07-15
chore(ci): pin github actions with sha hashes

### LINKED ISSUE #500: ci: SHA-pin third-party actions in lodestar-z release workflows

## Problem
The tracker calls out unpinned GitHub Actions in lodestar-z release/publish jobs, including `mlugg/setup-zig@v2` and `pnpm/action-setup@v4`. Production release workflows should use SHA-pinned actions to reduce supply-chain risk.

This issue tracks the lodestar-z portion of that hardening. zapi has its own action pinning concern and should be handled separately in ChainSafe/zapi.

## Acceptance Criteria
- SHA-pin third-party actions used by lodestar-z release/publish workflows.
- Prefer full commit SHAs over mutable major tags for production release jobs.
- Document any actions intentionally left unpinned and why.
- Confirm CI/publish workflows still run after pinning.

## References
- ChainSafe/lodestar#9647
- Related OIDC/supply-chain epic: ChainSafe/lodestar#8544

spiral-ladder:
closes #500 

We also need to enforce SHA-only workflows instead of version tags once this is merged

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,9 +33,9 @@ jobs:
 
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - name: Install Zig
-        uses: mlugg/setup-zig@v2
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
         with:
           version: ${{ env.ZIG_VERSION }}
           cache-key: ${{ matrix.os }}-${{ env.ZIG_VERSION }}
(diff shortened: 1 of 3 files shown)


GrapeBaBa:
Is this really needed? Is it a best practice Github recommend?

spiral-ladder:
> Is this really needed? Is it a best practice Github recommend?

1. It is possible for an attack to force push version tags to run malicious code, see https://socket.dev/blog/trivy-under-attack-again-github-actions-compromise

2. This pattern is also enforced in lodestar ci, see the GitHub workflows there

This is also recommended as best practice: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

spiral-ladder:
Also we don't lose anything from doing this. It might be a bit more cumbersome to update workflows but that's about it

GrapeBaBa (approved):
lgtm

wemeetagain (approved):

Issue #495, opened 2026-07-15
chore(release): publish lodestar-z build with syncPubkeys shrink fix


lodekeeper:
Parent tracker: ChainSafe/lodestar#9647
Origin: checkbox from the blst-z RC-readiness tracker.

## Problem
The Lodestar tracker says Lodestar must bump to a lodestar-z build containing `ed05a99` (the `syncPubkeys` shrink fix) before native state view wiring is enabled. The currently tracked audited package version was `0.1.2`, which does not include that fix.

This issue tracks the lodestar-z side of making that fixed build available for Lodestar to pin at RC time.

## Acceptance Criteria
- Confirm the `syncPubkeys` shrink fix commit is present on the release branch intended for Lodestar consumption.
- Publish or otherwise prepare a lodestar-z package version containing that fix.
- Link the package version / tag / release notes here.
- Coordinate with the Lodestar-side pin update before native state view paths are enabled.

## References
- ChainSafe/lodestar#9647
- Related migration PR: ChainSafe/lodestar#8900

lodekeeper:
Leaving this open: ChainSafe/lodestar-z#537 is the active PR adding the missing `syncPubkeys(validators)` binding exposed through N-API.

Once #537 merges, the remaining step here is to publish/prepare the lodestar-z package version containing it and coordinate the Lodestar-side pin before native state view paths are enabled.

spiral-ladder:
@lodekeeper is this still relevant even with the new pk cache api?

lodekeeper:
Confirmed done — closing. The syncPubkeys shrink fix shipped and Lodestar consumes it:

- **Shrink fix present:** `ed05a99` is in lodestar-z **v1.0.0** (`ed05a99...v1.0.0` → `behind_by: 0`).
- **Binding merged:** #537 ("feat(bindings): add pubkey cache syncPubkeys") merged 2026-08-04, included in v1.0.0.
- **Published:** v1.0.0 on npm with prebuilds for all 6 supported targets.
- **Lodestar pin + wiring:** `unstable` pins `^1.0.0` (via ChainSafe/lodestar#8900) and actively calls the API — `pubkeyCache.syncPubkeys(...)` in `state-transition/src/cache/epochCache.ts` and `cli/src/cmds/beacon/handler.ts`.

[...]

Issue #494, opened 2026-07-15
fix(bindings): reject holes in native pubkey cache


lodekeeper:
Parent tracker: ChainSafe/lodestar#9647
Origin: checkbox from the blst-z RC-readiness tracker.

## Problem
Cache holes are currently treated as valid keys: any `index < items.len` is assumed populated. The tracker reports that on the shipped binary, `getOrThrow(5)` after sparse `set(10)` can return a `PublicKey` over uninitialized memory, and `aggregate()` may include it, causing silently wrong verification instead of an error.

This should be fixed centrally in lodestar-z rather than relying on scattered resets in Lodestar tests or call sites.

## Acceptance Criteria
- Track initialized/populated pubkey slots separately from array length/capacity.
- `getOrThrow(index)` fails for holes, even when `index < items.len`.
- `aggregate()` rejects or skips holes according to a clearly documented invariant, with no uninitialized memory reads.
- Add tests for sparse `set()` followed by `getOrThrow()` and `aggregate()`.

## References
- ChainSafe/lodestar#9647
- Related migration PR: ChainSafe/lodestar#8900

matthewkeil:
I think this is an impossible condition with regard to the consumer in lodestar.  Need to think through if we should do a simple check on the zig side.

https://github.com/ChainSafe/lodestar/blob/27da921314288e2cb85878cc5cc8e16ad7067e7d/packages/state-transition/src/block/processDeposit.ts#L116-L125

lodekeeper:
Good catch on `processDeposit.ts:116-125` — that path, plus the `syncPubkeys` wrapper in `packages/state-transition/src/cache/pubkeyCache.ts`, only ever calls `.addPubkey()`/`.set()` one index at a time, strictly sequential (`validatorIndex = validators.length - 1`, or `for (i = pubkeyCache.size; i < newCount; i++)`). Today's Lodestar call graph alone can't produce a hole.

But the hole doesn't come from the call graph — it comes from `set()` itself not enforcing the invariant it claims. `bindings/napi/pubkeys.zig:238-270`:

```zig
// Since the cache is append only, if the index is less than
// the cache's items length, we assume it already exists
if (idx < state.index2pubkey.items.len)
    return;
...
if (idx >= state.index2pubkey.items.len) {
    try state.index2pubkey.resize(allocator, idx + 1);
}
...
state.index2pubkey.items[@intCast(idx)] = try bls.PublicKey.uncompress(pubkey_bytes);
```

[...]

PR #510, opened 2026-07-15
(don't merge) fix(bindings): make pubkey cache set error-atomic

GrapeBaBa:
## Summary

- decode the BLS public key before mutating either cache direction
- reserve hash map capacity before extending the index cache, so the commit path cannot fail halfway
- mark the post-resize commit point with `errdefer comptime unreachable` to reject future fallible work
- add a regression test proving a failed set leaves the cache unchanged and a valid retry succeeds

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -249,21 +249,23 @@ pub fn set(index: js.Number, pubkey: js.Uint8Array) !void {
     if (pubkey_slice.len != 48) return error.InvalidPubkeyLength;
 
     const pubkey_bytes = pubkey_slice[0..48];
+    // Decode before resize; otherwise a failure would leave the append-only cache half-updated.
     const native_pubkey = try bls.PublicKey.uncompress(pubkey_bytes);
 
-    // Ensure capacity if needed
     if (idx >= state.index2pubkey.capacity) {
         const new_cap: u32 = @intCast(@max(idx + 1, state.index2pubkey.capacity + growth_step));
         try state.pubkey2index.ensureTotalCapacity(new_cap);
         try state.index2pubkey.ensureTotalCapacityPrecise(allocator, new_cap);
     }
+    // Reserve the map before resize so the reverse-index write cannot fail afterward.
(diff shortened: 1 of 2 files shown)


bindings/napi/pubkeys.zig
```diff
@@ -249,24 +249,26 @@ pub fn set(index: js.Number, pubkey: js.Uint8Array) !void {
...
+    const native_pubkey = try bls.PublicKey.uncompress(pubkey_bytes);
 
-    // Ensure capacity if needed
     if (idx >= state.index2pubkey.capacity) {
         const new_cap: u32 = @intCast(@max(idx + 1, state.index2pubkey.capacity + growth_step));
         try state.pubkey2index.ensureTotalCapacity(new_cap);
         try state.index2pubkey.ensureTotalCapacityPrecise(allocator, new_cap);
     }
+    // Reserve the map before resize so the reverse-index write cannot fail afterward.
+    state.pubkey2index.ensureUnusedCapacity(1) catch
```
spiral-ladder, line 261: I'm thinking if we should even do this, because we want to allocate a chunk at once (see above) with `growth_step` rather than size up by 1. Can we add this comment as a note

```suggestion
    // Reserve the map before resize so the reverse-index write cannot fail afterward.
    // NOTE: This should under normal operations be a no-op. We always grow by
    // `growth_step` in the code block above, since we do not want to grow by 1
    state.pubkey2index.ensureUnusedCapacity(1) catch
```

spiral-ladder (commented):
i have one doubt i'm still thinking about

bindings/napi/pubkeys.zig
GrapeBaBa: The `ArrayList` and `HashMap` have different growth factor which could make skip the code above because the idx lt list's capacity, but the map already has no room. That means this is not always no-op, but it is probably not by designed. In addition #495 and #494, I assume you will revisit the whole design of this cache, just hold on this PR to see if it is needed then

GrapeBaBa:
superseded #522

PR #511, opened 2026-07-15
chore(ci): split build and publish steps

spiral-ladder:
Another PR to further secure the CI process. While `build.zig.zon` is data-only and not executable, `build.zig` is. This would be a risk factor in the off chance `build.zig` is maliciously edited somehow. In our case that would be `zbuild`

Still missing from this PR is some sort of validation step. Splitting the build step still does have use though because that means the build step does not have access to tokens that the publish step would

Leaving this as a draft for thoughts first.

TigerBeetle does [release validation on a cronjob every 6 hours](https://github.com/tigerbeetle/tigerbeetle/blob/main/.github/workflows/release_validate.yml). See [validate_release](https://github.com/tigerbeetle/tigerbeetle/blob/97c7a8ef385270ebe0e1b75959d3d21d134629df/src/scripts/ci.zig#L92).

As a validation step we can borrow the above idea to further validate releases tho this might be overkill. But `build.zig` being executable does exist as a potential vector even if we delegate the build process to `zbuild`

cc @wemeetagain @matthewkeil

### DIFF
--- a/.github/workflows/publish-bindings.yml
+++ b/.github/workflows/publish-bindings.yml
@@ -7,8 +7,8 @@ env:
   NODE_VERSION: "24"
 
 jobs:
-  publish-bindings:
-    name: publish bindings
+  build-bindings:
+    name: build bindings
     # [Lodestar CI] still runs on Ubuntu 22.04, which ships an incompatible GLIBC version
     # compared to ubuntu-latest runners.
     # This [incompatibility] causes DLOPEN to fail when loading bindings on those runners.
@@ -21,8 +21,8 @@ jobs:
     runs-on: ubuntu-22.04
     permissions:
       contents: read
-      id-token: write
     steps:
+      # Check out the exact source revision selected for this release.
       - name: Checkout Repository
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - name: Setup pnpm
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-bindings.yml
```diff
@@ -31,37 +31,112 @@ jobs:
...
+          name: publish-packages
+          path: publish
+      # Recheck transfer integrity before contacting the registry.
+      - name: Verify npm packages
+        run: |
+          (cd publish && sha256sum -c SHA256SUMS)
+      # Publish verified tarballs with provenance and all scripts disabled.
+      - name: Publish to npm
+        shell: bash
+        run: |
```
spiral-ladder, line 103: this is very ugly, if we want to do something like this we should maybe do it in zig

.github/workflows/publish-bindings.yml
```diff
@@ -31,37 +31,112 @@ jobs:
...
-          else
-            echo "Stable version $VERSION -> npm default dist-tag"
-            echo "tag_flag=" >> "$GITHUB_OUTPUT"
-          fi
-      - name: Build and publish to npm
+      # Build and pack every platform package with lifecycle scripts disabled.
+      - name: Build npm packages
         run: |
           pnpm zapi build-artifacts --optimize ReleaseSafe
           pnpm zapi prepublish
```
nazarhussain, line 53: Why are have these here in this repo?

.github/workflows/publish-bindings.yml
```diff
@@ -31,37 +31,112 @@ jobs:
...
+
+          mkdir -p publish
+          for package_dir in npm/*; do
+            (
+              cd "$package_dir"
+              pnpm --config.ignore-scripts=true pack --pack-destination "$GITHUB_WORKSPACE/publish"
+            )
+          done
```
nazarhussain, line 61: We should use `pnpm` workspaces instead of manually iterating over the packages

.github/workflows/publish-bindings.yml
```diff
@@ -31,37 +31,112 @@ jobs:
...
+          publish_args=(--access public --ignore-scripts --provenance)
+          if [[ "$version" == *-* ]]; then
+            tag="${version#*-}"
+            tag="${tag%%.*}"
+            if [[ ! "$tag" =~ ^[a-z0-9][a-z0-9._-]*$ ]]; then
+              echo "Invalid npm dist tag: $tag" >&2
+              exit 1
+            fi
+            publish_args+=(--tag "$tag")
+          fi
```
nazarhussain, line 136: That code seems overkill for the simple publish step. Can be easily managed by the release-please workflow.

.github/workflows/publish-bindings.yml
matthewkeil: agreed

matthewkeil (commented):
A lot of this scope was handled in the release-please PR #479 .  Maybe we can check that one out and see if it can be massaged to include anything that this one adds to the scope in that one.  I think most is covered there though, and it adds some other checks and cleans a few things up.  Please let me know if you still think there is value in updating this and getting it merged and we can do that.

spiral-ladder:
> A lot of this scope was handled in the release-please PR #479

largely yes but I think there is still some value in splitting the [`pnpm zapi build-artifacts` step](https://github.com/ChainSafe/lodestar-z/blob/ff6093117941934b4dff5058c35fa7422738a00c/.github/workflows/release.yml#L84) (that's where `zig build` runs) from the publish step because that's where malicious code can run. Ideally we should have already-built binaries so that the publish step does not need to run `zig build` at all, that eliminates attacks that can come from `build.zig` running zig code

see: https://github.com/ChainSafe/lodestar-z/pull/479#discussion_r3620181243

wemeetagain:
>  that eliminates attacks

I'm not sure I understand. Can't an attacker modify build.zig in step 1 to poison the artifacts, and in step 2 publish the artifacts?

spiral-ladder:
> > that eliminates attacks
> 
> I'm not sure I understand. Can't an attacker modify build.zig in step 1 to poison the artifacts, and in step 2 publish the artifacts?

theoretically one could run any arbitrary code in `build.zig` (but ofc that would take compromising `zbuild` in our case). it's very unlikely but still possible

wemeetagain:
> very unlikely

But this sounds like the motivation of this PR based on the PR description?

spiral-ladder:
> > very unlikely
> 
> But this sounds like the motivation of this PR based on the PR description?

yep, but if we split the build and publish steps in #479 ([comment](https://github.com/ChainSafe/lodestar-z/pull/479#discussion_r3620181243)), this PR is no longer needed

wemeetagain:
lets close this and sort it out in #479

PR #513, opened 2026-07-15
fix(bls): convert pippenger scratch bytes to element counts

GrapeBaBa:
## Summary

- convert blst Pippenger scratch sizes from bytes to `u64` element counts
- validate caller-provided scratch slices using the same units
- allocate exact-minimum G1 and G2 scratch buffers in tests


spiral-ladder (approved):
good catch, thanks!

nazarhussain (approved):

PR #514, opened 2026-07-16
fix(bindings): validate class types before unwrap

GrapeBaBa:
## Summary

- route every BLST class extraction through zapi typed conversion so N-API type tags are checked before `napi_unwrap`
- pin zapi to the merge commit of [ChainSafe/zapi#50](https://github.com/ChainSafe/zapi/pull/50) until the pointer-argument fix is released
- add regression coverage for wrong-class values across signature/public-key aggregation and randomized verification paths

## Motivation

Raw `Env.unwrap` accepts any wrapped N-API object. Passing a `SecretKey` where a `PublicKey` or `Signature` is expected can therefore reinterpret a smaller native allocation as a larger structure and read beyond its bounds.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -40,8 +40,8 @@
             .hash = "zig_yaml-0.1.0-C1161kFWAwDxjKAFmklKwWVDvz2mmq0Q__bDhGGjeyd3",
         },
         .zapi = .{
-            .url = "https://github.com/ChainSafe/zapi/archive/refs/tags/zapi-v3.1.0.tar.gz",
-            .hash = "zapi-3.1.0-rIqzUQJsBADCFUrmxoTZyOhSTkcW6m3t_OgOHc6oxnm9",
+            .url = "https://github.com/ChainSafe/zapi/archive/9dd211167c4774c5a79e3d6fd360c22adca0a138.tar.gz",
+            .hash = "zapi-3.1.0-rIqzUQtsBADx3WYJ6Fe4BY1M3Q6zGpsti2B_cvgxZgto",
         },
         .zbench = .{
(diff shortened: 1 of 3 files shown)


build.zig.zon
```diff
@@ -40,8 +40,8 @@
             .hash = "zig_yaml-0.1.0-C1161kFWAwDxjKAFmklKwWVDvz2mmq0Q__bDhGGjeyd3",
         },
         .zapi = .{
-            .url = "https://github.com/ChainSafe/zapi/archive/refs/tags/zapi-v3.1.0.tar.gz",
-            .hash = "zapi-3.1.0-rIqzUQJsBADCFUrmxoTZyOhSTkcW6m3t_OgOHc6oxnm9",
+            .url = "https://github.com/ChainSafe/zapi/archive/9dd211167c4774c5a79e3d6fd360c22adca0a138.tar.gz",
```
GrapeBaBa, line 43: need new zapi version

spiral-ladder (approved):

build.zig.zon
nazarhussain: LGTM. We can release `zapi - 3.1.1` if that's what we need to use here. Pinning to an untagged commit does not feel right.

spiral-ladder: FWIW i think it's OK until we do a release of lodestar-z, then it should be tagged with a proper release

nazarhussain: Yes it feels ok if we had a workflow in place that reminds us to pin to a version before release of lodestar-z, which we don't have and there are probable chances we forget to do so. 

I feel this PR is not a blocker for any issue, so we can hold it. Or we can release zapi newer version, which is one click away.

PR #516, opened 2026-07-16
refactor(bindings): rename blst Lifecycle to State

nazarhussain:
## Motivation

Follow-up to the naming discussion on #469: `blst.lifecycle` was a one-off, while `pool`, `pubkeys`, and `config` already share a `State`/`state` convention for native-only module state. As suggested by @wemeetagain, this adopts that convention for blst.

## Description

- Rename `Lifecycle`/`pub var lifecycle` to `State`/`pub var state` in `bindings/napi/blst.zig`, and `initThreadPool`/`deinitThreadPool` to `init(n_workers)`/`deinit`.
- Move the module-level `var thread_pool` into `State` as an owned field, matching the shape of the sibling modules (e.g. `pool_rc` in `pool.zig`), not just the name. Internal use sites now read `state.thread_pool`.
- `root.zig` now reads uniformly: `blst.state.init(...)` next to `pool.state.init()` / `pubkeys.state.init()` / `config.state.init()`.
- Fix (pre-existing): add `errdefer blst.state.deinit()` in `root.zig` `init()`. Previously, if `pool.state.init()` or `pubkeys.state.init()` failed after the BLS pool was created, the `errdefer napi_io.deinit()` tore down the io while the pool's worker threads still referenced it, and `thread_pool` stayed non-null so any retry failed with `PoolExists`. The idempotent sibling states self-heal on retry; blst was the only one that latched.

One deliberate divergence kept: `blst.state.init` still returns `error.PoolExists` on double init instead of the siblings' idempotent early-return, since it takes `n_workers` — silently ignoring a second init with a different worker count would hide a bug.

## Validation

- `zig build build-lib:bindings` compiles clean.
- `pnpm test`: 158 passed; the single failure is a missing local ERA fixture (`fixtures/era/mainnet-01628-47ac89fb.era`), unrelated to this change.

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -1,13 +1,13 @@
 //! NAPI bindings for BLS (blst) cryptographic operations used by lodestar.
 //!
-//! This module uses a **Zig ThreadPool** (`thread_pool`) — a fixed-size pool of OS threads
-//! initialized once via `initThreadPool`. Used by synchronous NAPI functions (`aggregateVerify`,
+//! This module uses a **Zig ThreadPool** (`state.thread_pool`) — a fixed-size pool of OS threads
+//! initialized once via `state.init`. Used by synchronous NAPI functions (`aggregateVerify`,
 //! `verifyMultipleAggregateSignatures`) to fan out pairing checks across worker threads. The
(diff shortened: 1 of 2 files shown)


spiral-ladder (approved):
lgtm,

wemeetagain (approved):

PR #517, opened 2026-07-16
fix(bindings): validate secret key hex length

GrapeBaBa:
## Summary

- validate `SecretKey.fromHex` input length before copying into its fixed-size buffer
- require exactly 32 decoded bytes after stripping an optional `0x` prefix
- cover prefixed and unprefixed valid keys, short and long inputs, odd-length hex, and non-ASCII input

## Why

Short hex strings previously reached a 32-byte slice operation without a length check, causing a ReleaseSafe bounds panic that aborted the Node process instead of throwing a JavaScript error. Checking the UTF-8 byte length before copying also prevents overlong strings from being silently truncated by N-API.

Supersedes #487, which contains unsigned commits.

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -278,8 +278,18 @@ pub const SecretKey = struct {
 
     /// Creates a `SecretKey` from a hex string.
     pub fn fromHex(hex_string: js.String) !SecretKey {
+        switch (try hex_string.len()) {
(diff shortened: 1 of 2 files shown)


spiral-ladder (approved):
lgtm, with [same comment as before](https://github.com/ChainSafe/lodestar-z/pull/487#pullrequestreview-4712600743)

wemeetagain (approved):

PR #518, opened 2026-07-16
refactor(bindings/pubkeys): simplify allocation strategy for aggregate

### LINKED ISSUE #496: perf(bindings): avoid page_allocator path for mainnet-size aggregate batches

## Problem
`aggregate()` falls back to `page_allocator` above 512 indices. Mainnet committees are commonly around 1100 validators, so gossip aggregate verification takes the mmap/munmap path on normal mainnet-sized inputs.

This is not a release-blocking correctness issue, but it is a cheap performance win before shipping the native BLS path.

## Acceptance Criteria
- Revisit the allocator threshold used by `aggregate()` for batch sizes around current mainnet committees.
- Avoid per-call mmap/munmap on normal mainnet committee sizes.
- Add or update a benchmark/test that covers a ~1100-index aggregate path.
- Document any chosen threshold and why it is safe.

## References
- ChainSafe/lodestar#9647

spiral-ladder:
closes #496 (indirectly, since we won't be using a page allocator for aggregation allocations anymore)

- 512 as a stack size was arbitrary, and it is a bit ugly to maintain
- separate allocators for cache and aggregation, since they're different lifetimes
- also scope aggregation allocator (at comptime) appropriately to the `aggregate` fn only to prevent accidental misuse by the cache

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -14,13 +14,6 @@ const preset = @import("preset").preset;
 /// See note above `growth_step`.
 const cache_allocator = std.heap.page_allocator;
 
-/// Owns the short-lived aggregation allocations, if compiled
-/// in Debug mode.
-///
-/// in Release{Safe|Fast} modes, this is `std.heap.c_allocator`. See
-/// `aggregation_allocator` for details.
-var aggregation_gpa: std.heap.DebugAllocator(.{}) = .init;
-
 const default_initial_capacity: u32 = 0;
 
 /// Capacity added when a set() outgrows the current capacity. Covers ~3 months of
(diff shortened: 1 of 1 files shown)


bindings/napi/pubkeys.zig
```diff
@@ -7,12 +8,20 @@ const Index2PubkeyCache = @import("state_transition").Index2PubkeyCache;
...
+/// NOTE: we should ideally never reallocate the pubkey2index after initialization.
+/// See note above `growth_step`.
+const cache_allocator = std.heap.page_allocator;
+
+/// Owns the short-lived aggregation allocations, if compiled
+/// in Debug mode.
+///
+/// in Release{Safe|Fast} modes, this is `std.heap.c_allocator`. See
+/// `aggregation_allocator` for details.
+var aggregation_gpa: std.heap.DebugAllocator(.{}) = .init;
```
GrapeBaBa: I suspect it don't need be global, and it looks like we missing a `deinit` call for `DebugAllocator`

GrapeBaBa:
not must, if we still want to use a bit stack allocation, https://ziglang.org/documentation/0.16.0/std/#std.heap.StackFallbackAllocator this seems easy to use

bindings/napi/pubkeys.zig
spiral-ladder: You're right, fixed in ff23368, thanks for the catch

GrapeBaBa (approved):
lgtm

spiral-ladder:
> not must, if we still want to use a bit stack allocation, https://ziglang.org/documentation/0.16.0/std/#std.heap.StackFallbackAllocator this seems easy to use

i think a sensible upper bound can be found for stack allocations but for now this is simpler, we can revisit this in future maybe

PR #519, opened 2026-07-16
fix(bindings): reject holes in pubkey cache 

### LINKED ISSUE #494: fix(bindings): reject holes in native pubkey cache

bomanaps:
Closes https://github.com/ChainSafe/lodestar-z/issues/494

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -22,13 +22,17 @@ const growth_step: u32 = preset.MAX_PENDING_DEPOSITS_PER_EPOCH * ((90 * 24 * 60 
 const State = struct {
     pubkey2index: PubkeyIndexMap = undefined,
     index2pubkey: Index2PubkeyCache = undefined,
+    populated: std.DynamicBitSetUnmanaged = .{},
     initialized: bool = false,
 
     pub fn init(self: *State) !void {
         if (self.initialized) return;
         self.pubkey2index = PubkeyIndexMap.init(allocator);
         try self.pubkey2index.ensureTotalCapacity(default_initial_capacity);
+        errdefer self.pubkey2index.deinit();
         self.index2pubkey = try Index2PubkeyCache.initCapacity(allocator, default_initial_capacity);
+        errdefer self.index2pubkey.deinit(allocator);
+        self.populated = try std.DynamicBitSetUnmanaged.initEmpty(allocator, default_initial_capacity);
         self.initialized = true;
     }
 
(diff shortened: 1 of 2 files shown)


matthewkeil:
@bomanaps thank you very much for the contribution!!  I think the AI hallucinated the suggested fix though.  I think we need to do some deeper architectural thought for the solution.  I wont close this just yet, but wanted to give you and others a heads up.  cc @spiral-ladder @GrapeBaBa

spiral-ladder:
@bomanaps Thanks for the PR! For more details see this [comment](https://github.com/ChainSafe/lodestar-z/issues/494#issuecomment-5006848914). This situation should never occur since we treat the data structure that holds the pubkeys as append-only. So the AI isn't really wrong that there's a mismatch in `set()` but the use-case makes it safe in this situation

bomanaps:
> Thanks for the PR! For more details see this [comment](https://github.com/ChainSafe/lodestar-z/issues/494#issuecomment-5006848914). This situation should never occur since we treat the data structure that holds the pubkeys as append-only. So the AI isn't really wrong that there's a mismatch in `set()` but the use-case makes it safe in this situation

So you want me to swap the bitset for a two-line set()  assert idx == items.len on growth, error.NonSequentialPubkeyIndex on gap and split the call syncPubkeys from NAPI refactor into a follow-up issue?

spiral-ladder:
> So you want me to swap the bitset for a two-line set() assert idx == items.len on growth, error.NonSequentialPubkeyIndex on gap and split the call syncPubkeys from NAPI refactor into a follow-up issue?

I think #522 will likely supercede this, it's a re-work of the entire pubkey cache

bomanaps:
> > So you want me to swap the bitset for a two-line set() assert idx == items.len on growth, error.NonSequentialPubkeyIndex on gap and split the call syncPubkeys from NAPI refactor into a follow-up issue?
> 
> I think #522 will likely supercede this, it's a re-work of the entire pubkey cache

Alright

PR #521, opened 2026-07-21
fix(bls): preserve aggregate outputs on failure

GrapeBaBa:
## Motivation

`aggregateWithRandomness` updated `pk_out` before the signature MSM completed. Keeping both output writes after the fallible work avoids exposing a partially updated result when the function returns an error.

## Description

Defer both affine output writes until the G1 and G2 MSMs complete successfully.


spiral-ladder (commented):
Reordering seems reasonable but I'm not sure if a regression test is necessary, if we fail one of the allocations would the output even be meaningful?

GrapeBaBa:
> Reordering seems reasonable but I'm not sure if a regression test is necessary, if we fail one of the allocations would the output even be meaningful?

make sense

GrapeBaBa:
Addressed in 4423bec9: removed the allocation-failure regression test and kept the output-write reordering.

spiral-ladder (approved):
lgtm

wemeetagain (approved):

PR #522, opened 2026-07-21
refactor(pubkeys): centralize the process-wide cache

wemeetagain:
## Summary

- Replace the binding-owned pubkey caches with one process-wide `PubkeyCache`, defined in state transition and owned by N-API.
- Use insertion-ordered `ArrayHashMap` storage, a cache-owned `RwLock`, and bounded parallel decompression.
- Thread the shared cache through `EpochCache` and signature construction.
- Add ABI-locked PKIX persistence for fast startup.

## Invariants and trust

Validator-registry `(index, pubkey)` history is unforkable by design. Every state processed by a node has a registry that is a prefix of the process-wide cache, and an existing index never maps to another pubkey.

PKIX is only for application-owned snapshots from the intended network. Loading checks the format, ABI, bounds, exact file size, and checksum. For speed, it intentionally skips semantic validation such as key uniqueness, compressed-to-affine correspondence, and BLS point validity. The checksum detects accidental corruption; it does not authenticate files.

## Impact

The JavaScript API remains append-oriented; replaying an existing index is a no-op. `save`, `load`, and `reset` are restricted to the first N-API environment for the lifetime of the shared cache. Previous PKIX formats are not supported.

## AI assistance

Implementation and review were primarily AI-assisted, with design direction and iterative review from the author and team.


src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,332 @@ const bls = @import("bls");
...
+/// and retains all allocated capacity.
+pub const PubkeyCache = struct {
+    allocator: std.mem.Allocator,
+    entries: PubkeyMap,
+    hash_key: HashKey,
+    /// Defensively enforces the application's single-writer contract and
+    /// serializes cache allocator use.
+    writer_mutex: std.Io.Mutex,
+    /// Guards movable map storage; acquired after writer_mutex when both are needed.
+    lock: std.Io.RwLock,
```
spiral-ladder, line 75: i haven't looked at the entire change yet but this seems directionally correct, when i was having a long convo with codex about this i came to this conclusion that it seems like a `RwLock` is unavoidable (and is very minimal overhead in our single writer many reader use case) for the situation where a write would grow the cache and move memory while a read was happening

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
-/// parallelizing BLS pubkey decompression across the `io` executor's worker pool
-/// via `std.Io.Group.concurrent`. The `pubkey_to_index` HashMap is updated
-/// single-threaded at the end (HashMap is not thread-safe).
-pub fn syncPubkeysParallel(
+};
+
+/// The dense entry index is the validator index. Production code only appends,
+/// so insertion order permanently supplies both lookup directions:
+/// compressed pubkey -> validator index and validator index -> affine pubkey.
+pub const PubkeyMap = std.ArrayHashMapUnmanaged(
```
GrapeBaBa: nit: use std.array_hash_map.Custom instead in 0.16.0

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
-        const pubkey = &validators[i].pubkey;
-        index_to_pubkey.items[i] = bls.PublicKey.uncompress(pubkey) catch {
-            uncompress_error.store(true, .release);
-            return;
-        };
+    pub fn eql(
+        _: PubkeyHashContext,
+        lhs: [48]u8,
+        rhs: [48]u8,
+        rhs_index: usize,
```
GrapeBaBa: nit: two ignore args using different ways

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
+);
+
+// Leave room for ArrayHashMap's geometric growth and for the dense allocation
+// size multiplication on 32-bit targets.
+const dense_capacity_limit =
+    std.math.maxInt(usize) / PubkeyMap.DataList.capacityInBytes(1);
+pub const max_capacity: usize = @min(
+    @as(usize, std.math.maxInt(u32)),
+    dense_capacity_limit / 2,
+);
```
GrapeBaBa, line 61: This looks like used for 32-bit target, but some codes in bls seems only support 64-bit target, should we support both?

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
-    try syncPubkeys(allocator, &validator_ptrs, &pubkey_to_index, &index_to_pubkey);
-
-    try testing.expectEqual(@as(usize, count), index_to_pubkey.items.len);
-    try testing.expectEqual(@as(u32, count), pubkey_to_index.count());
-
-    for (0..count) |i| {
-        const idx = pubkey_to_index.get(pubkeys[i]).?;
-        try testing.expectEqual(@as(u64, i), idx);
+    pub fn count(self: *const PubkeyCache, io: std.Io) u32 {
+        self.lockShared(io);
```
GrapeBaBa, line 134: Why shard lock using Uncancelable version but exclusive lock using cancelable version?

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
+        const suffix = validators[old_len..];
+        const prepared = try self.allocator.alloc(bls.PublicKey, suffix.len);
+        defer self.allocator.free(prepared);
+
+        const batch_count = (suffix.len - 1) / uncompress_batch_size + 1;
+        const batch_errors = try self.allocator.alloc(?bls.BlstError, batch_count);
+        defer self.allocator.free(batch_errors);
+        @memset(batch_errors, null);
+
+        var group: std.Io.Group = .init;
```
GrapeBaBa, line 295: `writer_mutex` seems removed from initial version, which seems could avoid this function hold the lock too much time, what's the intention for removing it

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
+    /// Aggregate the pubkeys at the requested validator indices.
+    pub fn aggregate(
+        self: *const PubkeyCache,
+        io: std.Io,
+        indices: []const u64,
+    ) !bls.PublicKey {
+        self.lockShared(io);
+        defer self.unlockShared(io);
+
+        if (indices.len == 0) return error.InvalidLength;
```
GrapeBaBa: nit: this check seems no need inside the lock

GrapeBaBa:
The append only writing makes me thinking about in WAL(write-ahead-log) which often use segment storage approach, which seems could be used.

- Store `validator_index -> pubkey` in fixed-size, append-only segments.   
- Forward readers acquire-load `published_count` and locate values using          
 segment/offset arithmetic, without taking the cache `RwLock`.                       
- Published segments are immutable and are only freed after all workers have      
 stopped, only last segment is mutable for writing.  
 
This makes `validator_index -> pubkey` query basic no lock, and don't need reallocate large memory and copy all original kvs. 
 
It is not a review comment for this PR but what I thought, it is probably a bit complex than current implementation and I am not sure if it is worse or better in hot path access pattern without actual benchmark

bindings/src/pubkeys.d.ts
```diff
@@ -1,30 +1,31 @@
...
   get(index: number): PublicKey | undefined;
   /** Same as get(), but throws if the index is not in the cache */
   getOrThrow(index: number): PublicKey;
   /** Aggregate cached public keys by validator index */
   aggregate(indices: number[]): PublicKey;
   /** Get validator index by pubkey bytes */
   getIndex(pubkey: Uint8Array): number | null;
-  /** Set both directions atomically — impl owns the PublicKey.fromBytes() deserialization */
+  /** Append the next pubkey; already-cached indices are ignored. */
   set(index: number, pubkey: Uint8Array): void;
```
spiral-ladder: makes me wonder if we should just rename this to `append()` or have both apis exposed to be clearer

bindings/src/pubkeys.d.ts
```diff
@@ -1,30 +1,31 @@
...
   /** Number of entries */
   readonly size: number;
-  /** Allocated native capacity */
+  /** Current allocated native capacity */
```
spiral-ladder: nit: what's the difference between current allocated and allocated?

bindings/napi/pubkeys.zig
```diff
@@ -204,105 +141,72 @@ pub fn get(index: js.Number) !?blst_bindings.PublicKey {
...
     const idx = try index.toU32();
+    const io = napi_io.get();
+    if (!state.initialized) return error.PubkeyIndexNotInitialized;
 
-    // Since the cache is append only, if the index is less than
-    // the cache's items length, we assume it already exists
-    if (idx < state.index2pubkey.items.len)
-        return;
+    // Existing entries are immutable, so replayed indices need no validation.
+    if (idx < state.cache.count(io)) return;
```
spiral-ladder: this seems inconsistent with the [inner fn](https://github.com/ChainSafe/lodestar-z/pull/522/changes#diff-49d7fec9ae12126cf0eb9fea69f9b5f76ab3b78bb11e6ecf4967ecb9d9be0315R250-R254):

```zig
        const current_len = self.entries.count();
        if (index < current_len) {
            if (!std.mem.eql(
                u8,
                &self.entries.keys()[@intCast(index)],
                &pubkey,
            )) return error.ConflictingPubkey;
            return;
        }
```

I think we should also throw an error at this level?

src/state_transition/cache/pubkey_cache.zig
```diff
@@ -3,254 +3,351 @@ const bls = @import("bls");
...
+
+        const current_len = self.entries.count();
+        if (index < current_len) {
+            if (!std.mem.eql(
+                u8,
+                &self.entries.keys()[@intCast(index)],
+                &pubkey,
+            )) return error.ConflictingPubkey;
+            return;
+        }
```
spiral-ladder, line 251: we can reorder this to avoid uncompressing before comparing the pubkeys

```suggestion
        try self.lock.lock(io);
        defer self.lock.unlock(io);
        
        const current_len = self.entries.count();
        if (index < current_len) {
            if (!std.mem.eql(
                u8,
                &self.entries.keys()[@intCast(index)],
                &pubkey,
            )) return error.ConflictingPubkey;
            return;
        }
        const affine = try bls.PublicKey.uncompress(&pubkey);
```

bindings/test/serializedState.ts
```diff
@@ -0,0 +1,29 @@
+import {ssz} from "@lodestar/types";
+
+/** Ninety days of worst-case validator-registry growth on the mainnet preset. */
+export const DEFAULT_PUBKEY_CACHE_HEADROOM = 324_000;
```
spiral-ladder, line 4: this value seems different from the previously calculated `growth_step`?

spiral-ladder (commented):
did a first pass, i wonder if it's worth adding a separate `append` api just so we don't have to be ambiguous about the usage of `set`

src/state_transition/cache/pubkey_cache.zig
wemeetagain: Yeah, I thought it really isn't necessary based on the usage patterns of the pubkey cache.

Really we have this sort of usage:

Startup:
- Either load from "pkix" file, or start fresh with capacity (pkix.load/install, initWithCapacity)
- Then sync any additional pubkeys from an anchor state (syncPubkeys)

Normal runtime:
- Append new pubkeys one by one, with a single writer, the block importer. (append)
- Very occasionally, once per N months of continuous operation, we need to resize (ensureTotalCapacity)
- Lots of reads, maybe from multiple workers, for gossip validation, block signature validation, STF (get, getPubkey, getIndices, getPubkeys)

Shutdown:
- Try to save the pubkey cache contents to a "pkix" file (pkix.save)

---

So given this pattern, we don't really care about reads hanging during syncPubkeys, because syncPubkeys is only a startup operation.

bindings/src/pubkeys.d.ts
wemeetagain: Yes `set` is not the real operation we need. Imo it is a holdover from overriding / using javascript `Map` interface which implies a more flexible semantics than we need.

bindings/src/pubkeys.d.ts
wemeetagain: right, why is the jsdoc for `size` not "Current number of entries"

bindings/napi/pubkeys.zig
wemeetagain: ah right. this short-circuit should not exist

src/state_transition/cache/pubkey_cache.zig
wemeetagain: Yes thats a good point. I don't think its necessary, though I would like to avoid hardcoded 64-bit-only code. Worth a broader discussion.

src/state_transition/cache/pubkey_cache.zig
wemeetagain: Using uncancelable here to avoid introducing `error.Cancelled` error for reads (`count`, `get`, `getPubkey`, etc).

The mutating operations already return errors so just left it cancelable.

src/state_transition/cache/pubkey_cache.zig
wemeetagain: The suggested reorder optimizes exact replays, but makes the normal path worse: every new validator append would hold the exclusive lock during BLS decompression, blocking all readers.

bindings/test/serializedState.ts
wemeetagain: The new model is:
  - Preallocate validator count plus headroom during startup.
  - Use `pubkeyCache.ensureCapacity` according to some application trigger to resize precisely
  - Grow automatically using the map’s native growth policy if that capacity is eventually exhausted
  - No longer grow in fixed 90-day chunks inside the N-API binding.
  
  Here, I think the value is the same, although not parameterized as before

bindings/napi/pubkeys.zig
```diff
@@ -1,199 +1,136 @@
...
 
-    // Write index2pubkey entries
-    try writer.writeAll(std.mem.sliceAsBytes(state.index2pubkey.items));
+    var write_buf: [4096]u8 = undefined;
+    var file_writer = atomic_file.file.writer(io, &write_buf);
+    try pkix.save(&state.cache, io, &file_writer.interface);
 
     try file_writer.end();
+    try atomic_file.file.sync(io);
+    try atomic_file.replace(io);
```
GrapeBaBa, line 67: Though it may not need strong guarantee, the real atomic write may need another fsync after rename like https://github.com/postgres/postgres/blob/b8d9cf512c1259f97f9896593cc1c8352c1118ac/src/common/file_utils.c#L478-L535

bindings/napi/pubkeys.zig
```diff
@@ -1,199 +1,136 @@
...
 
+    /// Last-environment cleanup runs after JS calls stop. Remaining view
+    /// finalizers do not dereference their retained pubkey-cache pointer.
     pub fn deinit(self: *State) void {
         if (!self.initialized) return;
-        self.pubkey2index.deinit();
-        self.index2pubkey.deinit(cache_allocator);
+        self.cache.deinit();
         self.initialized = false;
+        self.control_env = null;
```
GrapeBaBa, line 35: nit: self.* = undefined; Poison memory after free to catch use-after-free

bindings/napi/pubkeys.zig
```diff
@@ -1,199 +1,136 @@
...
-    std.mem.writeInt(u32, header[8..12], @intCast(state.index2pubkey.capacity), .little);
+    const path = try file_path.toOwnedSlice(allocator);
+    defer allocator.free(path);
+    const io = napi_io.get();
 
-    var write_buf: [4096]u8 = undefined;
-    var file_writer = file.writer(io, &write_buf);
-    var writer = &file_writer.interface;
-    try writer.writeAll(header[0..12]);
+    if (!state.initialized) return error.PubkeyIndexNotInitialized;
```
GrapeBaBa: nit: this check seems could be executed more early

bindings/napi/pubkeys.zig
```diff
@@ -204,105 +141,72 @@ pub fn get(index: js.Number) !?blst_bindings.PublicKey {
...
-    // Owns the short-lived aggregation allocations.
-    // Do not use for cache allocations.
-    const aggregation_allocator = if (builtin.mode == .Debug)
-        aggregation_gpa.allocator()
-    else
-        std.heap.c_allocator;
-
-    const pks = try aggregation_allocator.alloc(bls.PublicKey, len);
-    defer aggregation_allocator.free(pks);
+    defer if (len > indices_stack.len) allocator.free(exact_indices);
```
GrapeBaBa, line 154: nit: may use StackFallbackAllocator

src/state_transition/cache/pkix.zig
```diff
@@ -0,0 +1,272 @@
...
+        @alignOf(AffinePoint),
+        @offsetOf(AffinePoint, "y"),
+        @sizeOf(BaseField),
+        @alignOf(BaseField),
+        @sizeOf(Limbs),
+        @alignOf(Limbs),
+    },
+);
+
+const abi_fingerprint = std.hash.XxHash3.hash(
```
spiral-ladder, line 121: Q: should we fingerprint the underlying bls library version?

src/state_transition/cache/pkix.zig
```diff
@@ -0,0 +1,272 @@
...
+    payload_checksum_seed,
+    abi_description,
+);
+
+const PayloadLayout = struct {
+    keys_size: usize,
```
spiral-ladder: this doesn't seem to be used by the public interface, maybe we can drop it?

bindings/napi/pubkeys.zig
wemeetagain: State is static and reusable across a later 0→1 environment transition, so poisoning the whole State would make init read an undefined initialized flag. PubkeyCache.deinit already poisons the freed cache storage. Keeping the State fields defined is intentional.

bindings/napi/pubkeys.zig
wemeetagain: Agreed that parent-directory fsync is required for crash-durable rename. This save path only promises flushed file contents plus atomic destination replacement; PKIX is a rebuildable cache and load failure falls back to rebuilding. Zig Io currently has no portable Dir sync, so adding platform-specific directory fsync here is outside the intended durability contract.

bindings/napi/pubkeys.zig
wemeetagain: StackFallbackAllocator would reproduce the same threshold/fallback policy through a stateful allocator. The current typed stack slice and explicit fallback allocation are simpler and avoid allocator indirection, so I would keep the existing code.

src/state_transition/cache/pkix.zig
wemeetagain: I don't think its necessary considering we already cover the structural compatibility boundary really well: arch, endianness, zig version, blst point/field/limb sizes, alignments, and offsets. The residual risk would be that blst changes its internal limb encoding while preserving the exact C layout. Theoretically possible, but imo not really practically possible. Would be a huge rewrite, and blst has been historically very stable.

Since blst is pinned, and upgrades to blst are explicit, I would just bump the pkix version along with that upgrade at that point.

GrapeBaBa (approved):

spiral-ladder (approved):
i think this looks good, i feel like some of the doc comments are a lot more verbose than necessary but we can fix that after

nazarhussain (changes requested):
The design presented in this PR relies on premise that `index→pubkey` history is an **unforkable singleton**.

In post-Electra chain, `pending_deposits` ordering depends on which `deposit requests` each block includes, so two fork branches can assign the same validator index to different pubkeys.

The old `epoch_cache.addPubkey` silently overwrote in that case; the new append returns `ConflictingPubkey` and `syncPubkeys` returns `DuplicatePubkey` — aborting block processing on a valid competing branch until restart. 

If our intention is that we only ever feed finalized registries into this cache, then we should document this in the PR somehow, if that's not our intention then it's a blocker.

src/state_transition/cache/pkix.zig
```diff
@@ -0,0 +1,262 @@
...
+/// PKIX checks framing, ABI compatibility, bounds, and checksum, but does not
+/// authenticate or semantically validate entries. Callers must provide an
+/// application-owned file from the intended network.
+pub fn load(
+    allocator: std.mem.Allocator,
+    io: std.Io,
+    reader: *std.Io.Reader,
+    file_size: u64,
+    max_capacity: usize,
+) !PubkeyCache {
```
nazarhussain, line 188: This never checks that the compressed keys in the payload are unique. `reIndexContext` happily indexes duplicates, and lookups then collapse to the lowest index. 

We might verify `entries.count() == header.entry_count` after ` reIndexContext` to validate this.

bindings/src/index.d.ts
nazarhussain, line 392: This export still omits `pubkeys`, is it intentional?

nazarhussain (changes requested):

bindings/napi/pubkeys.zig
```diff
@@ -1,174 +1,106 @@
...
-    pub fn init(self: *State) !void {
+    pub fn init(self: *State, env: napi.Env) !void {
         if (self.initialized) return;
-        self.pubkey2index = PubkeyIndexMap.init(cache_allocator);
-        try self.pubkey2index.ensureTotalCapacity(default_initial_capacity);
-        self.index2pubkey = try Index2PubkeyCache.initCapacity(cache_allocator, default_initial_capacity);
+
+        self.cache = try PubkeyCache.initCapacity(allocator, js.io(), default_initial_capacity);
         self.initialized = true;
+        self.control_env = env.env;
```
nazarhussain, line 26: This captures the the first env in which native binding is loaded and then only clear in the `State.deinit`. 

If the binding is loaded in a worker and then for some reason worker exits then `control_env` points to a destroyed env. And any subsequent loads will cause `PubkeyCacheControlEnvironmentOnly`. 

We should clear this in `control_env` own's tear down not the `State.deinit`

nazarhussain (changes requested):

twoeths:
> The design presented in this PR relies on premise that `index→pubkey` history is an **unforkable singleton**.
> 
> In post-Electra chain, `pending_deposits` ordering depends on which `deposit requests` each block includes, so two fork branches can assign the same validator index to different pubkeys.


we only add validator to registry if its deposit slot is finalized https://github.com/ChainSafe/lodestar/blob/d96240a05e6c0694e84997bbe8ddd25100129ef6/packages/state-transition/src/epoch/processPendingDeposits.ts#L52

in forking condition, each block can has its own view of pending_deposits. Only pending validators on the same branch of one finalized state gets added, the other ones are orphaned

so there should not be the case where one index map to 2 pubkeys

spiral-ladder:
One thing i noticed is missing is `syncPubkeys` as a binding, we can add that in a followup PR to this

wemeetagain:
> The design presented in this PR relies on premise that index→pubkey history is an unforkable singleton.

Yes this is correct and is also how it works in lodestar

bindings/src/index.d.ts
wemeetagain: Yes it was intentional. It seemed that our way of using the bindings are subpath imports, so I didn't bother with this.

bindings/napi/pubkeys.zig
wemeetagain: > This captures the the first env in which native binding is loaded and then only clear in the State.deinit.
> If the binding is loaded in a worker...

Yes this is a fine imo. 

> We should clear this in control_env own's tear down not the State.deinit

I disagree with that. I think we should be designing for our usecase and not adding useless machinery.
Our usecase:
- Startup:
  - The control worker (main thread) boots
  - The control worker boots auxiliary workers
- Runtime:
  - no workers started or stopped
- Shutdown:
  - The control worker shuts down the auxiliary workers
  - The control worker (main thread) shuts down

src/state_transition/cache/pkix.zig
wemeetagain: I don't think that works for detecting duplicates.
The tradeoffs for this pkix format are that it intentionally trusts application-owned files and skips semantic validation for load speed. (We should update the documentation if this is not clear enough, but I think it is actually fine as-is)

spiral-ladder:
opened https://github.com/ChainSafe/lodestar-z/pull/537 as a followup PR

nazarhussain (approved):

PR #525, opened 2026-07-22
fix(bls): reject unknown BLST error codes

GrapeBaBa:
## Motivation

`errorFromInt` currently treats every unrecognized BLST status code as success. This can silently accept unexpected or newly introduced BLST errors.

## Description

- Handle `BLST_SUCCESS` explicitly as the only successful status.
- Return `BlstError.UnknownError` for unrecognized status codes.
- Remove the unused reverse conversion that fabricated a BLST status for `MergeError`.
- Add coverage for success, every known BLST error, and an unknown status code.
- Validate with the BLS test suite in Debug and ReleaseSafe, tracked Zig formatting checks, and `pnpm lint`.

The full `zig build test` command was attempted but exceeded 15 minutes while running the generated spec suites.

AI assistance: this implementation was primarily AI-authored.

### DIFF
--- a/src/bls/error.zig
+++ b/src/bls/error.zig
@@ -9,23 +9,12 @@ pub const BlstError = error{
     PkIsInfinity,
     BadScalar,
     MergeError,
+    UnknownError,
 };
-
-pub fn intFromError(e: BlstError) c_uint {
-    return switch (e) {
-        BlstError.BadEncoding => c.BLST_BAD_ENCODING,
-        BlstError.PointNotOnCurve => c.BLST_POINT_NOT_ON_CURVE,
-        BlstError.PointNotInGroup => c.BLST_POINT_NOT_IN_GROUP,
-        BlstError.AggrTypeMismatch => c.BLST_AGGR_TYPE_MISMATCH,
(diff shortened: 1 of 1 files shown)


src/bls/error.zig
```diff
@@ -9,30 +9,44 @@ pub const BlstError = error{
...
         c.BLST_PK_IS_INFINITY => return BlstError.PkIsInfinity,
         c.BLST_BAD_SCALAR => return BlstError.BadScalar,
-        else => return,
+        else => return BlstError.UnknownError,
     }
 }
+
+const std = @import("std");
+
+test "convert BLST error codes" {
```
spiral-ladder: im not sure if we need this naive test, doesnt seem to add much value

spiral-ladder (dismissed):
good catch, ideally `UnknownError` should never be hit but this helps to be 100% sure

src/bls/error.zig
GrapeBaBa: Removed the test entirely as suggested.

wemeetagain (approved):

spiral-ladder (approved):

PR #527, opened 2026-07-25
fix(bindings): clean up failed async BLS work

GrapeBaBa:
## Summary

- delete unqueued N-API async work when Promise creation or queueing fails
- settle the unreturned Promise on setup failures so Node can release its deferred handle
- transfer work/data ownership to the completion callback only after queueing succeeds

## Why

Previously, a successful `napi_create_async_work` followed by a failed `napi_queue_async_work` destroyed the callback data without deleting the work handle. This leaked the native handle and left it pointing at freed data.


bindings/napi/blst.zig
```diff
@@ -859,8 +861,19 @@ pub fn asyncAggregateWithRandomness(sets: js.Array) !js.Value {
...
         data,
     );
+    errdefer work.delete() catch |err| {
+        std.log.err("failed to delete unqueued async BLS work: {s}", .{@errorName(err)});
+    };
+
     data.work = work.work;
 
+    // Settle the unreturned Promise so Node can release its deferred handle.
+    const deferred_cleanup_value = try env.getUndefined();
```
nazarhussain: One case left, what if `env.getUndefined()` failed?

GrapeBaBa: Addressed in 2cd9686e. `getUndefined()` now runs before `createAsyncWork()`, so if it fails only the outer `data` cleanup is active and no async work handle needs rollback.

nazarhussain (approved):

spiral-ladder (approved):

PR #528, opened 2026-07-25
fix(bindings): harden random aggregate scalars

GrapeBaBa:
## Summary

- generate aggregate and batch-verification coefficients directly from the shared CSPRNG
- reject zero 64-bit coefficients in synchronous aggregation, asynchronous aggregation, and `verifyMultipleAggregateSignatures`
- bound zero-coefficient regeneration to eight retries and return `RandomScalarGenerationFailed` if the random source remains unusable

## Why

A zero coefficient omits its public key and signature from the random linear combination. The original `@chainsafe/blst` binding generates nonzero 64-bit coefficients in both synchronous and asynchronous aggregation.

The synchronous aggregation and batch-verification paths also seeded `DefaultPrng` once with 64 bits and derived every coefficient from that non-cryptographic PRNG. They now bulk-fill coefficient storage directly with `js.io().random`, consistent with the CSPRNG already used by the asynchronous path.

Random rejection sampling is bounded so a broken random source fails instead of spinning forever. The asynchronous and batch-verification storage keeps a 32-byte stride per input, while the BLS operations read the first 8 bytes as the coefficient.


bindings/napi/blst.zig
```diff
@@ -833,7 +828,17 @@ pub fn asyncAggregateWithRandomness(sets: js.Array) !js.Value {
     data.err = null;
     data.deferred = undefined;
     data.work = undefined;
-    napi_io.get().random(data.randomness[0 .. n * 32]);
+
+    const io = napi_io.get();
```
nazarhussain: This need to be replaced with `js.io()` and had merge conflicts because of it.

bindings/napi/blst.zig
nazarhussain, line 821: We need to apply same optimization in `verifyMultipleAggregateSignatures`

bindings/napi/blst.zig
```diff
@@ -833,7 +828,17 @@ pub fn asyncAggregateWithRandomness(sets: js.Array) !js.Value {
...
+    io.random(data.randomness[0 .. n * 32]);
+
+    // A zero scalar would omit its input from the random linear combination.
+    for (0..n) |i| {
+        const scalar = data.randomness[i * 32 ..][0..8];
+        while (std.mem.allEqual(u8, scalar, 0)) {
+            io.random(scalar);
+        }
```
nazarhussain: Hope that unbounded loop would not be a problem.

bindings/napi/blst.zig
GrapeBaBa: Merged current main and switched the affected paths to `js.io()`.

bindings/napi/blst.zig
GrapeBaBa: Applied the same bulk CSPRNG generation and nonzero coefficient check to `verifyMultipleAggregateSignatures`.

bindings/napi/blst.zig
GrapeBaBa: Bounded regeneration to eight retries. Exhaustion now returns `RandomScalarGenerationFailed` instead of looping forever.

nazarhussain (approved):

spiral-ladder (approved):

PR #529, opened 2026-07-25
fix(bindings): free metrics writer on scrape failure

GrapeBaBa:
## Summary

- deinitialize the allocating metrics writer on every exit path
- preserve the existing `ArrayList` ownership transfer and cleanup

## Why

`state_transition.metrics.write()` can fail after `std.Io.Writer.Allocating` has grown its buffer. Previously that error returned before either the writer or an owning `ArrayList` was deinitialized, leaking the buffer.

The unconditional `defer aw.deinit()` now releases the buffer when writing fails. On success, `toArrayList()` transfers ownership and resets the writer to an empty state, so the writer defer is safe while the list remains responsible for the transferred allocation.


bindings/napi/metrics.zig
```diff
@@ -21,7 +21,10 @@ pub fn init() !void {
 /// JS: metrics.scrapeMetrics() → string
 pub fn scrapeMetrics() !js.String {
     var aw: std.Io.Writer.Allocating = .init(allocator);
+    defer aw.deinit();
+
     try state_transition.metrics.write(&aw.writer);
+
     var list = aw.toArrayList();
     defer list.deinit(allocator);
     return js.String.from(list.items);
```
nazarhussain:
```suggestion
var aw: std.Io.Writer.Allocating = .init(allocator);
defer aw.deinit();

try state_transition.metrics.write(&aw.writer);
return js.String.from(aw.written());
```

nazarhussain (changes requested):

bindings/napi/metrics.zig
nazarhussain: `aw.written()` already returns the byte slice, so there's no need to extract an `ArrayList` just to read .items off it.

nazarhussain (approved):

spiral-ladder (approved):

PR #531, opened 2026-07-27
fix(bls): size pairing buffers for 32-bit targets

GrapeBaBa:
## Summary

- compute the 384-bit BLST field size independently of pointer width
- keep pointer-width sizing only for the `DST` and `DST_len` fields

## Why

`Pairing.sizeOf()` divided a 384-bit field width by `@sizeOf(usize)`. This happened to produce the correct 48-byte field size on 64-bit targets, but produced 96 bytes on 32-bit targets. BLST represents the field as either six 64-bit limbs or twelve 32-bit limbs, so its size remains 48 bytes.

The old calculation therefore reserved a 6352-byte pairing context on 32-bit targets instead of BLST's 3184 bytes. This was safe but doubled every pairing buffer, including the worker pool's buffer array.

This was identified while discussing 32-bit capacity handling in https://github.com/ChainSafe/lodestar-z/pull/522#discussion_r3640986862.

### DIFF
--- a/src/bls/Pairing.zig
+++ b/src/bls/Pairing.zig
@@ -3,8 +3,8 @@ ctx: *c.blst_pairing,
 
 const Self = @This();
 
-/// Required alignment for the pairing buffer. BLST fields require at most
-/// 8-byte alignment.
+/// Required alignment for the pairing buffer. The opaque C struct contains
+/// `uptr_t` (64-bit) fields that require 8-byte alignment.
 pub const buf_align = 8;
 
 /// Initializes a pairing context with the provided `buffer` and other parameters.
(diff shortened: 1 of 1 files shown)


spiral-ladder:
I don't think we even build for and support any 32-bit targets, not sure if this is entirely necessary?

GrapeBaBa:
> I don't think we even build for and support any 32-bit targets, not sure if this is entirely necessary?

Yes, so I asked in #522, but the `vec384_size` should be always 384 / 8 whatever 64/32 bit target

wemeetagain (approved):

spiral-ladder (approved):

PR #537, opened 2026-07-30
feat(bindings): add pubkey cache syncPubkeys

spiral-ladder:
needed [here](https://github.com/ChainSafe/lodestar/blob/ba5ee9aff0173134c0da82595326b39a9660debe/packages/state-transition/src/cache/pubkeyCache.ts#L69) which we do not expose atm in #522

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -6,6 +6,8 @@ const blst_bindings = @import("./blst.zig");
 const state_transition = @import("state_transition");
 const PubkeyCache = state_transition.PubkeyCache;
 const pkix = state_transition.pkix;
+const NativeValidator = @import("consensus_types").phase0.Validator.Type;
+const Validator = js.Object(struct { pubkey: js.Uint8Array });
 
 /// Uses the page allocator for the process-wide cache's internal allocations.
 const allocator = std.heap.page_allocator;
@@ -181,6 +183,40 @@ pub fn append(index: js.Number, pubkey: js.Uint8Array) !void {
     const pubkey_bytes = pubkey_slice[0..48].*;
 
     try state.cache.append(io, pubkey_bytes, idx);
+}
+
+/// JS: pubkeys.syncPubkeys(validators)
+pub fn syncPubkeys(validators: js.Array) !void {
+    if (!state.initialized) return error.PubkeyIndexNotInitialized;
+
+    const validator_count = try validators.length();
+    const io = js.io();
+    const cached_count = state.cache.count(io);
+    if (validator_count <= cached_count) return;
+
+    const missing_count = validator_count - cached_count;
+    const native_validators = try allocator.alloc(NativeValidator, missing_count);
+    defer allocator.free(native_validators);
+    const validator_ptrs = try allocator.alloc(*const NativeValidator, validator_count);
+    defer allocator.free(validator_ptrs);
+
+    // These are ignored by the native syncPubkeys. We skip work by
+    // just filling with dummy pointers
+    @memset(validator_ptrs[0..cached_count], &native_validators[0]);
+    for (cached_count..validator_count) |index| {
+        const value = try validators.get(@intCast(index));
+        try Validator.validateArg(value.val);
+        const validator = try (Validator{ .val = value.val }).get();
+        const pubkey = try validator.pubkey.toSlice();
+        if (pubkey.len != blst_bindings.PublicKey.COMPRESS_SIZE) return error.InvalidPubkeyLength;
+
+        const missing_index = index - cached_count;
+        native_validators[missing_index] = undefined;
+        native_validators[missing_index].pubkey = pubkey[0..48].*;
+        validator_ptrs[index] = &native_validators[missing_index];
+    }
+
+    try state.cache.syncPubkeys(io, validator_ptrs);
 }
 
 /// JS: pubkeys.size() → number
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -7,7 +7,6 @@ const state_transition = @import("state_transition");
 const PubkeyCache = state_transition.PubkeyCache;
 const pkix = state_transition.pkix;
 const NativeValidator = @import("consensus_types").phase0.Validator.Type;
-const Validator = js.Object(struct { pubkey: js.Uint8Array });
 
 /// Uses the page allocator for the process-wide cache's internal allocations.
 const allocator = std.heap.page_allocator;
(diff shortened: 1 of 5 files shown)


bindings/napi/pubkeys.zig
```diff
@@ -183,6 +185,40 @@ pub fn append(index: js.Number, pubkey: js.Uint8Array) !void {
...
+
+    const missing_count = validator_count - cached_count;
+    const native_validators = try allocator.alloc(NativeValidator, missing_count);
+    defer allocator.free(native_validators);
+    const validator_ptrs = try allocator.alloc(*const NativeValidator, validator_count);
+    defer allocator.free(validator_ptrs);
+
+    // These are ignored by the native syncPubkeys. We skip work by
+    // just filling with dummy pointers
+    @memset(validator_ptrs[0..cached_count], &native_validators[0]);
```
wemeetagain: Is this actually required? maybe better to leave as undefined?

bindings/napi/pubkeys.zig
```diff
@@ -183,6 +185,40 @@ pub fn append(index: js.Number, pubkey: js.Uint8Array) !void {
...
+    for (cached_count..validator_count) |index| {
+        const value = try validators.get(@intCast(index));
+        try Validator.validateArg(value.val);
+        const validator = try (Validator{ .val = value.val }).get();
```
wemeetagain: nit: this seems kludgey, somehow I would expect a way to unwrap and validate this easier? Eg: `try Validator.initFromValue(value)`

bindings/napi/pubkeys.zig
```diff
@@ -183,6 +185,40 @@ pub fn append(index: js.Number, pubkey: js.Uint8Array) !void {
...
+        const pubkey = try validator.pubkey.toSlice();
+        if (pubkey.len != blst_bindings.PublicKey.COMPRESS_SIZE) return error.InvalidPubkeyLength;
+
+        const missing_index = index - cached_count;
+        native_validators[missing_index] = undefined;
+        native_validators[missing_index].pubkey = pubkey[0..48].*;
```
wemeetagain: nit: use of both 48 and PublicKey.COMPRESS_SIZE. Best to pick one and stick with it

bindings/napi/pubkeys.zig
```diff
@@ -183,6 +185,40 @@ pub fn append(index: js.Number, pubkey: js.Uint8Array) !void {
...
+    defer allocator.free(validator_ptrs);
+
+    // These are ignored by the native syncPubkeys. We skip work by
+    // just filling with dummy pointers
+    @memset(validator_ptrs[0..cached_count], &native_validators[0]);
+    for (cached_count..validator_count) |index| {
+        const value = try validators.get(@intCast(index));
+        try Validator.validateArg(value.val);
+        const validator = try (Validator{ .val = value.val }).get();
+        const pubkey = try validator.pubkey.toSlice();
```
wemeetagain: would be cool to see a `toArray(comptime len): ![n]u8`
which could be used directly below

bindings/napi/pubkeys.zig
```diff
@@ -6,6 +6,8 @@ const blst_bindings = @import("./blst.zig");
 const state_transition = @import("state_transition");
 const PubkeyCache = state_transition.PubkeyCache;
 const pkix = state_transition.pkix;
+const NativeValidator = @import("consensus_types").phase0.Validator.Type;
+const Validator = js.Object(struct { pubkey: js.Uint8Array });
```
wemeetagain: nit: this could be defined inline for better readability imo. Its a bit of a strange type, since it really is quite specific to syncing pubkeys.

bindings/napi/pubkeys.zig
```diff
@@ -183,6 +185,40 @@ pub fn append(index: js.Number, pubkey: js.Uint8Array) !void {
...
+    // just filling with dummy pointers
+    @memset(validator_ptrs[0..cached_count], &native_validators[0]);
+    for (cached_count..validator_count) |index| {
+        const value = try validators.get(@intCast(index));
+        try Validator.validateArg(value.val);
+        const validator = try (Validator{ .val = value.val }).get();
+        const pubkey = try validator.pubkey.toSlice();
+        if (pubkey.len != blst_bindings.PublicKey.COMPRESS_SIZE) return error.InvalidPubkeyLength;
+
+        const missing_index = index - cached_count;
```
wemeetagain: nit: naming is a little confused imo. Preference for some more discipline around the naming the data, operations, how they relate, etc in this PR.
Eg: validators, indices, counts, ptrs, native, missing, cached. These are all names and adjectives introduced here. But somehow it feels unwieldy. Count vs index hygiene (classic off-by-one potential), native vs missing (two seemingly related concepts?), validator describing both native and js representations (eg `validators` vs `validator_ptrs`). Not prescriptive in how to handle this, just saying that it feels slightly off.

bindings/napi/pubkeys.zig
spiral-ladder: possible; i was also thinking if maybe `syncPubkeys` should just accept a length of the missing indices (the suffix) instead too because we don't do anything with those too

bindings/napi/pubkeys.zig
spiral-ladder: I don't think we have `initFromValue`, but we can use `asObject` though (but it will look quite similar)

bindings/napi/pubkeys.zig
spiral-ladder: I think we can leave it undefined with a safety note but we might also want to consider changing `syncPubkeys` to accept only the list of new incoming pubkeys + the total validator len. Thta way we don't have to mess with the suffix logic, deal with missing ptrs, etc.

what do you think?

bindings/napi/pubkeys.zig
spiral-ladder: i did a little tweaking in [9c7eb81](https://github.com/ChainSafe/lodestar-z/pull/537/commits/9c7eb8172afb0160843da0c94648121be3262538); specifically, `num_new_validators` and `new_validators` for more context, removed some of the necessary counts

bindings/napi/pubkeys.zig
spiral-ladder: updated in [a5dafb7](https://github.com/ChainSafe/lodestar-z/pull/537/commits/a5dafb73ac18b4f4175530ee8a774df3e9e99e56)

bindings/napi/pubkeys.zig
spiral-ladder: Let's create an issue in zapi for that

spiral-ladder:
trying out stacked PRs

bindings/napi/pubkeys.zig
spiral-ladder: makring this as ready for review, let me know if the naming is better/clearer, i agree it was a bit clumsy before and confusing

spiral-ladder:
had to force push bc the commit history on top of #522 was squashed and merged and this pr after changing its based to main inherited all the old commits, trying out stacked prs was a mistake

wemeetagain (approved):
looks good :)

PR #540, opened 2026-08-06
build(bls): migrate deprecated @cImport to addTranslateC

### LINKED ISSUE #530: build(bls): migrate deprecated @cImport to addTranslateC
## Problem

`src/bls/root.zig` imports `blst.h` with `@cImport`:

```zig
pub const c = @cImport({
    @cInclude("blst.h");
});
```

`@cImport` is deprecated in Zig 0.16. The BLS bindings should instead translate the header through the build system and import the generated Zig module.

## Proposed approach

- Add a small wrapper header that includes `blst.h`.
- Retain the result of `zbuild.configureBuild` in `build.zig`.
- Create a `std.Build.addTranslateC` step using the BLS module's target and optimization mode.
- Add the `blst` artifact's emitted include tree to the translation step.
- Inject the generated module into the BLS module and replace `@cImport` with `@import`.

`zbuild` 0.4.0 does not currently expose this through the declarative module configuration, so the generated module will likely need to be wired in after `configureBuild`.

## Acceptance criteria

- No `@cImport` remains in `src/bls/root.zig`.
- BLS unit tests and N-API bindings build successfully in Debug and ReleaseSafe modes.
- Header translation works for the project's supported targets, especially macOS/Linux on x86_64 and aarch64.

bomanaps:
Wire blst.h translation through an addTranslateC step after zbuild.configureBuild, replacing the now-deprecated @cImport  closes #530.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -3,5 +3,15 @@ const zbuild = @import("zbuild");
 
 pub fn build(b: *std.Build) !void {
     @setEvalBranchQuota(200_000);
-    _ = try zbuild.configureBuild(b, @import("build.zig.zon"), .{});
+    const zbuild_result = try zbuild.configureBuild(b, @import("build.zig.zon"), .{});
+
+    const bls_module = zbuild_result.module("bls").?;
+    const blst_artifact = zbuild_result.dependency("blst").?.artifact("blst");
+    const translate_blst = b.addTranslateC(.{
+        .root_source_file = b.path("src/bls/blst_c.h"),
+        .target = bls_module.resolved_target.?,
+        .optimize = bls_module.optimize.?,
+    });
+    translate_blst.addIncludePath(blst_artifact.getEmittedIncludeTree());
+    bls_module.addImport("blst_c", translate_blst.createModule());
 }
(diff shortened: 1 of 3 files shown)


spiral-ladder:
I think this should belong in [zbuild](https://github.com/ChainSafe/zbuild), not here, through adding a generic translate c support

bomanaps:
> I think this should belong in [zbuild](https://github.com/ChainSafe/zbuild), not here, through adding a generic translate c support

Should generic translate-c support land as a top-level .translate_c_modules section importable by any module, or have it as a nested .translate_c field on individual module declarations?

wemeetagain:
imo this should be handled by the bls repo (or another repo).

I don't know how zbuild can do this in a sane way, eg how can it create the proper header file?

bomanaps:
> hould generic translate-c support land as a top-level .translate_c_modules section importable by any module, or have it as a nested .translate_c field on individual module declarations?

 If we go this route, should blst.zig run translate-c against its own headers and export the result as a named module e.g. blst_c, so we have lodestar-z just imports blst:blst_c via zbuild's existing dependency module syntax and drops the wrapper header + build.zig wiring here entirely?

wemeetagain:
- closing in favor of #554

PR #541, opened 2026-08-07
fix: bound beacon runtime and sync work

lodekeeper-z:
## Summary
- subscribe to data-column subnets derived from the node's actual custody columns
- bound req/resp protocol opens and the complete response transfer so slow peers cannot monopolize the P2P loop
- apply BLS dispatch gates to recovered unknown-block gossip work
- preserve peer ownership across in-flight sync downloads and keep validation/rewind paths within the pending batch window
- exclude the temporary QUIC/gossipsub tracing dependency, debug metrics, trace scripts, and investigation-only artifacts

## Review notes
- Response timeouts use one absolute deadline for the complete multi-chunk transfer, matching Lodestar-TS `respTimeoutMs` semantics rather than resetting per chunk.
- Sync retry dispatch remains capped after a long awaiting-validation prefix is rewound.
- `Batch.startDownload` is transactional under allocator failure: no state, generation, metric, or callback changes occur unless peer ownership succeeds.

### DIFF
--- a/src/node/p2p_runtime.zig
+++ b/src/node/p2p_runtime.zig
@@ -28,6 +28,26 @@ const ReqRespContext = networking.ReqRespContext;
 const ConnectionDirection = networking.ConnectionDirection;
 const GoodbyeReason = networking.GoodbyeReason;
 const PeerAction = networking.PeerAction;
+test "data column subnet subscriptions use actual custody columns" {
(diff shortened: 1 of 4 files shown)


spiral-ladder:
you're alive!

lodekeeper-z:
Alive and catching up 😄

PR #544, opened 2026-08-07
chore: avoid slow tests in AGENTS.md

wemeetagain:
We shouldn't require running the full spec test suite locally for every PR.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -284,7 +284,7 @@ implementation was primarily AI-authored or whether AI was used only for codebas
 ### Pre-push checklist
 
 1. `zig fmt --check .`
-2. `zig build test`
+2. Run relevant tests, avoid slow `zig build test` unless necessary
 3. Relevant spec tests for consensus, SSZ, or BLS changes
 4. `pnpm lint` for binding source changes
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -284,7 +284,7 @@ implementation was primarily AI-authored or whether AI was used only for codebas
 ### Pre-push checklist
 
 1. `zig fmt --check .`
-2. `zig build test`
+2. Run relevant tests, avoid slow `zig build test` unless necessary
```
spiral-ladder, line 287: maybe we can be more specific? Not sure about exact wording, but i prefer stronger words like 'never'

```suggestion
2. Run relevant tests, never run slow `zig build test` unless changes touch spec logic
```

nflaig (approved):

spiral-ladder (approved):
lgtm with an optional comment, not sure which one is better, we can test things out

wemeetagain:
@nflaig @spiral-ladder can one of you take over / reopen the PR? I don't have my gpg key on this machine?

wemeetagain:
closing in favor of #546

PR #545, opened 2026-08-07
fix(bls): enforce 32-byte signing roots

### LINKED ISSUE #523: Remain arbitrary-length BLS messages in Zig and NAPI bindings
### Description                                                                                           
                                                                                                           
 BLST supports arbitrary-length messages, and Lodestar’s previous @chainsafe/blst Rust binding exposed     
 this capability without enforcing a 32-byte length.                                                       
                                                                                                           
 lodestar-z currently narrows this behavior:                                                               
                                                                                                           
 - Pairing.mulAndAggregate always passes 32 instead of msg.len.                                            
 - Some aggregate verification APIs use fixed [32]u8 messages.                                             
 - The NAPI binding rejects messages that are not exactly 32 bytes.                                        
                                                                                                           
 Both src/bls and bindings/napi should preserve the underlying BLST capability. Ethereum’s 32-byte         
 signing-root requirement should be enforced by the consensus or application layer.                        
                                                                                                           
 Proposed changes(nice to have):                                                                                         
                                                                                                           
 - Pass the actual message length to BLST.                                                                 
 - Accept arbitrary-length messages across Zig and NAPI verification APIs.

lodekeeper-z:
## Summary

- define one canonical `bls.SigningRoot = [32]u8` and use it throughout native BLS signing, verification, pairing, randomized batch verification, thread-pool, and state-transition APIs
- validate every dynamic N-API signing-root input with one exact-length helper and copy batch roots into native storage before worker-thread verification
- document the 32-byte TypeScript contract and add 31-/33-byte rejection coverage for every exported N-API signing and verification path
- make the BLS spec harness reject decoded messages that are not exactly one signing root

Closes #523

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -117,12 +117,6 @@ fn uint8SliceFromValue(value: js.Value) ![]u8 {
     return info.data;
 }
 
-/// Validates a dynamic N-API byte slice and narrows it to the fixed BLS message type.
-fn signingRootFromBytes(bytes: []const u8) !*const SigningRoot {
-    if (bytes.len != @sizeOf(SigningRoot)) return error.InvalidMessageLength;
-    return bytes[0..@sizeOf(SigningRoot)];
-}
-
 pub const PublicKey = struct {
     pub const js_meta = js.class(.{});
 
@@ -335,8 +329,9 @@ pub const SecretKey = struct {
 
     /// Signs a message with this `SecretKey`, returns a `Signature`.
     pub fn sign(self: *const SecretKey, msg: js.Uint8Array) !Signature {
-        const signing_root = try signingRootFromBytes(try msg.toSlice());
-        return .{ .raw = self.raw.sign(signing_root, DST, null) };
+        const msg_bytes = try msg.toSlice();
+        if (msg_bytes.len != @sizeOf(SigningRoot)) return error.InvalidMessageLength;
+        return .{ .raw = self.raw.sign(msg_bytes[0..@sizeOf(SigningRoot)], DST, null) };
     }
 
     /// Derives the PublicKey from this SecretKey.
@@ -366,11 +361,12 @@ pub const SecretKey = struct {
 /// 4) pk_validate: ?bool
 /// 5) sig_groupcheck: ?bool
 pub fn verify(msg: js.Uint8Array, pk: PublicKey, sig: Signature, pk_validate: ?js.Boolean, sig_groupcheck: ?js.Boolean) !js.Boolean {
-    const signing_root = try signingRootFromBytes(try msg.toSlice());
+    const msg_bytes = try msg.toSlice();
+    if (msg_bytes.len != @sizeOf(SigningRoot)) return error.InvalidMessageLength;
 
     sig.raw.verify(
         try boolOrDefault(sig_groupcheck, false),
-        signing_root,
+        msg_bytes[0..@sizeOf(SigningRoot)],
         DST,
         null,
         &pk.raw,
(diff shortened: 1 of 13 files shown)


wemeetagain (dismissed):

GrapeBaBa:
This PR seems go to another direction to limit the message size in the bls layer not consensus layer, though in Ethereum context both are OK.

lodekeeper-z:
Yes — this intentionally chooses the opposite layering from the issue’s proposed arbitrary-message API. The rationale is that every Lodestar caller here passes an Ethereum consensus signing root, so `SigningRoot = [32]u8` makes the native boundary compile-time safe while N-API rejects malformed dynamic input consistently.

If you want `src/bls`/N-API to remain a generic BLST wrapper, I can rework this to retain `[]const u8` there and enforce 32 bytes only at the consensus/application boundary. Which layering do you prefer for this package?

wemeetagain:
Yes I think with our opinionated Pubkey and Signature structs and our use of ethereum's DST, it makes more sense to stick with the more ethereum-specific direction at the bls layer, and only allow 32 byte msgs.

bindings/napi/blst.zig
```diff
@@ -116,6 +117,12 @@ fn uint8SliceFromValue(value: js.Value) ![]u8 {
     return info.data;
 }
 
+/// Validates a dynamic N-API byte slice and narrows it to the fixed BLS message type.
+fn signingRootFromBytes(bytes: []const u8) !*const SigningRoot {
+    if (bytes.len != @sizeOf(SigningRoot)) return error.InvalidMessageLength;
+    return @ptrCast(bytes.ptr);
```
GrapeBaBa:
```suggestion
    return bytes[0..@sizeOf(SigningRoot)];
```
no need cast

lodekeeper-z: Applied in 84c40ece. The length check still narrows the slice safely, without the pointer cast. Verified with the ReleaseSafe N-API build, all 161 BLS binding tests, `zig build test:bls`, Zig format on the changed file, and `pnpm lint`. The full `zig build test` remains blocked locally by unrelated io_uring `ArgumentsInvalid` failures in clock tests.

src/bls/Signature.zig
```diff
@@ -117,7 +117,7 @@ pub fn fastAggregateVerify(
     return try self.aggregateVerify(
         sig_groupcheck,
         buffer,
-        @ptrCast(msg),
+        @as([*]const SigningRoot, @ptrCast(msg))[0..1],
```
GrapeBaBa: why this need change?

lodekeeper-z: It does not strictly need to change. I made the one-element slice conversion explicit while replacing `[32]u8` with the `SigningRoot` alias, but the `aggregateVerify` parameter still gives `@ptrCast(msg)` enough context to infer the same slice, and I confirmed that form compiles as a one-element `[]const SigningRoot`. So this is an unnecessary verbosity change rather than a behavioral requirement.

GrapeBaBa: just keep original version

lodekeeper-z: Restored the original inferred slice form in bf2fb97d. Verified with `zig fmt --check src/bls/Signature.zig` and `zig build test:bls`.

bindings/napi/blst.zig
```diff
@@ -116,6 +117,12 @@ fn uint8SliceFromValue(value: js.Value) ![]u8 {
     return info.data;
 }
 
+/// Validates a dynamic N-API byte slice and narrows it to the fixed BLS message type.
+fn signingRootFromBytes(bytes: []const u8) !*const SigningRoot {
+    if (bytes.len != @sizeOf(SigningRoot)) return error.InvalidMessageLength;
+    return bytes[0..@sizeOf(SigningRoot)];
```
spiral-ladder: Can we just inline these? Seems unnecessary to have a separate 2 line function when its only consumer is within the same file

lodekeeper-z: Inlined the signing-root length checks and narrowing at each N-API call site in 520f8528, and removed the helper. Verified with the ReleaseSafe bindings build, all 161 BLS binding tests, `zig build test:bls`, Zig format, and `pnpm lint`.

spiral-ladder (dismissed):

lodekeeper-z:
Synced this branch with current `main` after #547 merged and resolved the overlapping batch-verification changes by keeping the structural `BatchVerifyItem` API with a fixed-size `SigningRoot` message. Verified `zig build test:bls` in Debug and ReleaseSafe (14/14), ReleaseSafe bindings build (10/10), all 161 BLS binding tests, Zig format on the resolved files, and `pnpm lint` (existing warnings only). The full binding run reached 192 passing tests but remains blocked by the absent ERA fixture for the unrelated state-view/teardown tests.

wemeetagain (approved):

PR #547, opened 2026-08-07
fix(bls): make batch cardinality structural

### LINKED ISSUE #543: fix(bls): validate batch verifier slice cardinalities
## Problem

`src/bls/fast_verify.zig:12-48` accepts `n_elems` independently from `msgs`, `pks`, `sigs`, and `rands`, but only rejects zero. The loop indexes all four slices through `n_elems` without validating their cardinalities.

A local probe at `542779fc` with `n_elems = 1` and empty slices produced:

```
index out of bounds: index 0, len 0
```

at `fast_verify.zig:35`.

The threaded implementation already validates every slice length at `src/bls/ThreadPool.zig:268-273`, showing the intended contract.

## Impact

Malformed exported Zig API inputs panic in safety-enabled builds. In unchecked builds, invalid pointers may be passed into BLST.

The current NAPI wrapper constructs matching arrays, so ordinary JavaScript callers cannot directly trigger this. This concerns the exported native API.

## Expected behavior

Reject inconsistent cardinalities with `BlstError.VerifyFail` before indexing any input.

## Acceptance criteria

- Validate `msgs.len`, `pks.len`, `sigs.len`, and `rands.len` against `n_elems`.
- Add regression tests mismatching each slice independently.
- Confirm `zig build test:bls` and ReleaseSafe tests pass.

Found during the BLS security/correctness sweep of `542779fc`.

lodekeeper-z:
## Summary

- replace the batch verifier's independent count and parallel message/public-key/signature/randomness slices with one `BatchVerifyItem` slice
- use the same item representation in the single-threaded verifier, thread-pool jobs, and N-API adapter
- derive the batch count from the item slice so cardinality mismatches are unrepresentable instead of becoming user-facing verification errors or release-mode out-of-bounds access

## Rationale

The mismatched cardinalities in #543 are programmer invariants, but an assertion alone disappears in unchecked release builds. Grouping each message, public key, signature, and random scalar into one item enforces the relationship structurally and mirrors Lodestar-TS's array-of-signature-sets API.

Empty batches retain the existing verification-failure behavior.

Closes #543

### DIFF
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -21,7 +21,6 @@ const AggregatePublicKey = blst.AggregatePublicKey;
 const AggregateSignature = blst.AggregateSignature;
 const BlstError = @import("error.zig").BlstError;
 const BatchVerifyItem = @import("fast_verify.zig").BatchVerifyItem;
-const fast_verify = @import("fast_verify.zig");
 const SecretKey = @import("SecretKey.zig");
 const pippenger = @import("pippenger.zig");
 
@@ -263,18 +262,6 @@ pub fn verifyMultipleAggregateSignatures(
 ) (BlstError || PoolError || std.Io.Cancelable)!bool {
     const n_elems = items.len;
     if (n_elems == 0) return BlstError.VerifyFail;
-
-    // Avoid queue overhead for small batches or a single-worker pool.
-    if (n_elems <= 2 or pool.n_workers <= 1) {
-        var pairing_buf: PairingBuf = .{};
-        return fast_verify.verifyMultipleAggregateSignatures(
-            &pairing_buf.data,
-            items,
-            dst,
-            pks_validate,
-            sigs_groupcheck,
-        );
-    }
 
     const n_active = @min(pool.n_workers, n_elems);
 
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -20,6 +20,8 @@ const Signature = blst.Signature;
 const AggregatePublicKey = blst.AggregatePublicKey;
 const AggregateSignature = blst.AggregateSignature;
 const BlstError = @import("error.zig").BlstError;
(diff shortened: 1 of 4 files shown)


src/bls/ThreadPool.zig
```diff
@@ -256,31 +256,32 @@ const VerifyMultiWorkItem = struct {
...
-        pks.len != n_elems or
-        sigs.len != n_elems or
-        msgs.len != n_elems or
-        rands.len != n_elems)
-        return BlstError.VerifyFail;
+    const n_elems = items.len;
+    if (n_elems == 0) return BlstError.VerifyFail;
+
+    // Avoid queue overhead for small batches or a single-worker pool.
+    if (n_elems <= 2 or pool.n_workers <= 1) {
```
GrapeBaBa: this change may be better in a separate PR

lodekeeper-z: Agreed. Removed the small-batch/single-worker fast path from this PR in 2b4ea51a, so the change is now limited to the structural batch-item representation. Verified `zig build test:bls --summary all` and ReleaseSafe (14/14), the ReleaseSafe mainnet bindings build (10/10), and `pnpm lint` (passes with the same 10 warnings). The full local aggregate test run was blocked by the sandbox rejecting io_uring initialization in unrelated clock tests.

GrapeBaBa (approved):

src/bls/ThreadPool.zig
GrapeBaBa: You can open a separate PR for that

spiral-ladder (approved):

src/bls/ThreadPool.zig
lodekeeper-z: Opened #553 with the small-batch and single-worker direct-verification path isolated from this structural change. The focused BLS tests pass in Debug and ReleaseSafe (15/15), and lint passes with the existing warnings.

wemeetagain (approved):

PR #548, opened 2026-08-07
fix(bls): bound randomized aggregation inputs

### LINKED ISSUE #542: fix(bls): validate randomized aggregate input bounds before BLST
## Problem

The native randomized aggregate helpers trust caller-provided lengths before indexing fixed-size buffers or entering BLST:

- `src/bls/AggregatePublicKey.zig:46-74`
- `src/bls/AggregateSignature.zig:43-71`

Both write scalar pointers into `[MAX_AGGREGATE_PER_JOB]*const u8` without rejecting more than 128 inputs, and index `randomness[i * 32]` without checking that the randomness slice contains `inputs.len * 32` bytes.

The signature variant also accepts an empty input and calls `blst_p2s_mult_pippenger` with zero points. A local probe at `542779fc` produced a general-protection exception inside BLST.

## Impact

Malformed native API inputs can cause:

- a bounds panic with insufficient randomness,
- a fixed-array overflow above 128 inputs in unchecked builds,
- a BLST crash for empty randomized signature aggregation.

The NAPI and thread-pool paths already enforce analogous bounds, but these exported native helpers do not.

## Expected behavior

Return `BlstError.AggrTypeMismatch` before indexing or calling BLST when:

- the input is empty,
- input length exceeds `MAX_AGGREGATE_PER_JOB`,
- randomness length is less than `inputs.len * 32`.

## Acceptance criteria

- Apply consistent validation to both randomized PK and signature aggregation.
- Add regression tests for empty input, 129 inputs, zero-length randomness, and `32*n - 1` randomness.
- Confirm `zig build test:bls` and ReleaseSafe tests pass.

Found during the BLS security/correctness sweep of `542779fc`.

lodekeeper-z:
## Summary

- replace independent randomized-aggregate value and randomness slices with fixed-width per-item inputs
- reject empty and over-capacity native calls with dedicated recoverable errors before entering BLST
- preserve the 128-item boundary while making insufficient randomness unrepresentable
- update native tests and fuzz harnesses for the new public contract

## Rationale

The randomized aggregate helpers are exported native APIs, so empty and over-capacity slices are representable malformed inputs rather than impossible internal states. Assertions are not a stable public contract and disappear in unchecked release builds. Fixed-width per-item randomness removes the parallel-length invariant, while explicit errors protect the remaining runtime bounds.

Closes #542

### DIFF
--- a/src/bls/AggregatePublicKey.zig
+++ b/src/bls/AggregatePublicKey.zig
@@ -4,6 +4,12 @@ const Self = @This();
 /// An aggregate public key that can be used to verify aggregate signatures.
 point: c.blst_p1 = c.blst_p1{},
 
+/// A public key paired with its fixed-width random aggregation coefficient.
+pub const RandomizedPublicKey = struct {
+    public_key: *const PublicKey,
+    randomness: [32]u8,
+};
+
 /// Converts an aggregate public key back to a regular public key.
 /// This converts from projective coordinates back to affine coordinates.
 pub fn toPublicKey(self: *const Self) PublicKey {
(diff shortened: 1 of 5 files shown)


GrapeBaBa (approved):
LGTM

wemeetagain (approved):

PR #549, opened 2026-08-08
test(state_transition): harden PKIX load bounds and document trust boundary

### LINKED ISSUE #539: Harden PKIX load before production load/save use
Origin: follow-up from the 2026-08-04 blst-z/lodestar-z audit comment on ChainSafe/lodestar#9647.

## Problem

`pkix.load` / `pkix.save` are intended for an application-owned local cache snapshot, not operator-provided or cross-network input. The loader already checks framing, ABI compatibility, file size, capacity bounds, and checksum. Because the whole file is checksummed, restored affine points do not need separate semantic validation as long as the file remains inside that application-owned cache boundary.

This is acceptable while Lodestar does not call `pubkeyCache.load/save` today, and remains acceptable if future production use keeps the file under an application-owned, network-scoped cache path. Before wiring PKIX snapshots into production, the application should make that trust boundary explicit and avoid treating the file as generic user input.

## Acceptance Criteria

- Document the production trust boundary as application-owned local cache only.
- Keep the current capacity/file-size bounds, and add tests that crafted headers cannot force excessive allocation.
- Add malformed/corrupt/checksum fixture coverage for the chosen behavior.
- Ensure the application owns network scoping for the cache path; no separate "wrong-network" PKIX fixture is required.
- Link the Lodestar PR/commit that first wires production `load/save` usage, if any.

## Notes

This is not a current Lodestar #9728 blocker because Lodestar does not call `pubkeyCache.load/save` today. It is a gate before making PKIX snapshots part of a production startup path.

5t4rg4z3r:
## Motivation

`pkix.load` and `pkix.save` handle an application-owned local cache snapshot, but nothing in the code said so. A reader could reasonably mistake the loader's framing, ABI, bounds, and checksum checks for input validation and pass the file an operator-supplied path. The existing bounds tests asserted only the returned error, so they could not distinguish a header rejected before allocation from one rejected because an allocation failed, which is the property that actually matters for a crafted header.

Closes #539.

## Description

- Document the PKIX trust boundary as application-owned local cache data on the Zig module, the `load` function, and the `bindings/src/pubkeys.d.ts` declarations.
- Add `expectLoadRejectedWithoutAllocation`, which asserts a rejection happened with zero allocation attempts rather than because of a failed one.
- Cover crafted headers that declare oversized entry counts, an all-zero file, and a short file.
- Cover spare capacity clamping, so a header declaring `1 << 30` capacity still honors the caller limit.
- Cover payload and header corruption: bit flips across the key and affine regions, an entry reorder, and both checksum fields.

No behavior change. The loader's existing capacity and file-size bounds are unchanged.

## Verification

- `zig build test:state_transition test:ssz test:bls` (376 passed)
- `zig fmt --check src bindings test bench examples scripts build.zig`
- `pnpm lint`
- `pnpm exec vitest run bindings/test/pubkeys.test.ts` (19 passed)

Spec and ERA suites were not run locally; their download steps are absent in my environment and this change touches no consensus, SSZ, or ERA behavior.

This PR was written by an AI agent (stargazer), supervised by a human.

### DIFF
--- a/bindings/src/pubkeys.d.ts
+++ b/bindings/src/pubkeys.d.ts
@@ -21,11 +21,17 @@ export interface PubkeyCache {
    * Load a compatible PKIX file from the control environment while no workers are using the cache.
    * The explicit capacity limit bounds both the entry count and native allocation;
    * spare capacity recorded in the file is discarded above this limit.
+   *
+   * The file is application-owned local cache data, in the same trust class as the application's
+   * database files. The caller must keep it under a private cache directory and must scope the path
+   * to one network. PKIX stores no network identifier. The loader checks framing, ABI
+   * compatibility, bounds, and a checksum to detect corruption or an incompatible build.
+   * It does not authenticate the file. Do not pass an operator-supplied or network-sourced path.
    */
   load(filepath: string, maxCapacity: number): void;
   /** Testing-only reset from the control environment while no workers are using the cache. */
   reset(): void;
-  /** Save from the control environment. */
+  /** Save from the control environment. Writes application-owned local cache data. */
   save(filepath: string): void;
   /** Reserve exact native capacity when growing; existing larger capacity is retained. */
   ensureCapacity(capacity: number): void;
--- a/bindings/src/pubkeys.d.ts
+++ b/bindings/src/pubkeys.d.ts
@@ -22,11 +22,8 @@ export interface PubkeyCache {
    * The explicit capacity limit bounds both the entry count and native allocation;
    * spare capacity recorded in the file is discarded above this limit.
    *
-   * The file is application-owned local cache data, in the same trust class as the application's
-   * database files. The caller must keep it under a private cache directory and must scope the path
-   * to one network. PKIX stores no network identifier. The loader checks framing, ABI
-   * compatibility, bounds, and a checksum to detect corruption or an incompatible build.
-   * It does not authenticate the file. Do not pass an operator-supplied or network-sourced path.
+   * The file is application-owned local cache data. See the top-level doc comment in
+   * `src/state_transition/cache/pkix.zig` for the trust assumptions the caller must meet.
    */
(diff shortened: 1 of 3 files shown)


spiral-ladder:
just testing stargazer out

bindings/src/pubkeys.d.ts
```diff
@@ -21,11 +21,17 @@ export interface PubkeyCache {
    * Load a compatible PKIX file from the control environment while no workers are using the cache.
    * The explicit capacity limit bounds both the entry count and native allocation;
    * spare capacity recorded in the file is discarded above this limit.
+   *
+   * The file is application-owned local cache data, in the same trust class as the application's
+   * database files. The caller must keep it under a private cache directory and must scope the path
+   * to one network. PKIX stores no network identifier. The loader checks framing, ABI
+   * compatibility, bounds, and a checksum to detect corruption or an incompatible build.
+   * It does not authenticate the file. Do not pass an operator-supplied or network-sourced path.
```
spiral-ladder: This is very verbose and a repeat of what's in pkix.zig, can u just refer readers to that file's top level doc comment for trust assumptions

spiral-ladder (changes requested):

bindings/src/pubkeys.d.ts
5t4rg4z3r: Done in 3a2fec1cba52b4bd2a33e4479cb2176148f428c5. The `.d.ts` block is now two lines that point to the top-level doc comment in `src/state_transition/cache/pkix.zig`. The full trust assumptions stay in one place.

Verification: `zig fmt --check src bindings test bench examples scripts build.zig`, `pnpm lint` (10 pre-existing warnings, none in `pubkeys.d.ts`), `zig build test:state_transition` (123/123 pass).

This reply was written by an AI agent (stargazer), supervised by a human.

PR #553, opened 2026-08-10
perf(bls): bypass worker queue for small batches

lodekeeper-z:
## Summary

- route batches of one or two signature sets directly through the single-threaded verifier
- use the direct verifier for pools configured with one worker
- keep a call-local pairing buffer so concurrent callers do not share verifier state

## Rationale

This keeps the small-batch optimization removed from #547 in its own focused change, as requested in the review discussion. It avoids queue submission and wake-up overhead when parallel dispatch cannot help.

The branch is based on current `main`. I will refresh the small call-site change onto #547's structural `BatchVerifyItem` API after #547 lands if needed.

### DIFF
--- a/src/bls/ThreadPool.zig
+++ b/src/bls/ThreadPool.zig
@@ -20,6 +20,7 @@ const Signature = blst.Signature;
 const AggregatePublicKey = blst.AggregatePublicKey;
 const AggregateSignature = blst.AggregateSignature;
 const BlstError = @import("error.zig").BlstError;
+const fast_verify = @import("fast_verify.zig");
 const SecretKey = @import("SecretKey.zig");
 const pippenger = @import("pippenger.zig");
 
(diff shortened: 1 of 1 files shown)


src/bls/ThreadPool.zig
```diff
@@ -248,6 +249,17 @@ const VerifyMultiWorkItem = struct {
     }
 };
 
+fn useDirectVerification(n_elems: usize, n_workers: usize) bool {
```
GrapeBaBa: please inline the function and remove the useless test

lodekeeper-z: Done in 29946448. I inlined the direct-path condition and removed the helper-only test. `zig fmt --check`, `git diff --check`, BLS tests in Debug and ReleaseSafe, and `pnpm lint` all pass (lint has the same 10 existing warnings).

wemeetagain (approved):

PR #555, opened 2026-08-11
feat(pubkeys): add getPubkeyBytes binding

spiral-ladder:
Expose the cached 48-byte compressed pubkey bytes by validator index without materializing a PublicKey wrapper or serializing in JS. The BLS worker job path in Lodestar needs Uint8Array pubkeys, so this avoids the getOrThrow() + toBytes() round trip on the SingleAttestation gossip validation hot path.

Refs https://github.com/ChainSafe/lodestar/pull/9728#issuecomment-5237784361

### DIFF
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -132,6 +132,16 @@ pub fn get(index: js.Number) !?blst_bindings.PublicKey {
     return .{ .raw = public_key };
 }
 
+/// JS: pubkeys.getPubkeyBytes(index) → Uint8Array | undefined
+pub fn getPubkeyBytes(index: js.Number) !?js.Uint8Array {
+    if (!state.initialized) return error.PubkeyIndexNotInitialized;
+
+    const idx = try index.toU32();
+    const io = js.io();
+    const pubkey_bytes = state.cache.getPubkeyBytes(io, idx) orelse return null;
+    return try js.Uint8Array.fromExternal(pubkey_bytes[0..]);
+}
+
 /// Aggregate multiple `PublicKey`s by the given
 /// validator `indices` into one.
 ///
--- a/bindings/napi/pubkeys.zig
+++ b/bindings/napi/pubkeys.zig
@@ -139,7 +139,7 @@ pub fn getPubkeyBytes(index: js.Number) !?js.Uint8Array {
     const idx = try index.toU32();
     const io = js.io();
     const pubkey_bytes = state.cache.getPubkeyBytes(io, idx) orelse return null;
-    return try js.Uint8Array.fromExternal(pubkey_bytes[0..]);
+    return js.Uint8Array.from(pubkey_bytes[0..]);
 }
 
 /// Aggregate multiple `PublicKey`s by the given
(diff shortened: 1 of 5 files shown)


wemeetagain (dismissed):

bindings/napi/pubkeys.zig
```diff
@@ -132,6 +132,16 @@ pub fn get(index: js.Number) !?blst_bindings.PublicKey {
...
 }
 
+/// JS: pubkeys.getPubkeyBytes(index) → Uint8Array | undefined
+pub fn getPubkeyBytes(index: js.Number) !?js.Uint8Array {
+    if (!state.initialized) return error.PubkeyIndexNotInitialized;
+
+    const idx = try index.toU32();
+    const io = js.io();
+    const pubkey_bytes = state.cache.getPubkeyBytes(io, idx) orelse return null;
+    return try js.Uint8Array.fromExternal(pubkey_bytes[0..]);
```
twoeths: how about 
```suggestion
    return try js.Uint8Array.from(pubkey_bytes[0..]);
```

usually at lodestar side, the returned Uint8Array is a short-lived object that gets "gc" before we clone it via bls worker

twoeths (commented):
the general concern/question for me is when/how to use `fromExternal()` vs `from()`
given the current `fromExternal()` implementation in https://github.com/ChainSafe/zapi/blob/d6b21e193c9230d8da5fe4453331acfaff94396e/src/js/typed_arrays.zig#L75

```zig
pub fn fromExternal(slice: []const Element) !Self {
            const e = context.env();
            const buf = try context.allocator().dupe(Element, slice);

            const byte_len = slice.len * @sizeOf(Element);
            const len_hint: ?*anyopaque = @ptrFromInt(slice.len);
            const finalize_cb = comptime napi.wrapSliceFinalizeCallback(Element, externalFinalizer);
            const arraybuffer = e.createExternalArrayBuffer(std.mem.sliceAsBytes(buf), finalize_cb, len_hint) catch |err| {
                context.allocator().free(buf);
                return err;
            };

            _ = try e.adjustExternalMemory(@intCast(byte_len));
            const val = try e.createTypedarray(array_type, slice.len, arraybuffer, 0);
            return .{ .val = val };
        }
```
here we always alloc a memory at native there and count on a finalizer to free it
if we keep this `fromExternal()` implementation, we'll most likely to always go with `from()` where we let v8 alloc a memory and manage gc for us

bindings/napi/pubkeys.zig
spiral-ladder: makes sense

spiral-ladder: addressed in [5071c7c](https://github.com/ChainSafe/lodestar-z/pull/555/commits/5071c7cf66e353de19109fa021f9412b18e62582)

nazarhussain (dismissed):

bindings/src/pubkeys.js
```diff
@@ -25,6 +25,19 @@ export const pubkeyCache = {
...
+  },
+
+  getPubkeyBytesOrThrow(index) {
+     const pubkey = native.getPubkeyBytes(index);
+     if (pubkey === undefined) {
+       throw new Error(`pubkeyCache: index ${index} not found`);
+     }
+
+     return pubkey;
+   },
```
nazarhussain, line 39: LGTM, but if we can add a test specific for failure case of this function that will be very helpful.

nazarhussain (approved):

PR #557, opened 2026-08-12
docs: document security threat model

wemeetagain:
## Summary

- add a concise normative threat model covering assets, actors, trust boundaries, consensus trust invariants, security objectives, and finding classification
- separate volatile integration maturity, current call paths, cache mechanics, persistence details, controls, and open questions into a dated implementation map with an explicit owner
- require reviewers and contributors in `AGENTS.md` to consult both documents and revisit them when security-relevant integration boundaries change

## Why

Security reports have produced faulty findings because Lodestar and Ethereum trust assumptions were implicit. In particular, agents treated trusted beacon states and local ERA/database data as hostile, confused ancestry checks with full validation, and overlooked validation performed by Lodestar-ts.

Separating the stable security contract from current implementation mappings also reduces verbosity and limits documentation bitrot. The normative core changes only when trust or classification changes; implementation details can evolve in the audit map.

## Impact

Documentation only. Maintainers and automated reviewers gain a shorter basis for distinguishing security vulnerabilities, integration-readiness issues, consensus bugs, boundary hardening, trusted-input reliability problems, and operator misuse.

## AI assistance

This documentation was primarily AI-authored through an extended collaborative threat-modeling discussion with maintainer-provided Lodestar and Ethereum domain guidance.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,10 +12,14 @@
 - **Relative imports:** use the `.js` extension in TypeScript ESM imports.
 - **No `any`:** avoid `any` and `as any`; use proper types or a justified Biome suppression.
 - **Follow existing patterns** before introducing new abstractions.
-- **Security reviews:** read `THREAT_MODEL.md` before reporting a security finding. Identify the
-  attacker, current supported path or identified planned integration path, integration maturity,
-  trusted preconditions, violated security objective, and concrete impact. Classify same-process
-  hardening and operator misuse accurately.
+- **Security reviews:** read `THREAT_MODEL.md` and `docs/security/IMPLEMENTATION_MAP.md` before
+  reporting a security finding. Identify the attacker, current supported path or identified planned
+  integration path, integration maturity, trusted preconditions, violated security objective, and
+  concrete impact. Classify same-process hardening and operator misuse accurately.
+- **Security-model maintenance:** review both security documents when a change adds or alters a
+  trust boundary, native dependency, persistence path or format, shared mutable cache or pool,
+  externally influenced native input, or supported integration. Update the normative threat model
+  only when the security contract changes; otherwise update the implementation map.
 - **Incremental commits:** after review starts, do not force-push unless a maintainer requests it.
 - **Communication style:** do not use em dashes. Keep communication succinct and human-friendly.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,10 +12,11 @@
 - **Relative imports:** use the `.js` extension in TypeScript ESM imports.
 - **No `any`:** avoid `any` and `as any`; use proper types or a justified Biome suppression.
 - **Follow existing patterns** before introducing new abstractions.
-- **Security reviews:** read `THREAT_MODEL.md` and `docs/security/IMPLEMENTATION_MAP.md` before
-  reporting a security finding. Identify the attacker, current supported path or identified planned
-  integration path, integration maturity, trusted preconditions, violated security objective, and
-  concrete impact. Classify same-process hardening and operator misuse accurately.
+- **Security reviews:** first surface candidate security-objective violations, then use
+  `THREAT_MODEL.md` and `docs/security/IMPLEMENTATION_MAP.md` to classify them. The model is not an
+  allowlist. Downgrade a candidate only after verifying the applicable trust assumption against the
+  current or planned call path. If code and documentation conflict, report the code behavior and the
+  documentation gap.
 - **Security-model maintenance:** review both security documents when a change adds or alters a
   trust boundary, native dependency, persistence path or format, shared mutable cache or pool,
   externally influenced native input, or supported integration. Update the normative threat model
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,8 +13,9 @@
 - **No `any`:** avoid `any` and `as any`; use proper types or a justified Biome suppression.
 - **Follow existing patterns** before introducing new abstractions.
 - **Security reviews:** read `THREAT_MODEL.md` before reporting a security finding. Identify the
-  attacker, supported call path, trusted preconditions, violated security objective, and concrete
-  impact. Classify same-process hardening and operator misuse accurately.
+  attacker, current supported path or identified planned integration path, integration maturity,
+  trusted preconditions, violated security objective, and concrete impact. Classify same-process
+  hardening and operator misuse accurately.
 - **Incremental commits:** after review starts, do not force-push unless a maintainer requests it.
 - **Communication style:** do not use em dashes. Keep communication succinct and human-friendly.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,9 @@
 - **Relative imports:** use the `.js` extension in TypeScript ESM imports.
 - **No `any`:** avoid `any` and `as any`; use proper types or a justified Biome suppression.
 - **Follow existing patterns** before introducing new abstractions.
+- **Security reviews:** read `THREAT_MODEL.md` before reporting a security finding. Identify the
+  attacker, supported call path, trusted preconditions, violated security objective, and concrete
+  impact. Classify same-process hardening and operator misuse accurately.
 - **Incremental commits:** after review starts, do not force-push unless a maintainer requests it.
 - **Communication style:** do not use em dashes. Keep communication succinct and human-friendly.
 
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -73,10 +73,13 @@ Lodestar-z aims to preserve the following properties in supported builds and dep
 3. **Memory safety.** Malformed externally sourced bytes, points, indices, offsets, and collection
    lengths do not cause out-of-bounds access, use-after-free, double free, uninitialized-memory use,
    or allocator mismatch.
-4. **Process availability.** A single malformed or protocol-bounded remote input fails with an error
-   rather than a panic, deadlock, infinite loop, or unreasonable amplification of CPU or memory.
+4. **Process availability.** Attacker-controlled inputs and bounded sequences of inputs fail safely
+   and do not cause panics, deadlocks, infinite loops, cumulative leaks, or unbounded queue or cache
+   growth. CPU and memory amplification remain within explicit protocol, transport, and application
+   bounds.
 5. **State integrity.** Rejected candidate blocks do not partially modify the trusted pre-state,
-   publish derived epoch-cache entries, enter fork choice, or corrupt process-wide caches.
+   publish branch-specific epoch-cache entries, enter fork choice, or violate the documented
+   process-wide pubkey-cache invariant.
 6. **Lifecycle safety.** Supported Node.js worker creation, concurrent cache access, environment
    teardown, and finalization do not race or access freed global state.
 7. **Artifact integrity.** Published native artifacts are built from reviewed sources and pinned
@@ -94,7 +97,8 @@ The main assets are:
 - the correctness of beacon states, state roots, fork selection, validator accounting, and head
   selection;
 - Ethereum consensus safety and liveness for a Lodestar node using this implementation;
-- availability and memory integrity of the hosting Node.js process;
+- availability, integrity, and confidentiality of the hosting Node.js process memory, including
+  unrelated validator material, JWTs, API credentials, and other resident secrets;
 - correctness of BLS signature verification and aggregation;
 - integrity of process-wide node pools, pubkey caches, configuration, and worker-pool state;
 - confidentiality of BLS secret-key material intentionally passed to this package;
@@ -132,8 +136,13 @@ library can contain.
 
 Code loaded into the same process can invoke exported functions, allocate memory, read and write
 files using its own runtime privileges, terminate the process, and serialize a `SecretKey` by design.
-It is not a security boundary. N-API still validates types, lengths, indexes, and encodings to prevent
-accidental misuse from becoming native undefined behavior or a process crash.
+It is not an authorization or OS-isolation boundary. It remains a JavaScript-to-native memory-safety,
+host-process confidentiality, and cross-environment integrity boundary. Same-process callers,
+including compromised dependencies, may intentionally supply malformed arrays, coercing objects,
+stale handles, and adversarial buffers; those inputs must not yield arbitrary native memory access or
+cross-worker corruption. N-API entry points must validate runtime types, lengths, indexes, encodings,
+and buffer ranges before unsafe native access. Gaps are boundary-hardening or security findings
+according to current or planned reachability.
 
 ### Checkpoint state provider
 
@@ -166,9 +175,10 @@ publishing, and provenance are the relevant controls.
 | JavaScript to N-API | Runtime types, byte contents, lengths, indices, offsets, arrays, and object shapes | Same-process caller chooses which method to call and owns policy options |
 | P2P object to SSZ | Bytes, offsets, list lengths, bitfields, and encodings are hostile | Lodestar-ts applies transport bounds; native decoding still enforces canonical SSZ and consensus limits |
 | Beacon-state deserialization | Bytes come from a trusted anchor, trusted database, or trusted ERA source | Deserialization establishes structural SSZ validity only; provenance establishes state trust |
-| State transition | Serialized signed block contents are hostile | Pre-state is trusted; full STF or equivalent prior checks are required; execution and DA statuses are accurate |
+| State transition and import | Serialized signed block contents are hostile | Pre-state is trusted; full STF or equivalent prior checks are required; publication is gated on joined execution and DA results |
 | Fork choice | Valid blocks and validated attestations still represent competing branches | Only STF-valid blocks enter; attestations passed the applicable upstream validation; local time is trusted |
 | BLS API | Serialized points, messages, list contents, and cardinality | Boolean validation options are caller policy; proof-of-possession preconditions apply where documented |
+| Zig to C/native dependencies | JavaScript- or network-influenced values may reach dependency calls | Lodestar-z ensures valid representation, initialization, cardinality, pointer lifetime, requested validation, and ABI compatibility; pinned dependencies are trusted to honor their documented contracts |
 | Shared native state | Worker scheduling and teardown are asynchronous | Configuration and administrative cache operations follow their lifecycle contract |
 | Checkpoint provider | Response may be malformed; provider is semitrusted | User-provided checkpoint root is authoritative, or checkpoint selection is explicitly delegated |
 | Database, PKIX, and ERA/E2S | Fallible local I/O, format compatibility, and accidental corruption | Contents and provenance are trusted; malicious local replacement is out of scope |
@@ -212,15 +222,16 @@ or construction entry point requires trusted state bytes. Trust originates at on
 - a user-provided checkpoint root, after the downloaded state's hash-tree-root is matched to it; or
 - explicit delegation of checkpoint selection to a checkpoint state provider.
 
-Subsequent states inherit trust through successful state transitions. A state written to trusted
-storage retains the trust it had when written. SSZ deserialization checks structural validity; it
-does not establish provenance, consensus validity, canonicality, or finality.
+Subsequent states inherit authenticated provenance through successful state transitions. A state
+written to trusted storage retains the provenance and validity status it had when written. SSZ
+deserialization checks structural validity; it does not establish provenance, consensus validity,
+canonicality, or finality.
 
 ```text
 genesis or authenticated checkpoint
                  |
                  v
-       trusted beacon pre-state
+      eligible trusted pre-state
                  |
         hostile candidate block
                  |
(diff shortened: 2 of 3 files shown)


AGENTS.md
```diff
@@ -12,6 +12,9 @@
 - **Relative imports:** use the `.js` extension in TypeScript ESM imports.
 - **No `any`:** avoid `any` and `as any`; use proper types or a justified Biome suppression.
 - **Follow existing patterns** before introducing new abstractions.
+- **Security reviews:** read `THREAT_MODEL.md` before reporting a security finding. Identify the
+  attacker, supported call path, trusted preconditions, violated security objective, and concrete
```
lodekeeper-z: This conflicts with the maturity model below. State transition and Zig fork choice intentionally have no current supported production path, but the threat model still wants release-blocking security-readiness findings against an identified planned path. Please say `current supported path or identified planned integration path`, then require the report to classify its maturity. Otherwise this critical rule tells automated reviewers to suppress exactly the planned-component findings that THREAT_MODEL.md:46-60 permits.

lodekeeper-z (commented):
The structure and attacker/path/classification discipline are strong, but I found five threat-model accuracy gaps that should be resolved before this becomes the review baseline. The most important are the DA ordering claim and the process-wide pubkey-cache rollback/finalization description. Details are inline.

Review scope: AGENTS.md and THREAT_MODEL.md at b8d60d4e. I also checked the corresponding Lodestar-ts block verification path and current Lodestar-z state-transition/cache code. CI is green on Ubuntu, ARM, bindings, title, and spec preparation; macOS was still pending at the time of review.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+executing in the Node.js process.
+
+## Assets and impact
+
+The main assets are:
+
+- the correctness of beacon states, state roots, fork selection, validator accounting, and head
+  selection;
+- Ethereum consensus safety and liveness for a Lodestar node using this implementation;
+- availability and memory integrity of the hosting Node.js process;
```
lodekeeper-z: Please include confidentiality of host-process memory here, not only integrity/availability and BLS secret keys intentionally passed to this package. A remotely reachable native out-of-bounds read could disclose JWTs, validator material, API credentials, or other Lodestar secrets resident in the Node process even though arbitrary same-process JavaScript is out of scope. That is still an in-scope consequence of crossing a remote-to-native memory-safety boundary.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+checkpoint or initial state, file paths, cache sizing, execution status, data-availability status,
+verification policy, and wall-clock time. These are trusted control-plane inputs. Accidental
+mistakes should fail clearly where practical, but a malicious operator is not an attacker this
+library can contain.
+
+### Same-process JavaScript or Zig code
+
+Code loaded into the same process can invoke exported functions, allocate memory, read and write
+files using its own runtime privileges, terminate the process, and serialize a `SecretKey` by design.
+It is not a security boundary. N-API still validates types, lengths, indexes, and encodings to prevent
```
lodekeeper-z: This is written as a current blanket guarantee, but the current binding does not establish it everywhere. For example, `BeaconStateView.stateTransition` uses `std.debug.assert(bytes.len >= 12)` and then reads `bytes[offset..][0..8]` before deserialization (`bindings/napi/BeaconStateView.zig:1311-1320`). Please make this normative: N-API entry points *must* validate these properties, and gaps are boundary-hardening or security findings according to reachability. That also aligns this paragraph with lines 197-200 instead of overstating the implementation.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+pre-states afterward.
+
+An execution-optimistic state is conditionally trusted. Each block carries its own execution status.
+While the EL is syncing, optimistic blocks and their descendants remain `syncing`. When the EL later
+reports valid or invalid, fork choice propagates that result through the affected branch. Latest
+valid hash processing makes an invalid block and all descendants immediately ineligible, revokes
+their conditional trust, and prevents their use as trusted pre-states. Lodestar-ts is responsible for
+preventing validator duties while the node is execution optimistic.
+
+DA must be satisfied before STF. Inside the DA window this means DA validation has completed. A
```
lodekeeper-z: This ordering is not accurate for the current Lodestar import pipeline. `verifyBlock.ts:113-152` runs DA verification and STF concurrently in one `Promise.all`, while STF is deliberately passed `DataAvailabilityStatus.Available`; the real DA statuses are joined before import. Please say DA must be satisfied before *block import/acceptance*, and explain that it may be verified concurrently with STF when the all-or-none result gates publication. Otherwise this document would incorrectly flag the intended production path as violating its trust contract.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+The addon shares BLS worker-pool state, the persistent Merkle node pool, configuration, metrics, and
+the pubkey cache across Node.js environments. Sharing is intentional.
+
+- Loading and unloading supported workers must not tear down state still used by another
+  environment.
+- Pubkey-cache movable storage is protected by its lock, and references into it must not escape a
+  resize.
+- PKIX save, load, and reset are restricted to the control environment. Reads and supported append
+  operations may be shared.
+- The pubkey cache contains finalized validator-registry history. Pending deposits are processed into
```
lodekeeper-z: The finalization/rollback description does not match the current cache behavior. `EpochCache.clone` shares the same `PubkeyCache` pointer, and `addValidatorToRegistry` appends immediately during candidate processing (`src/state_transition/cache/epoch_cache.zig:490-495`, `src/state_transition/block/process_deposit.zig:175-185`). A later operation or state-root failure destroys the clone but does not roll back that append. The intended invariant appears to be *unforkable append-only validator history*, with state-length checks preventing a cached future index from being treated as present, not publication only at finalization. Please document that precise invariant and separate former Eth1 deposits from pending execution-layer deposits processed after finalization. Lines 386-389 should also avoid claiming that every failed STF leaves no validity-influencing cache mutation unless the implementation actually stages these appends.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+```
+
+P2P objects are the principal hostile data source. Lodestar-ts currently owns networking, message
+size and decompression limits, peer scoring, and orchestration. P2P objects that will cross into
+Lodestar-z may still contain adversarial SSZ bytes. Beacon states, ERA data, database contents,
+configuration, wall-clock time, and application policy have different trust contracts described
+below. Reviews must not treat every serialized object or JavaScript argument as remotely controlled.
+
+The package also supports direct use by JavaScript and Zig callers. Those callers are in the same
+process and privilege domain as Lodestar-z. The addon is not a sandbox around hostile JavaScript.
```
lodekeeper-z: `Same process and privilege domain` is accurate for authorization and OS privileges, but it should not erase the JavaScript-to-native memory-safety boundary. A compromised dependency or plugin can intentionally supply malformed arrays, coercing objects, stale handles, and adversarial buffers; it does not thereby already possess arbitrary native-memory access or cross-worker corruption capability. Please distinguish `not an authorization/OS isolation boundary` from the still-required memory-safety, process-confidentiality, and cross-environment integrity boundary.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+   all required checks completed locally or equivalently beforehand, and accurate external execution
+   and data-availability results, state transition and fork choice agree with the consensus
+   specification version pinned in `build.zig.zon`.
+2. **Cryptographic correctness.** BLS operations do not accept invalid signatures or points when the
+   relevant validation is requested by their API contract. Randomized batch operations use
+   unpredictable, nonzero coefficients and preserve the required cardinality.
+3. **Memory safety.** Malformed externally sourced bytes, points, indices, offsets, and collection
+   lengths do not cause out-of-bounds access, use-after-free, double free, uninitialized-memory use,
+   or allocator mismatch.
+4. **Process availability.** A single malformed or protocol-bounded remote input fails with an error
```
lodekeeper-z: The availability objective is limited to a single malformed or protocol-bounded input, which misses cumulative attacks where each item is individually bounded but repeated items leak allocations, grow a queue/cache, retain references, or starve workers. Please cover attacker-controlled sequences and rates permitted by supported integrations, including cumulative leaks and unbounded queue/cache growth. Peer scoring can matter to severity, but it does not make these patterns satisfy the objective.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+A checkpoint state provider is usually semitrusted. If the user supplies a checkpoint root, matching
+the downloaded state's hash-tree-root to that root is a mandatory part of establishing the trust
+chain. The exact Lodestar-ts or Lodestar-z integration point that performs this check is not yet
+fixed. If the user supplies no root, trust in checkpoint selection is explicitly delegated to the
+provider. A malicious provider under delegated trust is outside this threat model, while safely
+handling malformed responses remains worthwhile boundary hardening.
+
+### Trusted local storage
+
+Beacon states read from the database, ERA/E2S data, and PKIX snapshots are application-owned trusted
```
lodekeeper-z: The storage trust model should be conditional on filesystem provenance rather than treating every local replacement as host compromise. A less-privileged local user or co-tenant may be able to replace, redirect, or race a cache/ERA file or writable parent directory while Lodestar runs under a more privileged service account. Semantic authentication can remain the caller's responsibility, but memory-safety and privilege-escalation impact from such a path should be in scope when ownership, permissions, mounts, or symlink handling do not preserve provenance.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+| JavaScript to N-API | Runtime types, byte contents, lengths, indices, offsets, arrays, and object shapes | Same-process caller chooses which method to call and owns policy options |
+| P2P object to SSZ | Bytes, offsets, list lengths, bitfields, and encodings are hostile | Lodestar-ts applies transport bounds; native decoding still enforces canonical SSZ and consensus limits |
+| Beacon-state deserialization | Bytes come from a trusted anchor, trusted database, or trusted ERA source | Deserialization establishes structural SSZ validity only; provenance establishes state trust |
+| State transition | Serialized signed block contents are hostile | Pre-state is trusted; full STF or equivalent prior checks are required; execution and DA statuses are accurate |
+| Fork choice | Valid blocks and validated attestations still represent competing branches | Only STF-valid blocks enter; attestations passed the applicable upstream validation; local time is trusted |
+| BLS API | Serialized points, messages, list contents, and cardinality | Boolean validation options are caller policy; proof-of-possession preconditions apply where documented |
```
lodekeeper-z: Please add the Zig-to-C/native-dependency boundary explicitly. Attacker-shaped BLS inputs ultimately determine counts, pointer arrays, validation flags, scalar widths, and work passed to BLST. Pinned versions and valid serialized points do not protect against a wrapper ABI/cardinality/lifetime bug. Lodestar-z remains responsible for representation, initialization, cardinality, lifetime, and validation correctness at that boundary.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+                               |
+                    fork choice or trusted storage
+```
+
+For this model, a trusted beacon state is:
+
+- structurally valid SSZ;
+- consensus-layer valid, conditional on the execution and DA assertions supplied by the application;
+- descended from a trusted anchor; and
+- associated with a block branch that is eligible for fork choice at the time it is accepted.
```
lodekeeper-z: This definition conflates provenance/consensus validity with current fork-choice eligibility, even though the next section says these properties must not be conflated. A state can be valid and descended from a trusted anchor while no longer being eligible as a pre-state after finalization. Please define an authenticated or consensus-valid state independently, then define an `eligible trusted pre-state` as that status plus current branch eligibility. Execution invalidity can revoke conditional validity; finalization of a competing branch should revoke eligibility, not historical validity/provenance.

THREAT_MODEL.md
```diff
@@ -0,0 +1,586 @@
...
+
+| Property | Guarantee for a trusted beacon state |
+| --- | --- |
+| Structural SSZ validity | Always |
+| Consensus-layer validity | Always, conditional on correct external execution and DA assertions |
+| Descent from trusted anchor | Always |
+| Canonical | Not necessarily |
+| Finalized | Not necessarily |
+| Persisted | Not necessarily |
+| Execution status | Valid or conditionally trusted while `syncing` |
```
lodekeeper-z: The execution-status table omits Gloas `payload_separated`. Current Zig fork choice explicitly treats both `syncing` and `payload_separated` as execution-optimistic (`src/fork_choice/proto_array.zig`, `src/fork_choice/fork_choice.zig`). Please include `payload_separated` and state that it is conditionally trusted/optimistic until payload resolution.

lodekeeper-z (commented):
Supplemental review after the independent security and consistency passes completed. These are distinct from the five findings in my first review and focus on boundary coverage and terminology.

THREAT_MODEL.md
wemeetagain: Thanks for calling this out. We discussed it and are keeping the existing storage trust assumption. Database, PKIX, and ERA contents and provenance are trusted, and malicious local replacement remains outside this threat model. Normal filesystem ownership and isolation are deployment preconditions rather than a Lodestar-z-specific adversarial boundary. Remote replacement through a supported workflow, or violations of the documented loader contracts, remain in scope.

THREAT_MODEL.md
wemeetagain: Thanks. We discussed this and are not adding `payload_separated` to the threat model. Gloas payload resolution (`EMPTY` versus `FULL`) is orthogonal to execution validity (`valid`, `syncing`, or `invalid`), so treating payload separation as an execution-optimistic status would conflate two dimensions. The current Zig `ExecutionStatus.payload_separated` is an implementation concern to address separately rather than a trust-model concept.

THREAT_MODEL.md
```diff
@@ -0,0 +1,615 @@
...
+according to current or planned reachability.
+
+### Checkpoint state provider
+
+A checkpoint state provider is usually semitrusted. If the user supplies a checkpoint root, matching
+the downloaded state's hash-tree-root to that root is a mandatory part of establishing the trust
+chain. The exact Lodestar-ts or Lodestar-z integration point that performs this check is not yet
+fixed. If the user supplies no root, trust in checkpoint selection is explicitly delegated to the
+provider. A malicious provider under delegated trust is outside this threat model, while safely
+handling malformed responses remains worthwhile boundary hardening.
```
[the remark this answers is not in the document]
wemeetagain: Raw checkpoint bytes are hostile to whichever component first decodes them before root authentication. The current Lodestar-z contract requires trusted state bytes, and no planned path yet makes `createFromBytes` that initial decoder. If checkpoint response decoding moves into Lodestar-z, that entry point becomes a hostile-input boundary and must use checked parsing rather than assertions.

lodekeeper-z (approved):
Re-reviewed at 2639d32e.

The split addresses my main maintainability concern: `THREAT_MODEL.md` is now a substantially shorter normative contract, while current integration status and implementation mechanics are isolated in `docs/security/IMPLEMENTATION_MAP.md` with ownership, a review date, and explicit update triggers in `AGENTS.md`.

The revised text also resolves the substantive issues from my earlier reviews: planned-path classification, host-process confidentiality, N-API validation wording, concurrent DA/STF gating, append-only pubkey-cache behavior, the JavaScript/native and Zig/native boundaries, cumulative exhaustion, state provenance versus eligibility, and the Gloas status distinction. The two deliberately out-of-scope cases are now stated consistently rather than left ambiguous.

I found no remaining blocking issues. Relative links and `git diff --check` pass. The primary build and binding checks are green; the remaining spec matrix is still running.

THREAT_MODEL.md
```diff
@@ -0,0 +1,246 @@
...
+   dependencies through the repository release process.
+
+## Actors and trusted inputs
+
+| Actor or source | Capability and trust |
+| --- | --- |
+| Remote consensus participant | Controls protocol-shaped and malformed P2P objects. It does not directly control native options, configuration, local files, or capacity APIs without a demonstrated Lodestar path. |
+| Remote API user | Controls only values forwarded by Lodestar's API layer after its validation and limits. Reports must trace that path. |
+| Operator and host application | Trusted to choose the preset, configuration, initial state, local files, verification policy, external statuses, capacities, and wall-clock time. Accidental misuse should fail clearly, but a malicious operator is out of scope. |
+| Same-process JavaScript or Zig code | Has ambient application privileges and may intentionally provide malformed or coercing values. It is not an authorization or OS-isolation boundary, but it does not thereby possess arbitrary native-memory or cross-worker access. |
```
lodekeeper-z: This combines callers with different capabilities. Same-process JavaScript does not thereby have arbitrary native-memory access, so malformed values crossing N-API can add capability. Arbitrary same-process Zig code already executes natively and can construct/dereference pointers, so it is not the same memory-safety boundary. Please split JavaScript and Zig callers so later security-vs-hardening classification does not rely on a false shared capability model.

THREAT_MODEL.md
```diff
@@ -0,0 +1,246 @@
...
+
+A state read from trusted storage retains the provenance and validity status it had when written.
+Subsequent states inherit authenticated provenance through successful consensus-layer state
+transitions.
+
+An **authenticated CL-derived state** is structurally valid SSZ, descended from a trusted anchor,
+produced by a successful consensus-layer transition, and accompanied by the execution and DA status
+used at acceptance.
+
+An **eligible trusted pre-state** is an authenticated CL-derived state whose branch remains eligible
```
lodekeeper-z: This definition excludes the initial anchors above. Genesis and an authenticated checkpoint have trusted provenance but were not themselves produced by a successful CL transition, so neither is an `authenticated CL-derived state` under the preceding definition. As written, they cannot be eligible pre-states for their first child. Please define eligibility over authenticated anchors *or* CL-derived states, or include anchors in the authenticated-state definition.

docs/security/IMPLEMENTATION_MAP.md
```diff
@@ -0,0 +1,194 @@
...
+Update this map when a change adds or alters a trust boundary, native dependency, persistence path
+or format, shared mutable cache or pool, externally influenced native input, or supported
+integration. Update the threat model only when the security contract or finding classification also
+changes.
+
+## Integration maturity
+
+| Component | Current status | Review treatment |
+| --- | --- | --- |
+| BLS bindings | Integrated | Production-reachable |
```
lodekeeper-z: `Integrated` / `Production-reachable` overstates current Lodestar-ts maturity for both this row and the pubkey-cache row below. Current Lodestar-ts has no `@chainsafe/lodestar-z` dependency/import; its production BLS paths still use `@chainsafe/blst` (for example `packages/beacon-node/src/chain/bls/interface.ts`) and state transition still uses the TypeScript pubkey cache (`packages/state-transition/src/cache/pubkeyCache.ts`). Please distinguish exported/implemented addon surfaces from components actually wired into Lodestar production. This also affects the security-vulnerability versus security-readiness classification.

docs/security/IMPLEMENTATION_MAP.md
```diff
@@ -0,0 +1,194 @@
...
+
+Explicit local APIs remain caller-controlled policy. For example, allowing the caller of a cache
+save API to choose its output path is not independently path traversal.
+
+### Beacon-state loading
+
+`BeaconStateView.createFromBytes` and ERA state loading currently accept state bytes under the
+trusted-state contract. SSZ deserialization checks structure but does not authenticate provenance.
+
+Checkpoint-root authentication is not yet assigned to a stable Lodestar-ts or Lodestar-z integration
```
lodekeeper-z: Current Lodestar-ts does have an established checkpoint-authentication path. `packages/state-transition/src/util/weakSubjectivity.ts:166-178` computes the downloaded state's checkpoint root/epoch and compares both with the supplied checkpoint; `packages/cli/src/cmds/beacon/initBeaconState.ts:34-74` invokes it during checkpoint initialization. The state is decoded before that comparison. Please document that current boundary, while separately noting that ownership may change if decoding moves into Lodestar-z.

docs/security/IMPLEMENTATION_MAP.md
```diff
@@ -0,0 +1,194 @@
...
+point. If Lodestar-z becomes the first decoder of raw checkpoint responses, its slot extraction and
+state decoding must become checked hostile-input parsing before root comparison.
+
+### P2P block import
+
+Lodestar-ts currently owns networking, framing, decompression limits, peer scoring, and import
+orchestration.
+
+- Gossip blocks receive the consensus-spec gossip validation applicable to that object.
+- Range, unknown-block, and backward sync establish ancestry through the parent-root hash chain.
```
lodekeeper-z: Please separate unknown-parent recovery/forward sync from archive-only historical backfill. Unknown-parent recovery processes descendants forward through `chain.processBlock`, but `packages/beacon-node/src/sync/backfill/backfill.ts:813-858` verifies reverse hash-chain continuity and proposer signatures, then writes directly to `blockArchive` without full STF. The current wording plus “Full STF is required before every import” can incorrectly classify archive backfill behavior as a contract violation; scope the STF requirement to blocks entering live chain/fork choice.

docs/security/IMPLEMENTATION_MAP.md
```diff
@@ -0,0 +1,194 @@
...
+- an index at or beyond the current state's validator count is treated as absent;
+- later valid processing must reproduce the same pubkey at that index; and
+- conflicting, duplicate, or sparse appends fail.
+
+Former Eth1 bridge deposits can register validators and append during block processing. In Electra,
+the new validator initially has zero balance while its amount remains pending. Execution-layer
+deposit requests remain pending and register a new validator only after their source slot is
+finalized.
+
+Other state-transition caches are generally epoch-scoped and candidate-specific. A failed STF may
```
lodekeeper-z: This is not accurate for the reused epoch-transition cache. `src/state_transition/cache/epoch_transition_cache.zig:50-160` defines a process-global cache; `EpochTransitionCache.init` mutates it after the lookup/resize lock is released, and returned candidate caches borrow its arrays (`:255-345`, `:536-545`). Please list this shared cache and its serialization/lifecycle assumptions. Describing all other STF caches as candidate-specific would hide a real overlapping-transition / multi-environment review boundary.

lodekeeper-z (changes requested):
Follow-up verification against the current Lodestar-ts and Lodestar-z sources found several accuracy gaps that the implementation map is intended to prevent. This review supersedes my earlier approval. The normative/implementation split remains a strong improvement; the requested changes are scoped corrections to the resulting baseline.

lodekeeper-z (approved):
Re-reviewed at 98483192. The follow-up resolves all five accuracy gaps from my changes-requested review: JavaScript and native Zig callers now have distinct capability models; authenticated anchors are included in eligible trusted pre-states; the BLS/pubkey-cache surfaces are correctly classified as exported but not wired into Lodestar-ts production; the current Lodestar-ts checkpoint decode/authentication path is documented; live-chain import is separated from archive-only backfill; and the process-global reused epoch-transition cache plus its borrowed-lifetime serialization requirements are explicit.

I cross-checked the changed wording against the current Lodestar-ts and Lodestar-z sources. `git diff --check` passes, all relative documentation links resolve, and the current CI checks are green or still running with no failures. No blocking issues remain.

THREAT_MODEL.md
```diff
@@ -0,0 +1,138 @@
...
+
+Lodestar-z is a Zig consensus library and Node.js native addon. Its host integration owns networking,
+API exposure, checkpoint acquisition, execution-layer communication, and validator duties.
+
+The assets protected here are:
+
+- Ethereum consensus safety and liveness;
+- state-transition, fork-choice, and BLS correctness;
+- native memory safety and availability of the hosting process; and
+- integrity of shared native state across calls and Node.js environments.
```
lodekeeper-z: **Blocking:** The simplification removes source-dependency and native release-artifact integrity from the protected assets, and the later objectives/boundaries no longer cover build or release trust. Because classification now requires a violated security objective, a malicious dependency pin, workflow, or substituted published addon would fall outside this baseline even though it is a Lodestar-z security path. Please retain a concise artifact-integrity asset/objective and a build/release trust boundary; volatile workflow details can remain in the implementation map.

lodekeeper-z (changes requested):
Re-reviewed the simplification at 552c360b. The shorter runtime trust model preserves the earlier integration and shared-cache corrections, and the implementation-map links and source anchors validate. One scope regression remains: the rewrite drops supply-chain and release-artifact integrity from the protected assets, objectives, boundaries, and implementation map. That makes dependency, workflow, or artifact-substitution findings impossible to classify under the document's own required-objective framework. Please retain a concise supply-chain/release objective and boundary.

lodekeeper-z (approved):
Re-reviewed at 489787d6. This restores the missing supply-chain and release-artifact integrity coverage across the protected assets, attacker model, security objectives, trust boundaries, and implementation map. I verified the cited dependency pins and release workflow controls (SHA-pinned actions, ReleaseSafe artifacts, frozen lockfile, and npm provenance); relative links and `git diff --check` pass. No blocking issues remain.

spiral-ladder (approved):
overall looks good, we can tweak the doc according as we find issues

PR #559, opened 2026-08-13
feat(shuffle): add swap-or-not shuffling module and binding

nazarhussain:
## Summary

Adds a Zig `swap_or_not_shuffle` module and JS bindings for the shuffle APIs Lodestar uses, so ChainSafe/lodestar can drop `@chainsafe/swap-or-not-shuffle`.

- Shares one shuffle implementation between `state_transition` and the JS binding.
- Matches the upstream Rust/reference behavior for implemented paths.
- Returns explicit errors for degenerate inputs that the reference package would crash or hang on.
- Leaves async binding support for a future zapi release, since Lodestar does not use it.

## Testing

- Ported upstream shuffle reference tests and existing lodestar-z shuffle vectors.
- Added Zig coverage for validation order, error paths, proposer index, and sync committee selection.
- Checked beta metrics against unstable; no performance regression found.

## Follow-ups

- #563 adds PTC sampling on top of this.
- Async shuffle bindings can be added later if a consumer appears.


bindings/src/bindings.js
```diff
@@ -3,4 +3,11 @@ import {requireNapiLibrary} from "@chainsafe/zapi";
 
 const bindings = requireNapiLibrary(join(import.meta.dirname, "../.."));
 
+// Namespace-level constants mirroring @chainsafe/swap-or-not-shuffle; the
+// zapi DSL only auto-exports functions and classes, so values are attached
+// here instead of via a native register hook.
+bindings.shuffle.SHUFFLE_ROUNDS_MAINNET = 90;
+bindings.shuffle.SHUFFLE_ROUNDS_MINIMAL = 10;
+bindings.shuffle.ByteCount = {One: 1, Two: 2};
```
nazarhussain, line 11: This will soon go away as we will support namepsace `pub const` and `enum` in the Zapi.

bindings/napi/async_task.zig
```diff
@@ -0,0 +1,158 @@
...
+//!
+//! zapi has no async DSL — `js.Promise` requires dropping to the low-level
+//! `napi.AsyncWork`/`Deferred` layer for worker-thread resolution — so this
+//! module confines that raw plumbing to one place. It is an upstream
+//! candidate for a zapi `js.AsyncTask`; `transferOwnedSlice` is superseded
+//! by zapi#68 `OwnedTypedArray` once zapi 4.0 ships.
+//!
+//! A Task is any struct providing:
+//!  - `compute(self: *Task) !void` — libuv worker thread; MUST NOT call napi APIs
+//!  - `resolve(self: *Task, env: napi.Env) !napi.Value` — JS thread; builds the
```
wemeetagain: optional `reject` too?

Also which are required and which are optional? Would be nice to include as part of the interface definition.

nazarhussain: Resolved in latest commit.

bindings/napi/async_task.zig
```diff
@@ -0,0 +1,177 @@
+//! Interim generic async-task plumbing for napi bindings.
+//!
+//! zapi has no async DSL — `js.Promise` requires dropping to the low-level
+//! `napi.AsyncWork`/`Deferred` layer for worker-thread resolution — so this
+//! module confines that raw plumbing to one place. It is an upstream
+//! candidate for a zapi `js.AsyncTask`; `transferOwnedSlice` is superseded
+//! by zapi#68 `OwnedTypedArray` once zapi 4.0 ships.
```
spiral-ladder: a lot of ai slop doc imo, keep it more concise, imo the next reader doesnt need to know that zapi has no async DSL in the context of this, that should be left for an issue in zapi repo to support such a DSL if we want

```suggestion
//! Async-task plumbing for napi bindings.
```

bindings/napi/async_task.zig
```diff
@@ -0,0 +1,177 @@
...
+//! Optional:
+//!  - `reject(self: *Task, env: napi.Env, err: anyerror) !napi.Value` — JS
+//!    thread; builds the rejection value for a failed `compute`. Takes
+//!    precedence over `errorMessage`.
+//!  - `errorMessage(err: anyerror) [:0]const u8` — maps a failed `compute`'s
+//!    error to the rejection Error's message. Defaults to `@errorName(err)`.
+//!
+//! Ownership: on `spawn` error the task is NOT consumed — the caller's
+//! errdefers must free its resources. After `spawn` returns successfully the
+//! helper owns the task and calls `deinit` in the complete callback.
```
spiral-ladder: This feels like it should belong in spawn rather than here

bindings/src/bindings.js
```diff
@@ -3,4 +3,11 @@ import {requireNapiLibrary} from "@chainsafe/zapi";
 
 const bindings = requireNapiLibrary(join(import.meta.dirname, "../.."));
 
+// Namespace-level constants mirroring @chainsafe/swap-or-not-shuffle; the
+// zapi DSL only auto-exports functions and classes, so values are attached
+// here instead of via a native register hook.
```
spiral-ladder: dont need to know that it mirrors swap-or-not-shuffle, the inherent assumption is it does, we also dont need to know how zapi works here

```suggestion
```

bindings/napi/shuffle.zig
```diff
@@ -9,8 +52,211 @@ pub fn innerShuffleList(list: js.Uint32Array, seed: js.Uint8Array, rounds: js.Nu
...
+    return shuffleListSync(active_indices, seed, rounds, true);
+}
+
+/// JS: unshuffleList(activeIndices: Uint32Array, seed: Uint8Array, rounds: number): Uint32Array
+pub fn unshuffleList(active_indices: js.Uint32Array, seed: js.Uint8Array, rounds: js.Number) !js.Uint32Array {
+    return shuffleListSync(active_indices, seed, rounds, false);
+}
+
+/// The input array is copied, shuffled in-place inside the copy, and returned;
+/// the caller's array is never mutated (same as the reference).
```
spiral-ladder:
```suggestion
/// Copies the input array and shuffles the copy.
///
/// Returns the copy.
```

bindings/src/index.d.ts
```diff
@@ -369,15 +369,78 @@ export declare class BeaconStateView {
...
   pool: {
     ensureCapacity: (capacity: number) => void;
   };
   config: {
     set: (chainConfig: object, genesisValidatorsRoot: Uint8Array) => void;
   };
+  /**
+   * API-compatible port of the `@chainsafe/swap-or-not-shuffle` package,
+   * plus the in-place `innerShuffleList` variant.
+   */
```
spiral-ladder:
```suggestion
```

bindings/napi/async_task.zig
```diff
@@ -0,0 +1,177 @@
...
+
+    return .{ .val = ctx.deferred.getPromise() };
+}
+
+/// Transfers an `allocator`-owned slice to a JS TypedArray without copying;
+/// V8 frees the buffer via a finalizer when the array is collected. On
+/// success `data.*` becomes empty so a later unconditional `free` is a no-op
+/// (mirrors zapi#68's transactional `OwnedTypedArray` contract). Empty input
+/// resolves to a regular zero-length typed array.
+pub fn transferOwnedSlice(
```
spiral-ladder: can we not just release a new zapi rather than use this?

spiral-ladder (commented):
mostly doc related comments, i kinda am of the opinion that we should cut down or limit the amount of comments or the quality of AI comments because they tend to write in a very roundabout way. in an isolated PR it seems fine but over the course of many PRs our codebase comments will become meaningless

wemeetagain (approved):
LGTM

Worth noting, we don't actually use the async variants of any shuffling code in production in lodestar (and I don't see a future where we ever will use it -- we'd rather just lower the entire STF!).

wemeetagain (changes requested):
Just noticed @spiral-ladder 's comments, changing my review to red on his behalf

bindings/napi/shuffle.zig
```diff
@@ -1,6 +1,42 @@
...
+const builtin = @import("builtin");
 const js = @import("zapi:zapi").js;
-const stInnerShuffleList = @import("swap_or_not_shuffle").innerShuffleList;
+const sons = @import("swap_or_not_shuffle");
+
+var gpa: std.heap.DebugAllocator(.{}) = .init;
+const allocator = if (builtin.mode == .Debug) gpa.allocator() else std.heap.c_allocator;
+
+/// Verbatim reference error messages.
+fn shuffleErrorMessage(err: anyerror) [:0]const u8 {
```
GrapeBaBa: Why must do there conversion for the first three error? Seems we don't need next two functions if we don't do this

lodekeeper: Good point. With the async/zapi-enhancement work out of scope, I do not think we need the extra reference-message conversion here.

I simplified the binding so it surfaces the native Zig/zapi error names consistently instead of mapping the first shuffle errors to `@chainsafe/swap-or-not-shuffle` prose messages, and removed the now-unneeded helper path. The tests now assert `InvalidSeedLength` / `InvalidNumberOfRounds` for these cases.

I cannot push directly to `ChainSafe/lodestar-z` (`lodekeeper` only has read permission), so I pushed the signed patch here: https://github.com/lodekeeper/lodestar-z/commit/a8148ba

Verification:
- `PATH=/tmp/zig-0.16.0:$PATH zig fmt --check bindings/napi/shuffle.zig src/swap_or_not_shuffle/root.zig`
- `PATH=/tmp/zig-0.16.0:$PATH sh -c 'git ls-files "*.zig" | xargs zig fmt --check'`
- `PATH=/tmp/zig-0.16.0:$PATH pnpm prepare-mainnet`
- `pnpm exec vitest run bindings/test/shuffle.test.ts` (18 tests)
- `pnpm exec biome check bindings/test/shuffle.test.ts`
- `pnpm lint` exited 0 with existing warnings in `bindings/src/index.d.ts`

Note: plain `zig fmt --check .` still reports unformatted files under downloaded `zig-pkg/` dependency sources after the local build, so I used the tracked-file format check above for the actual repo files.

bindings/napi/shuffle.zig
```diff
@@ -1,6 +1,42 @@
+//! JS binding for the `swap_or_not_shuffle` module. Function names, argument
+//! order and error messages match `@chainsafe/swap-or-not-shuffle`.
```
spiral-ladder:
```suggestion
//! JS binding for the `swap_or_not_shuffle` module.
```

src/swap_or_not_shuffle/root.zig
```diff
@@ -1,87 +1,75 @@
+//! Port of https://github.com/ChainSafe/swap-or-not-shuffle/blob/main/src/lib.rs
+//!
+//! Output values, validation order and error conditions match the reference.
```
spiral-ladder:
```suggestion
//! Port of https://github.com/ChainSafe/swap-or-not-shuffle/blob/main/src/lib.rs
```

src/swap_or_not_shuffle/root.zig
```diff
@@ -1,87 +1,75 @@
...
-        Sha256.hash(self.buf[0..PIVOT_VIEW_SIZE], digest[0..], .{});
-        const slice = std.mem.bytesAsSlice(u64, digest[0..8]);
-        const value = slice[0];
-        return if (native_endian == .big) @byteSwap(value) else value;
+    /// Returns the new pivot. It is "raw" because it has not modulo the list
+    /// size (this must be done by the caller).
+    fn rawPivot(self: *const ShufflingManager) u64 {
+        var digest: [Sha256.digest_length]u8 = undefined;
+        Sha256.hash(self.buf[0..PIVOT_VIEW_SIZE], &digest, .{});
+        return std.mem.readInt(u64, digest[0..8], .little);
```
spiral-ladder: nit, for parity

```suggestion
        return std.mem.readInt(u64, digest[0..@sizeOf(u64)], .little);
```

src/swap_or_not_shuffle/test.zig
spiral-ladder, line 1: these tests look kinda different from what we have in that repo, where are these from?

src/swap_or_not_shuffle/root.zig
```diff
@@ -1,87 +1,75 @@
...
-/// Returns `None` under any of the following conditions:
-///  - `list_size == 0`
-///  - `list_size > 2**24`
-///  - `list_size > usize::MAX / 2`
-/// T should be u32 for Bun binding and ValidatorIndex/u64 for zig application
-pub fn innerShuffleList(comptime T: type, out: []T, seed: []const u8, rounds: u8, forwards: bool) !void {
+/// `T` is `u32` for the JS binding and `ValidatorIndex`/`u64` for Zig callers.
+pub fn innerShuffleList(comptime T: type, out: []T, seed: []const u8, rounds: i32, forwards: bool) ShufflingError!void {
+    // Validation order mirrors the reference: rounds == 0 and short lists
+    // return successfully even when the seed or rounds are invalid.
```
spiral-ladder:
```suggestion
```

src/swap_or_not_shuffle/root.zig
```diff
@@ -184,27 +171,293 @@ pub fn innerShuffleList(comptime T: type, out: []T, seed: []const u8, rounds: u8
...
+    source_buffer: [TOTAL_SIZE]u8,
+    index_count: u32,
+    rounds: u32,
+
+    pub fn init(parent_allocator: Allocator, seed: []const u8, index_count: u32, rounds: u32) !ComputeShuffledIndex {
+        if (seed.len != SEED_SIZE) {
+            return error.InvalidSeedLength;
+        }
+
+        var arena = std.heap.ArenaAllocator.init(parent_allocator);
```
spiral-ladder, line 202: Q: why arena?

src/swap_or_not_shuffle/root.zig
```diff
@@ -184,27 +171,293 @@ pub fn innerShuffleList(comptime T: type, out: []T, seed: []const u8, rounds: u8
...
+    seed: []const u8,
+    active_indices: []const u32,
+    effective_balance_increments: []const u16,
+    rand_byte_count: ByteCount,
+    max_effective_balance: i64,
+    effective_balance_increment: i64,
+    rounds: u32,
+) ![]u32 {
+    // The reference panics on these inputs (`% 0` / division by zero); return
+    // errors instead of invoking unchecked arithmetic.
```
spiral-ladder:
```suggestion
    // reference divergence: error instead of risking invoking unchecked arithmetic
```

src/swap_or_not_shuffle/root.zig
nazarhussain: Documented the usage in the comments. 

Arena seems fit for this use case. 
> The caches below are written until `deinit` and never freed piecemeal,
>  so an arena keeps their many small allocations to one teardown.

src/swap_or_not_shuffle/test.zig
nazarhussain: These tests were not just coming from that repo, it is mix of those and existing shuffling tests.

wemeetagain (approved):
Looks good!

Can you simplify / shorten the PR description before merging to only the most important information?

PR #561, opened 2026-08-13
[do not merge] feat(bindings): asyncAggregateWithRandomness by indices

spiral-ladder:


twoeths (commented):
as a library, we can define a completely new api for this and let lodestar decide

PR #562, opened 2026-08-13
perf(bls): add cache-aware signature verifier

wemeetagain:
## Summary

Add native, cache-aware verification for Lodestar signature sets.

## Implementation details

- `src/bls`
  - Add `Signature.sigValidate`, matching `PublicKey.keyValidate`.
  - Add `verifier.zig` and `VerifySignatureSetsOptions`.
  - Borrow `BatchVerifyItem.message` as `*const SigningRoot`.
  - Return `false` for cryptographic failures.
  - Propagate thread-pool lifecycle errors.

- `src/state_transition`
  - Add fixed-capacity `SignatureSetBatch` and `SameMessageSignatureSetBatch`.
  - Fall back to ordered individual verification.
  - Add `PubkeyCache.aggregateIndices` to support `[]u32` input.

- `bindings`
  - Add the `blsVerifier` N-API module and `@chainsafe/lodestar-z/bls-verifier` export.
  - Support indexed, aggregate-index, and raw-key sets.
  - Throw on malformed inputs, cache misses, or unavailable state.
  - Reuse the verifier from the existing `blst` bindings.

- Bounds
  - General batches: 256 sets.
  - Same-message batches: 128 sets.
  - Aggregate indices: consensus committee limits.

## Why

Keeping validator indices through the native boundary reduces JavaScript/native serialization and reuses the native public-key cache. Native same-message fallback also avoids another worker-queue pass when identifying invalid signatures.

Paired with [ChainSafe/lodestar#9820](https://github.com/ChainSafe/lodestar/pull/9820).

## AI assistance

Written with Codex assistance.


bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+            if (!pubkeys.state.initialized) return error.PubkeyIndexNotInitialized;
+            const index = try uint32(try set.getNamedProperty("index"));
+            break :blk pubkeys.state.cache.getPubkey(io, index) orelse
+                return error.PubkeyIndexNotFound;
+        },
+        .aggregate => blk: {
+            if (!pubkeys.state.initialized) return error.PubkeyIndexNotInitialized;
+            const indices = try uint32Slice(try set.getNamedProperty("indices"));
+            if (indices.len == 0) return error.EmptyIndices;
+            if (indices.len > max_indices_per_set) return error.TooManyIndices;
```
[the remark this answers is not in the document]
wemeetagain: This is good example of why something like #557 may be helpful

bindings/napi/bls_verifier.zig
spiral-ladder, line 1: why do we have `bls_verifier` and `blst_verifier`?

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+            @intFromEnum(SetType.single) => .single,
+            else => return error.InvalidSetType,
+        };
+
+        messages[i] = try parseMessage(set);
+
+        public_keys[i] = (try resolvePublicKey(set, set_type)) orelse
+            return js.Boolean.from(false);
+
+        signatures[i] = (try parseSignature(set)) orelse
```
lodekeeper-z: The early `false` return makes cache-error behavior order-dependent. A batch whose first set has a malformed signature returns `false` without resolving later sets, so a later missing validator index is hidden instead of throwing as the API contract says. I reproduced this at this head: `[malformed(index 0), valid(index 99)]` returned `false`, while reversing the same two sets threw `PubkeyIndexNotFound`. That can classify a local cache desynchronization as an invalid remote signature. Please resolve/validate every set's interface and cache state before returning the cryptographic result, and add an ordering regression test.

lodekeeper-z (commented):
Reviewed the native verifier and companion contract. The core design is coherent and the targeted build/tests pass. I found one additional error-classification issue below. I also confirmed the existing cumulative aggregate-index concern in the open thread: the maximum 256 x 131,072-index call took about 16.8 seconds synchronously in a local probe, so I consider a cumulative bound blocking before merge.

bindings/napi/bls_verifier.zig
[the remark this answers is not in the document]
lodekeeper-z: Correction: the measurements are valid for the direct in-process API, but my merge-blocking classification was not. Per the threat model in #557, I needed to trace a least-privileged attacker through a supported Lodestar path, account for protocol bounds and rate/peer controls, and compare against the existing cryptographic workload. This probe instead constructed the Cartesian maximum through arbitrary same-process input. I withdraw this as a blocker.

bindings/napi/bls_verifier.zig
wemeetagain: Its a poor name, but one is essentially shared utilities between the cache-aware verifier and low-level blst bindings, and the other is the cache-aware verifier.

lodekeeper-z (approved):
Approved. Correction to my earlier review: after applying the repository threat model in #557, I withdraw the cumulative aggregate-work bound as a blocking finding. My probe exercised an arbitrary same-process Cartesian maximum; it did not establish a least-privileged hostile production path, account for protocol and Lodestar rate bounds, or show amplification beyond expected cryptographic work. The measured timing remains useful performance data, not a demonstrated merge blocker. The remaining cache-error ordering note is non-blocking API semantics. Native implementation, companion contract, and targeted verification look good.

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+const blst_bindings = @import("./blst.zig");
+const blst_verifier = @import("./blst_verifier.zig");
+const pubkeys = @import("./pubkeys.zig");
+
+const NativePublicKey = bls.PublicKey;
+const NativeSignature = bls.Signature;
+const BatchVerifyItem = bls.BatchVerifyItem;
+const DST = bls.DST;
+
+const max_verify_sets = 256;
```
spiral-ladder, line 14: would be nice to justify why 256

bindings/napi/blst_verifier.zig
```diff
@@ -0,0 +1,105 @@
...
+}
+
+pub fn ensureNonzeroRandomScalar(io: std.Io, scalar: *[8]u8) !void {
+    if (!std.mem.allEqual(u8, scalar, 0)) return;
+
+    for (0..random_scalar_retries_max) |_| {
+        io.random(scalar);
+        if (!std.mem.allEqual(u8, scalar, 0)) return;
+    }
+    return error.RandomScalarGenerationFailed;
```
spiral-ladder: this seems wrong, i don't think we want to fail if we can't generate random scalars 8 times in a row, should be a while true i think

bindings/napi/blst_verifier.zig
```diff
@@ -0,0 +1,105 @@
...
+
+pub fn parseSignature(bytes: []const u8) ?NativeSignature {
+    if (bytes.len != NativeSignature.COMPRESS_SIZE and
+        bytes.len != NativeSignature.SERIALIZE_SIZE)
+    {
+        return null;
+    }
+
+    var signature = NativeSignature.deserialize(bytes) catch return null;
+    signature.validate(true) catch return null;
```
spiral-ladder: 1. there's two `parseSignature`s, seems like the agent wrote a lot more than necessary
2. i would prefer to inline all these, [`deserialize` ](https://github.com/ChainSafe/lodestar-z/blob/e56efb2ddf0577aefdd2102f06f81416972329ae/src/bls/Signature.zig#L187-L188)already does length checks for example

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+fn parsePublicKey(set: napi.Value) !?NativePublicKey {
+    const bytes = try uint8Slice(try set.getNamedProperty("pubkey"));
+    if (bytes.len != NativePublicKey.COMPRESS_SIZE and
+        bytes.len != NativePublicKey.SERIALIZE_SIZE)
+    {
+        return null;
+    }
+    var public_key = NativePublicKey.deserialize(bytes) catch return null;
+    public_key.validate() catch return null;
+    return public_key;
```
spiral-ladder: this is only used once, lets just do it inline because [`deserialize`](https://github.com/ChainSafe/lodestar-z/blob/e56efb2ddf0577aefdd2102f06f81416972329ae/src/bls/PublicKey.zig#L76-L77) catches the length issues anyway

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+    if (!std.math.isFinite(number) or number < 0 or number > max_u32 or @floor(number) != number) {
+        return error.InvalidUint32;
+    }
+    return @intFromFloat(number);
+}
+
+fn parseMessage(set: napi.Value) ![32]u8 {
+    const message = try uint8Slice(try set.getNamedProperty("message"));
+    if (message.len != 32) return error.InvalidMessageLength;
+    return message[0..32].*;
```
spiral-ladder: only used once, inline?

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+
+    const pool = blst_bindings.state.thread_pool orelse
+        return error.ThreadPoolNotInitialized;
+    const result = try blst_verifier.verifySignatureSets(
+        js.io(),
+        pool,
+        items[0..count],
+        false,
+        false,
+        true,
```
spiral-ladder: i would avoid blind booleans by using a `verifySigSetOpts` struct instead

```zig
/// everything true by default for safety, explicitly use false if you want to turn off validation
const VerifySigSetOpts = struct {
    pks_validate: bool = true,
    sigs_groupcheck: bool = true,
    propagate_pool_shutdown: bool = true,
}
```

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+                false,
+                &exact_message,
+                DST,
+                null,
+                &public_keys[i],
+                false,
+            ) catch break :blk false;
+            break :blk true;
+        };
+        try results.set(@intCast(i), js.Boolean.from(is_valid));
```
spiral-ladder: once `can_aggregate` is false, we can already start validating signatures one by one. We should move this to the else block above rather than do another for loop

spiral-ladder (changes requested):

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+    return blst_verifier.uint8Slice(value);
+}
+
+fn uint32Slice(value: napi.Value) ![]u32 {
+    if (!(try value.isTypedarray())) return error.TypeMismatch;
+    const info = try value.getTypedarrayInfo();
+    if (info.array_type != .uint32) return error.TypeMismatch;
+
+    const byte_ptr: [*]u8 = info.data.ptr;
+    const typed_ptr: [*]u32 = @ptrCast(@alignCast(byte_ptr));
```
GrapeBaBa: It may be panic cause of uncorrect align since zapi code early return before this cast. 

And looks like zapi API updated, may switch to new API. Also suite for `uint8Slice`.

```zig
  fn uint32Slice(value: napi.Value) ![]u32 {
      try js.Uint32Array.validateArg(value);
      return (js.Uint32Array{ .val = value }).toSlice();
  }
```

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+const max_same_message_sets = bls.MAX_AGGREGATE_PER_JOB;
+const max_indices_per_set = preset.MAX_VALIDATORS_PER_COMMITTEE * preset.MAX_COMMITTEES_PER_SLOT;
+
+const SetType = enum(u32) {
+    indexed = 0,
+    aggregate = 1,
+    single = 2,
+};
+
+fn uint8Slice(value: napi.Value) ![]u8 {
```
GrapeBaBa: nit: may no need the wrapper?

bindings/napi/blst_verifier.zig
spiral-ladder: though I guess there is an argument for bounding it just to be safe, non-zero chance that scalar generation bugs out and we dont know if this is a infinite loop, maybe it's not so bad

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+    if (!(try value.isTypedarray())) return error.TypeMismatch;
+    const info = try value.getTypedarrayInfo();
+    if (info.array_type != .uint32) return error.TypeMismatch;
+
+    const byte_ptr: [*]u8 = info.data.ptr;
+    const typed_ptr: [*]u32 = @ptrCast(@alignCast(byte_ptr));
+    return typed_ptr[0..info.length];
+}
+
+fn uint32(value: napi.Value) !u32 {
```
GrapeBaBa: may bump the zapi to use [PR](https://github.com/ChainSafe/zapi/pull/71)

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+            break :blk pubkeys.state.cache.getPubkey(io, index) orelse
+                return error.PubkeyIndexNotFound;
+        },
+        .aggregate => blk: {
+            if (!pubkeys.state.initialized) return error.PubkeyIndexNotInitialized;
+            const indices = try uint32Slice(try set.getNamedProperty("indices"));
+            if (indices.len == 0) return error.EmptyIndices;
+            if (indices.len > max_indices_per_set) return error.TooManyIndices;
+
+            break :blk pubkeys.state.cache.aggregateU32(io, indices) catch |err| switch (err) {
```
spiral-ladder: i would prefer this rather than adding `aggregateU32`

```suggestion
            break :blk pubkeys.state.cache.aggregateIndices(io, u32, indices) catch |err| switch (err) {
```

bindings/napi/bls_verifier.zig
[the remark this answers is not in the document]
spiral-ladder: fwiw my branch does validation like this: https://github.com/ChainSafe/lodestar-z/compare/main...bing/verify-by-index#diff-8e101a8ae444f23fc711a37e880a414376a90e99941d92150f52e72ed1633850R163

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+        },
+        .single => try parsePublicKey(set),
+    };
+}
+
+/// Verify indexed, aggregate, and raw-pubkey signature sets synchronously.
+///
+/// Invalid cryptographic inputs return false. Cache misses and malformed
+/// interface inputs throw so callers do not classify operational failures as
+/// invalid signatures.
```
spiral-ladder: reword ai slop doc comment to be straightforward 😁 

```suggestion
/// Verify indexed, aggregate, and raw-pubkey signature sets synchronously.
///
/// Returns false if verification fails. Throws for malformed inputs and cache misses.
```

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+        messages[i] = try parseMessage(set);
+
+        public_keys[i] = (try resolvePublicKey(set, set_type)) orelse
+            return js.Boolean.from(false);
+
+        signatures[i] = (try parseSignature(set)) orelse
+            return js.Boolean.from(false);
+
+        items[i] = .{
+            .message = messages[i],
```
GrapeBaBa: Looks like don't need middle `messages[i]` to  avoid a copy

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+        bytes.len != NativePublicKey.SERIALIZE_SIZE)
+    {
+        return null;
+    }
+    var public_key = NativePublicKey.deserialize(bytes) catch return null;
+    public_key.validate() catch return null;
+    return public_key;
+}
+
+fn resolvePublicKey(set: napi.Value, set_type: SetType) !?NativePublicKey {
```
spiral-ladder: i feel we should just inline this

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+        false,
+        true,
+    );
+
+    return js.Boolean.from(result);
+}
+
+/// Randomly aggregate and verify indexed signatures over the same message.
+/// Returns one result per input, falling back to individual checks only when
+/// the aggregate check fails.
```
spiral-ladder: one thing i like to know is the context, so adding this might be helpful

```suggestion
/// Randomly aggregate and verify indexed signatures over the same message.
/// 
/// Returns one boolean result per input set in order to account for individual attestations. A result is true if the signature is valid, false otherwise.
/// Throws for invalid input, missing pubkeys or uninitialized pool.
```

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,254 @@
...
+
+    for (0..count) |i| {
+        const is_valid = signature_valid[i] and blk: {
+            signatures[i].verify(
+                false,
+                &exact_message,
+                DST,
+                null,
+                &public_keys[i],
+                false,
```
spiral-ladder: should groupcheck?
```suggestion
                true,
```

bindings/napi/bls_verifier.zig
wemeetagain: Addressed in 748bf8c7, file removed

bindings/napi/bls_verifier.zig
wemeetagain: We are keeping intentional short-circuit semantics

bindings/napi/bls_verifier.zig
wemeetagain: Signatures are already parsed with `Signature.sigValidate(bytes, true)`

bindings/napi/bls_verifier.zig
wemeetagain: yes that would be nice. will leave for a separate PR since next release of zapi will be breaking

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,174 @@
...
+/// verification with individual fallback. Throws on invalid input, cache
+/// errors, or pool unavailability.
+pub fn verifySignatureSetsSameMessage(sets: js.Array, message: js.Uint8Array) !js.Array {
+    const count = try sets.length();
+    if (count > max_same_message_sets) return error.TooManySets;
+
+    const results = js.Array.createWithLength(count);
+    if (count == 0) return results;
+
+    const message_slice = try message.toSlice();
```
GrapeBaBa, line 122: The refactor making message as a pointer seems violate the NAPI v8 memory rule, because there are some js function calls may triggered GC before get the value later. And store the pointer in the item seems also not safety.

lodekeeper-z: Confirmed. The shared message slice was retained across per-set property reads that can execute JavaScript. I added a regression where a signature getter mutates the source message, watched it fail as `[false]`, then fixed the boundary by copying the 32-byte message before further NAPI calls. Since this account cannot push the original branch, the verified fix is in #564. The focused binding suite passes all 9 tests.

GrapeBaBa: Can you check all such kind of code including the existing code? I suspected there are some similar issues.

lodekeeper-z: I audited the BLS NAPI typed-array lifetimes and found three additional affected paths: the mixed-set verifier, `fastAggregateVerify`, and the existing `verifyMultipleAggregateSignatures`. Each retained a message backing-store pointer across a property or array read that can execute JavaScript. I added red/green getter-mutation regressions and native snapshots for all three in #564. The combined focused suites pass all 173 tests, and the ReleaseSafe mainnet binding build passes.

wemeetagain: I don't think issue is GC, by itself, here. Its a synchronous function with a handle scope around the Uint8Array. The handle scope ensures that the Uint8Array doesn't get GC'd. Any synchronous function has the same guarantees.

But there is a real (a contrived, but real nonetheless) re-entrency risk, where an overridden getter of a later object access can modify the Uint8Array, or detach the underlying ArrayBuffer.

Imo these kinds of getters are deliberately contrived and malicious, and obviously not what Lodestar does or what these bindings need to handle (This is all trusted input, lodestar just constructs vanilla objects).

Happy to change it but its worth noting that this is a very low-risk correctness hardening, not a real production bug.

GrapeBaBa: > I don't think issue is GC, by itself, here. Its a synchronous function with a handle scope around the Uint8Array. The handle scope ensures that the Uint8Array doesn't get GC'd. Any synchronous function has the same guarantees.
> 
> But there is a real (a contrived, but real nonetheless) re-entrency risk, where an overridden getter of a later object access can modify the Uint8Array, or detach the underlying ArrayBuffer.
> 
> Imo these kinds of getters are deliberately contrived and malicious, and obviously not what Lodestar does or what these bindings need to handle (This is all trusted input, lodestar just constructs vanilla objects).
> 
> Happy to change it but its worth noting that this is a very low-risk correctness hardening, not a real production bug.

Yes, I noticed most are `get` accessors called here, I am not quite sure if it is possible to trigger GC or not. But current keeping the message pointer in item, seems there are a lot JS function be called throughout the lifecycle, is it same safety?

lodekeeper-z: Yes—the distinction is liveness versus stability. The handle scope keeps the `Uint8Array` reachable, so GC alone should not invalidate it, but it does not snapshot or freeze the backing bytes. Once later `sets.get`, object-property reads, or getters execute JavaScript, that code can mutate (or potentially detach) a previously captured backing store. In `verifyMultipleAggregateSignatures`, an earlier item’s pointer survives all subsequent set/property reads before native verification, so it has that same re-entrancy exposure. #564 removes the dependency by copying each 32-byte message into native-owned `SigningRoot` storage as it is extracted. I would therefore describe this as backing-store stability/re-entrancy hardening, not a demonstrated GC lifetime bug; the regressions in #564 specifically prove mutation across later JS calls.

GrapeBaBa: OK, seems correct, but the zapi comment for typed array seems a little bit inaccurate, so that we don't need https://github.com/ChainSafe/lodestar-z/pull/564.

GrapeBaBa (dismissed):

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,174 @@
...
+const bls = @import("bls");
+const preset = @import("preset").preset;
+const signature_set_verifier = @import("state_transition").signature_set_verifier;
+
+const blst_bindings = @import("./blst.zig");
+const pubkeys = @import("./pubkeys.zig");
+
+// Bound synchronous NAPI work and the fixed stack buffers below. Lodestar's
+// worker jobs normally contain at most 128 sets, so 256 provides headroom while
+// requiring unusually large direct callers to chunk explicitly.
```
spiral-ladder:
```suggestion
/// Bound synchronous NAPI work and the fixed stack buffers below. Lodestar's
/// worker jobs normally contain at most 128 sets, so 256 provides headroom while
/// requiring unusually large direct callers to chunk explicitly.
```

bindings/napi/bls_verifier.zig
```diff
@@ -0,0 +1,174 @@
...
+    return @intFromFloat(number);
+}
+
+/// Verify indexed, aggregate, and raw-pubkey signature sets synchronously.
+///
+/// Returns false on cryptographic failure. Throws for malformed inputs and
+/// cache misses encountered before a result is known.
+pub fn verifySignatureSets(sets: js.Array) !js.Boolean {
+    const count = try sets.length();
+    if (count == 0) return js.Boolean.from(false);
```
spiral-ladder, line 58: semantics nit: would passing an empty set count as cryptographic failure? I guess in some ways it can be interpreted as such (passing in sets of length 0 = nothing to verify = not verified?)

src/bls/ThreadPool.zig
```diff
@@ -254,16 +254,17 @@ const VerifyMultiWorkItem = struct {
 /// This is the multi-threaded version of the same function in `fast_verify.zig`.
 /// Multiple callers may invoke this concurrently — each call owns its own
 /// pairing buffers and job state, workers pull from a shared queue.
+/// Invalid cryptographic inputs return false; pool lifecycle errors propagate.
```
spiral-ladder:
```suggestion
/// Returns false for invalid cryptographic inputs. Propagates pool lifecycle errors.
```

src/bls/verifier.zig
```diff
@@ -0,0 +1,167 @@
...
+const Signature = bls.Signature;
+const SigningRoot = bls.SigningRoot;
+const Pairing = bls.Pairing;
+const ThreadPool = bls.ThreadPool;
+const BatchVerifyItem = bls.BatchVerifyItem;
+const DST = bls.DST;
+const MAX_AGGREGATE_PER_JOB = bls.MAX_AGGREGATE_PER_JOB;
+
+// A correct RNG produces an all-zero 64-bit scalar with probability 2^-64.
+// Keep retries bounded so a broken RNG fails instead of looping forever.
```
spiral-ladder:
```suggestion
/// A correct RNG produces an all-zero 64-bit scalar with probability 2^-64.
/// Keep retries bounded so a broken RNG fails instead of looping forever.
```

spiral-ladder (commented):
bunch of nits

spiral-ladder (approved):
LGTM

GrapeBaBa (approved):
Looks good 👍

PR #563, opened 2026-08-14
feat(shuffle): add PTC sampling

nazarhussain:
## Summary

Adds payload timeliness committee (PTC) sampling to the Zig `swap_or_not_shuffle` module and its JS binding, so ChainSafe/lodestar#9263 can stop computing it in JS.

- Ports `compute_ptc_indices` and `compute_ptc_indices_for_epoch`, whose original implementation is ChainSafe/swap-or-not-shuffle#24 (`453b639b`).
- Exposes a per-slot entry point and a whole-epoch one that derives each slot's seed and returns the committees concatenated.
- Returns explicit errors for degenerate inputs that the reference package would crash or hang on.
- Runs the epoch variant sequentially where the Rust uses rayon; still roughly 287x the naive JS path Lodestar uses today.

## Testing

- Added Zig vectors generated from the naive Lodestar sampler, plus epoch/per-slot consistency and error-path coverage.
- Added JS tests comparing the binding against the naive sampler at full PTC size for one slot and across a full epoch.
- Added per-slot and per-epoch benches alongside the existing shuffle ones.

## Follow-ups

- The epoch variant can be parallelised if its sequential cost ever matters.

### DIFF
--- a/bindings/napi/shuffle.zig
+++ b/bindings/napi/shuffle.zig
@@ -98,3 +98,100 @@ pub fn computeSyncCommitteeIndices(
     defer allocator.free(result);
     return js.Uint32Array.from(result);
 }
+
+pub fn computePtcIndices(
+    seed: js.Uint8Array,
+    indices: js.Uint32Array,
+    effective_balance_increments: js.Uint16Array,
+    ptc_size: js.Number,
+    max_effective_balance_electra: js.Number,
+    effective_balance_increment: js.Number,
+) !js.Uint32Array {
+    const out_array = try js.Uint32Array.alloc(try ptc_size.toU32Exact());
+    try sons.computePtcIndicesInto(
+        try seed.toSlice(),
+        try indices.toSlice(),
+        try effective_balance_increments.toSlice(),
+        max_effective_balance_electra.assertI64(),
+        effective_balance_increment.assertI64(),
+        try out_array.toSlice(),
+    );
+    return out_array;
+}
+
+pub fn computePtcIndicesForEpoch(
+    epoch_seed: js.Uint8Array,
+    start_slot: js.Number,
+    slots_per_epoch: js.Number,
+    shuffling: js.Uint32Array,
+    slot_offsets: js.Uint32Array,
+    effective_balance_increments: js.Uint16Array,
+    ptc_size: js.Number,
+    max_effective_balance_electra: js.Number,
+    effective_balance_increment: js.Number,
+) !js.Uint32Array {
+    const slots = try slots_per_epoch.toU32Exact();
+    const size = try ptc_size.toU32Exact();
+    const out_array = try js.Uint32Array.alloc(@as(usize, slots) * size);
+    try sons.computePtcIndicesForEpochInto(
+        try epoch_seed.toSlice(),
+        try start_slot.toU32Exact(),
+        slots,
+        try shuffling.toSlice(),
+        try slot_offsets.toSlice(),
+        try effective_balance_increments.toSlice(),
+        size,
+        max_effective_balance_electra.assertI64(),
+        effective_balance_increment.assertI64(),
+        try out_array.toSlice(),
+    );
+    return out_array;
+}
+
+/// Fills the caller's `out` array instead of allocating a new one; its length
+/// is the committee size.
+pub fn computePtcIndicesInto(
+    out: js.Uint32Array,
+    seed: js.Uint8Array,
+    indices: js.Uint32Array,
+    effective_balance_increments: js.Uint16Array,
+    max_effective_balance_electra: js.Number,
+    effective_balance_increment: js.Number,
+) !void {
+    try sons.computePtcIndicesInto(
+        try seed.toSlice(),
+        try indices.toSlice(),
+        try effective_balance_increments.toSlice(),
+        max_effective_balance_electra.assertI64(),
+        effective_balance_increment.assertI64(),
+        try out.toSlice(),
+    );
+}
+
+/// Fills the caller's `out` array instead of allocating a new one; it must hold
+/// `slotsPerEpoch * ptcSize` entries.
+pub fn computePtcIndicesForEpochInto(
+    out: js.Uint32Array,
+    epoch_seed: js.Uint8Array,
+    start_slot: js.Number,
+    slots_per_epoch: js.Number,
+    shuffling: js.Uint32Array,
+    slot_offsets: js.Uint32Array,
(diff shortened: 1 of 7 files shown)


bindings/napi/shuffle.zig
```diff
@@ -98,3 +98,51 @@ pub fn computeSyncCommitteeIndices(
...
 }
+
+pub fn computePtcIndices(
+    seed: js.Uint8Array,
+    indices: js.Uint32Array,
+    effective_balance_increments: js.Uint16Array,
+    ptc_size: js.Number,
+    max_effective_balance_electra: js.Number,
+    effective_balance_increment: js.Number,
+) !js.Uint32Array {
```
GrapeBaBa, line 109: Is that possible to use a `*Into` API with `out` arg, seems a little bit effective for these kinds of invocation

nazarhussain: The module already had an out-based inner function, so this was mostly deleting the allocation wrapper.

wemeetagain: I think grapebaba is requesting an additional function(s) to the binding, *Into variants?

src/swap_or_not_shuffle/root.zig
```diff
@@ -455,6 +455,149 @@ pub fn computeSyncCommitteeIndicesElectra(
...
+
+/// Samples one slot's payload timeliness committee into `out`, whose length is
+/// the committee size. Caller owns `out`.
+pub fn computePtcIndicesInto(
+    out: []u32,
+    seed: []const u8,
+    indices: []const u32,
+    effective_balance_increments: []const u16,
+    max_effective_balance_electra: i64,
+    effective_balance_increment: i64,
```
spiral-ladder, line 514: out should be the last param

```suggestion
pub fn computePtcIndicesInto(
    seed: []const u8,
    indices: []const u32,
    effective_balance_increments: []const u16,
    max_effective_balance_electra: i64,
    effective_balance_increment: i64,
    out: []u32,
```

bindings/napi/shuffle.zig
```diff
@@ -85,10 +91,59 @@ pub fn computeSyncCommitteeIndices(
...
     defer allocator.free(result);
     return js.Uint32Array.from(result);
 }
+
+pub fn computePtcIndices(
+    seed: js.Uint8Array,
+    indices: js.Uint32Array,
+    effective_balance_increments: js.Uint16Array,
+    ptc_size: js.Number,
+    max_effective_balance_electra: js.Number,
```
spiral-ladder, line 107: since this is known at comptime, can't we just not provide this as a param? Or are there reasons to provide this as a param

src/swap_or_not_shuffle/test.zig
```diff
@@ -278,3 +278,213 @@ test "getCommitteeIndices rejects out-of-bounds candidate index" {
...
+        out[0..0],
+        good_seed[0..],
+        data.indices[0..],
+        data.increments[0..],
+        MAX_EFFECTIVE_BALANCE_ELECTRA,
+        EFFECTIVE_BALANCE_INCREMENT,
+    );
+}
+
+test "computePtcIndicesForEpochInto rejects invalid slot offsets" {
```
spiral-ladder: i kinda prefer we put all invalid cases into the same test like in test "computePtcIndicesInto rejects invalid inputs gracefully"

src/swap_or_not_shuffle/root.zig
```diff
@@ -455,6 +455,149 @@ pub fn computeSyncCommitteeIndicesElectra(
...
+    const indices_len = indices.len;
+
+    var hash_input = [_]u8{0} ** (SEED_SIZE + 8);
+    @memcpy(hash_input[0..SEED_SIZE], seed);
+
+    var digest: [Sha256.digest_length]u8 = undefined;
+    var count: usize = 0;
+
+    var i: usize = 0;
+    outer: while (true) : (i += 16) {
```
spiral-ladder: i don't really like the while true here and think we should avoid this pattern as much as possible, can't we bound with `count`?

This makes an empty `out` a valid input without causing non-termination since an empty `out` would skip this entirely. with that said we should probably still check `out.len > 0

```suggestion
    outer: while (count < out.len) : (i += 16) {
```

bindings/napi/shuffle.zig
wemeetagain: Really good point, and points to some pain we will have to figure out.

Current lodestar-ts handles both mainnet and minimal without recompilation, rather relies on load-time selection of the preset.

Lodestar-z handles the preset at compile time. So we will have to figure out how we can bundle code for both presets or else deploy separate versions for each preset.

bindings/napi/shuffle.zig
nazarhussain: Added the new variants.

bindings/napi/shuffle.zig
nazarhussain: Suggest to keep it as a parameter for now. 

- `ptcSize` doesn't just size the output buffer — it changes the sampled result. Baking it in means the values a build produces depend on the compiled preset, which is exactly the coupling that makes the mainnet/minimal problem @wemeetagain describes worse: **one binary can serve both presets today, and it couldn't if sizes were compiled in.**

- The caller already has the constant. Lodestar passes `PTC_SIZE` from `@lodestar/params` the same way it passes `SYNC_COMMITTEE_SIZE` and `SHUFFLE_ROUND_COUNT`, so nothing is gained on the JS side.


I would love to handle this comptime preset topic as a separate PR.

wemeetagain (dismissed):

GrapeBaBa (dismissed):

src/swap_or_not_shuffle/root.zig
```diff
@@ -455,6 +455,151 @@ pub fn computeSyncCommitteeIndicesElectra(
     );
 }
 
+/// Fills `out` by sampling from the pre-shuffled `indices`, weighted by
+/// effective balance; one SHA-256 per 16 candidates.
+/// Port of https://github.com/ChainSafe/swap-or-not-shuffle/pull/24 (453b639b).
+/// Callers guarantee `out.len > 0`, otherwise this never returns.
```
spiral-ladder:
```suggestion
```

outdated now that we bound by `count < out.len`

spiral-ladder (dismissed):
lgtm once a doc comment is fixed

spiral-ladder (approved):

PR #572, opened 2026-08-19
test: organize memory-safety checks by module

GrapeBaBa:
## Motivation

Memory-leak, orphan-node, double-free, and OOM-cleanup tests are currently embedded across SSZ and state-transition implementation files. Grouping them by module makes the production files easier to read and gives the memory-safety coverage one clear home.

## Description

- Move 19 SSZ memory-safety tests into `src/ssz/memory_safety_test.zig`.
- Move 2 state-transition memory-safety tests into `src/state_transition/memory_safety_test.zig`.
- Remove the standalone `CachedBeaconState.clone()` leak smoke test because `CachedBeaconState.clone() epoch cache isolation` exercises the same allocator, setup, clone options, and cleanup with stronger assertions.
- Register both files through their existing module test roots.
- Remove only test-only imports made unused by the move.
- Keep all retained test names, bodies, bounds, assertions, and existing module test steps unchanged.

This is a test-only organization and deduplication change. It adds no production API, build step, or CI job.

## AI assistance

The test migration and deduplication were primarily AI-authored and independently reviewed with AI assistance.


spiral-ladder (commented):
What about putting these tests in for example, `list_composite_test.zig`?

We can follow the convention of `<file_name>_test.zig`

wemeetagain (approved):

spiral-ladder (approved):
i kinda think we might want to keep the convention of `<file_name>_test.zig`. but this is not a blocking decision, we can come back to this again in future

PR #575, opened 2026-08-19
chore: deprecate unused blst APIs

spiral-ladder:
These following calls are no longer used by prod lodestar following [ChainSafe/9820](https://github.com/ChainSafe/lodestar/pull/9820) and are redundant exports/impls:

- `aggregateWithRandomness` and `asyncAggregateWithRandomness` N-API exports
- `AggregatePublicKey.aggregateWithRandomness`
- `AggregateSignature.aggregateWithRandomness`

## Rationale

I don't think these would ever be called upstream again, any necessary aggregation (w/ randomness) work will be done internally within our ThreadPool impl. In fact the standalone `aggregateWithRandomness` in `AggregateSignature` and `AggregatePublicKey` were no longer called since we are using MSMs now.

We remove their impl as well as their exports (if any). The pool version of aggregateWithRandomness is still kept since it is used.


GrapeBaBa (approved):
Looks good, almost 1k LOC deleted

wemeetagain (approved):

PR #577, opened 2026-08-20
refactor(shuffle): drop JS signature comments and fix balance param name

nazarhussain:
## Summary

Removes the `/// JS: <signature>` comments from the shuffle binding and fixes a name they had drifted on.

- The comments duplicated `bindings/src/index.d.ts`, which is the contract consumers actually read; the pattern is used in a few small binding files but absent from the larger ones (`BeaconStateView.zig`, `blst.zig`).
- They had already drifted: `computeProposerIndex` and `computeSyncCommitteeIndices` called the balance argument `maxEffectiveBalanceElectra`, but those take an explicit `randByteCount` and the pre-electra path passes `MAX_EFFECTIVE_BALANCE`.
- Renames that parameter to `max_effective_balance` in the binding and module, keeping `_electra` only on the Electra wrappers, where it is accurate.

No behavior change; parameters are positional, so no call site moves.

## Testing

- `zig build test:swap_or_not_shuffle`, `test:state_transition`, bindings build, and the shuffle vitest suite all pass.

### DIFF
--- a/bindings/napi/shuffle.zig
+++ b/bindings/napi/shuffle.zig
@@ -18,8 +18,6 @@ fn byteCountFromJs(rand_byte_count: js.Number) !sons.ByteCount {
     };
 }
 
-/// JS: innerShuffleList(out: Uint32Array, seed: Uint8Array, rounds: number, forwards: boolean): void
-///
 /// Shuffles `out` in place, no allocation.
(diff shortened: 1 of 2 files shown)


GrapeBaBa (approved):
lgtm

PR #580, opened 2026-08-20
fix(shuffle): reject out-of-range numbers instead of coercing

nazarhussain:
## Summary

The shuffle binding read sizes and round counts with `assertU32`/`assertI32`, which coerce through ToUint32/ToInt32 and therefore wrap instead of rejecting.

- A negative `syncCommitteeSize` became ~4.29e9 and asked for a 17 GB allocation rather than throwing.
- A `rounds` value of `2**32 + 10` silently shuffled with 10 rounds, so a caller bug produced a wrong result instead of an error.
- Sizes and round counts now require exact integers, using zapi 4.0's `toU32Exact` plus a small helper for the two signed sites.

`unshuffleList` still forwards in-range negatives to the module, so empty and single-element lists keep returning before rounds validation, matching the reference's validation order.

## Testing

- Added a JS case covering the coercion paths above; existing validation-order cases still pass (a single-element list with a 31-byte seed and `rounds = -1` still returns unchanged).
- `zig build test:swap_or_not_shuffle`, bindings build, shuffle vitest suite, and biome all pass.

### DIFF
--- a/bindings/napi/shuffle.zig
+++ b/bindings/napi/shuffle.zig
@@ -10,6 +10,19 @@ const sons = @import("swap_or_not_shuffle");
 var gpa: std.heap.DebugAllocator(.{}) = .init;
 const allocator = if (builtin.mode == .Debug) gpa.allocator() else std.heap.c_allocator;
 
+/// `assertI32`/`assertU32` coerce with ToInt32/ToUint32, which wrap: a caller
+/// passing 2**32 + 10 would silently be given 10. Require an exact integer so
+/// out-of-range values are rejected instead.
+fn exactI32(value: js.Number) !i32 {
+    const f = try value.toF64();
+    if (!std.math.isFinite(f) or @trunc(f) != f or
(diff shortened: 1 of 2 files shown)


bindings/napi/shuffle.zig
```diff
@@ -10,6 +10,19 @@ const sons = @import("swap_or_not_shuffle");
 var gpa: std.heap.DebugAllocator(.{}) = .init;
 const allocator = if (builtin.mode == .Debug) gpa.allocator() else std.heap.c_allocator;
 
+/// `assertI32`/`assertU32` coerce with ToInt32/ToUint32, which wrap: a caller
+/// passing 2**32 + 10 would silently be given 10. Require an exact integer so
+/// out-of-range values are rejected instead.
+fn exactI32(value: js.Number) !i32 {
```
GrapeBaBa, line 16: Why this function not like `toU32Exact` into the zapi code?

spiral-ladder: +1, can we use `toU32Exact`? Or implement a similar `toI32Exact`

spiral-ladder (changes requested):

PR #582, opened 2026-08-20
refactor(bindings): use owned typed arrays for BLS outputs

GrapeBaBa:
## Motivation

zapi 4.0 provides `OwnedUint8Array` for transferring allocator-owned native memory to JavaScript without copying. The existing BLS byte-output path first serialized into a stack array, then copied it into separately allocated external ArrayBuffer storage through `Uint8Array.fromExternal()`.

Using the owned transfer API makes the allocation, ownership transfer, failure handling, and finalizer cleanup one explicit lifecycle. This is primarily a memory-lifecycle improvement. Removing the intermediate copy is a secondary benefit.

## Description

- Allocate exact-size native buffers for compressed and serialized public keys, signatures, and secret keys.
- Let BLST write directly into those buffers.
- Return `OwnedUint8Array` values so zapi transfers ownership to the JavaScript ArrayBuffer finalizer.
- Preserve the existing JavaScript `Uint8Array` API and propagate `NoExternalBuffersAllowed` without adding a copy fallback.

### DIFF
--- a/bindings/napi/blst.zig
+++ b/bindings/napi/blst.zig
@@ -147,13 +147,22 @@ pub const PublicKey = struct {
     }
 
     /// Serializes this public key to bytes.
-    pub fn toBytes(self: *const PublicKey, compress: ?js.Boolean) !js.Uint8Array {
+    pub fn toBytes(self: *const PublicKey, compress: ?js.Boolean) !js.OwnedUint8Array {
+        const c_allocator = js.allocator();
+
         if (try boolOrDefault(compress, true)) {
-            const bytes = self.raw.compress();
-            return js.Uint8Array.fromExternal(bytes[0..]);
+            const bytes = try c_allocator.alloc(u8, NativePublicKey.COMPRESS_SIZE);
+            errdefer c_allocator.free(bytes);
+
+            bls.c.blst_p1_affine_compress(bytes.ptr, &self.raw.point);
+            return js.OwnedUint8Array.fromOwnedSlice(c_allocator, bytes);
         }
-        const bytes = self.raw.serialize();
-        return js.Uint8Array.fromExternal(bytes[0..]);
+
+        const bytes = try c_allocator.alloc(u8, NativePublicKey.SERIALIZE_SIZE);
+        errdefer c_allocator.free(bytes);
+
+        bls.c.blst_p1_affine_serialize(bytes.ptr, &self.raw.point);
+        return js.OwnedUint8Array.fromOwnedSlice(c_allocator, bytes);
     }
 
     pub fn toHex(self: *const PublicKey, compress: ?js.Boolean) !js.String {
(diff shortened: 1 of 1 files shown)


nazarhussain (changes requested):
#### Review

The lifecycle reasoning in the description is sound, and I verified the change is correct: byte-identical output on all five paths. 

My one concern is that the measured benefit isn't there, and the memory side is
slightly worse. See the steps as I understood. 

**`main` branch (`fromExternal`), per call:**
1. one `malloc` for the buffer (inside `dupe`)
2. one 48-byte `memcpy` from the stack array
3. register an external ArrayBuffer + finalizer (the length is passed as the finalize hint via `@ptrFromInt`, so no allocation)

**This PR (`OwnedUint8Array`), per call:**
1. one `malloc` for the buffer
2. no `memcpy` — BLST writes straight into it 
3. **a second `malloc`**: `OwnedTypedArray.intoValue` calls `moveToHeap`, which does `allocator.create(Self)` to heap-allocate the owner struct (`{allocator, slice}` = 32 bytes)
4. register an external ArrayBuffer + finalizer

So we trade a 48-byte `memcpy` for an extra heap allocation and free. A 48-byte `memcpy` is a couple of nanoseconds; a `malloc`/`free` pair is not.

The performance measurements are: 

| op | main | this PR |
| --- | --- | --- |
| `pk.toBytes(true)` 48 B | 525.8 | 514.5 |
| `pk.toBytes(false)` 96 B | 551.7 | 540.1 |
| `sig.toBytes(true)` 96 B | 583.9 | 551.4 |
| `sig.toBytes(false)` 192 B | 619.0 | 612.7 |
| `sk.toBytes()` 32 B | 470.1 | 487.7 |
| `pk.toHex(true)` | 332.7 | 336.4 |

RSS after 300k calls with a forced GC, three consecutive runs each:

| | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| main | 72.8 MB | 72.8 MB | 72.8 MB |
| this PR | 82.0 MB | 82.1 MB | 82.0 MB |

That +9.2 MB is reproducible to 0.1 MB, and it matches the extra allocation above:
300,000 × 32 bytes ≈ 9.6 MB.

---- 

#### Suggestion

For small fixed-size outputs like these, `js.Uint8Array.alloc(N)` + writing into `try result.toSlice()` — the pattern already used in [`BeaconStateView.zig:229-236`](bindings/napi/BeaconStateView.zig#L229-L236) avoids native allocation entirely, V8 owns the buffer, no finalizer, no external-ArrayBuffer support needed.

I tried above suggestion and here is the comparison. 


| op | main | this PR | alloc+toSlice |
| --- | --- | --- | --- |
| `pk.toBytes(true)` 48 B | 525.8 | 514.5 | **331** | 
| `pk.toBytes(false)` 96 B | 551.7 | 540.1 | **354** | 
| `sig.toBytes(true)` 96 B | 583.9 | 551.4 | **421** | 
| `sig.toBytes(false)` 192 B | 619.0 | 612.7 | **402** | 
| `sk.toBytes()` 32 B | 470.1 | 487.7 | **267** | 
| `pk.toHex(true)` | 332.7 | 336.4 | **326** | 

The performance increase feels like sufficient enough to use that pattern. It also keeps the return type as `js.Uint8Array`, so TS sides needs no change.

PR #584, opened 2026-08-21
refactor(pubkeys): drop deserialized lookup wrappers

lodekeeper-z:
## Summary

- remove the unused deserialized `PubkeyCache.get()` and `getOrThrow()` bindings
- drop the JS-level `PublicKey` object map and its invalidation paths
- route single-key aggregation through the native cache like multi-key aggregation
- keep cache tests on the compressed-byte API

Follow-up to https://github.com/ChainSafe/lodestar/pull/9820#discussion_r3827240760

## Verification

- `pnpm prepare-mainnet`
- `pnpm vitest run bindings/test/pubkeys.test.ts bindings/test/worker-isolation.test.ts` (25 passed)
- `pnpm test` (200 passed; fixture-dependent BeaconStateView/teardown tests blocked because `fixtures/era/mainnet-01628-47ac89fb.era` is not present)
- `zig fmt --check bindings/napi/pubkeys.zig`
- `pnpm lint` (passes with pre-existing warnings in `bindings/src/index.d.ts`)
- `git diff --check`

The regression asserting that deserialized lookup helpers are absent was run before implementation and failed as expected, then passed after the change.

### DIFF
--- a/bindings/src/pubkeys.d.ts
+++ b/bindings/src/pubkeys.d.ts
@@ -1,10 +1,6 @@
 import type {PublicKey} from "./blst.js";
 
 export interface PubkeyCache {
-  /** Get a cached, deserialized PublicKey. */
-  get(index: number): PublicKey | undefined;
-  /** Same as get(), but throws if the index is not in the cache */
-  getOrThrow(index: number): PublicKey;
   /** Get the cached 48-byte compressed pubkey bytes without materializing a PublicKey wrapper. */
   getPubkeyBytes(index: number): Uint8Array | undefined;
   /** Same as getPubkeyBytes(), but throws if the index is not in the cache. */
--- a/bindings/src/pubkeys.d.ts
+++ b/bindings/src/pubkeys.d.ts
@@ -1,6 +1,10 @@
 import type {PublicKey} from "./blst.js";
 
 export interface PubkeyCache {
+  /** Get a cached, deserialized PublicKey. */
+  get(index: number): PublicKey | undefined;
+  /** Same as get(), but throws if the index is not in the cache */
+  getOrThrow(index: number): PublicKey;
   /** Get the cached 48-byte compressed pubkey bytes without materializing a PublicKey wrapper. */
   getPubkeyBytes(index: number): Uint8Array | undefined;
   /** Same as getPubkeyBytes(), but throws if the index is not in the cache. */
--- a/bindings/src/pubkeys.js
+++ b/bindings/src/pubkeys.js
@@ -4,6 +4,18 @@ const native = bindings.pubkeys;
 
 /** @type {import("./pubkeys.d.ts").PubkeyCache} */
 export const pubkeyCache = {
+  get(index) {
+    return native.get(index);
+  },
+
+  getOrThrow(index) {
+    const pk = pubkeyCache.get(index);
+    if (pk === undefined) {
+      throw Error(`pubkeyCache: index ${index} not found`);
+    }
+    return pk;
+  },
+
   getPubkeyBytes(index) {
     return native.getPubkeyBytes(index);
   },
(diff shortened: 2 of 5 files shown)


bindings/src/pubkeys.d.ts
```diff
@@ -1,10 +1,6 @@
 import type {PublicKey} from "./blst.js";
 
 export interface PubkeyCache {
-  /** Get a cached, deserialized PublicKey. */
-  get(index: number): PublicKey | undefined;
-  /** Same as get(), but throws if the index is not in the cache */
-  getOrThrow(index: number): PublicKey;
   /** Get the cached 48-byte compressed pubkey bytes without materializing a PublicKey wrapper. */
```
spiral-ladder, line 8: these should still remain, the only thing i want dropped is the memoization map `pkCache` in `pubkeys.js`

spiral-ladder (changes requested):

bindings/src/pubkeys.d.ts
lodekeeper-z: Fixed in 6eb97bd9. I restored `get`/`getOrThrow` and the native lookup path, and narrowed the change to removing only the JS `pkCache` memoization map. The focused pubkey binding suite passes (23/23), and Biome passes on the changed JS/test files.

bindings/src/pubkeys.js
```diff
@@ -2,19 +2,10 @@ import bindings from "./bindings.js";
...
 export const pubkeyCache = {
   get(index) {
-    let pk = pkCache.get(index);
-    if (pk !== undefined) return pk;
-    pk = native.get(index);
-    if (pk !== undefined) {
-      pkCache.set(index, pk);
-    }
-    return pk;
+    return native.get(index);
```
spiral-ladder, line 8: keep the set

spiral-ladder (changes requested):

bindings/src/pubkeys.js
lodekeeper-z: Do you mean restoring a `set(index, pubkey)` alias alongside `append`, or keeping the deleted `pkCache.set(index, pk)` call? The latter requires retaining the JS memoization map, which conflicts with the earlier request to drop only `pkCache`. This PR does not otherwise remove the native write path (`append`).

spiral-ladder: misread, shuould be good

spiral-ladder (approved):

bindings/test/pubkeys.test.ts
```diff
@@ -85,10 +85,12 @@ describe("pubkeys", () => {
...
   });
 
-  it("get caches deserialized values", () => {
+  it("get returns fresh deserialized values without memoizing wrappers", () => {
     const pk1 = pubkeyCache.getOrThrow(0);
     const pk2 = pubkeyCache.getOrThrow(0);
-    expect(pk1).toBe(pk2);
+    expect(pk1).not.toBe(pk2);
+    expect(pk1.toBytes()).toEqual(keypairs[0].pubkeyBytes);
+    expect(pk2.toBytes()).toEqual(keypairs[0].pubkeyBytes);
```
spiral-ladder, line 93: how come this is different?

lodekeeper-z: The difference is intentional because this PR removes only the JS `pkCache` map. Previously, repeated `getOrThrow(0)` calls returned the exact same memoized JS `PublicKey`, so the test asserted `toBe`. Now each call goes through `native.get(0)` and produces a fresh wrapper, so `not.toBe` proves the JS memoization is gone; the two byte assertions prove both wrappers still represent the same cached native pubkey.

wemeetagain (approved):

PR #585, opened 2026-08-21
chore(docs): document bls/pk cache trust assumptions

spiral-ladder:
Some of the trust assumptions were not/poorly documented. Let's make this contract more clear to JS consumers (lodestar)

Also add a test to cover the invalid signature set cases.


src/state_transition/cache/pubkey_cache.zig
```diff
@@ -1,3 +1,24 @@
...
+//! 1. Lookup by validator index or compressed public key.
+//! 2. Aggregation by validator index.
+//! 3. Bulk synchronization from a validator list.
+//!
+//! # Safety
+//!
+//! Decoding rejects malformed compressed encodings and off-curve points. It
+//! does not reject cryptographically invalid keys.
+//!
+//! Before cache insertion, each population path must establish these missing properties:
```
spiral-ladder:
```suggestion
//! Before cache insertion, each population path must establish these properties:
```

wemeetagain (approved):

PR #595, opened 2026-08-25
fix(ssz): free variable list offsets on decode errors

GrapeBaBa:
## Motivation

VariableList decoding allocates a temporary offset array before validating every later offset. If a later offset was malformed, the array was abandoned when the validation error propagated.

## Description

- Free the temporary VariableList offset array on later decode errors.
- Add a malformed-offset regression at the value deserialization boundary.

### DIFF
--- a/src/ssz/memory_safety_test.zig
+++ b/src/ssz/memory_safety_test.zig
@@ -22,6 +22,8 @@ const Checkpoint = FixedContainerType(struct {
 
 test "VariableList deserializeFromBytes should free offsets on malformed later offset" {
     const ListType = VariableListType(ByteListType(8), 4);
+    // Offset 8 declares two elements, so decoding allocates the offsets array.
+    // Offset 4 then moves backward, triggering offsetNotIncreasing after allocation.
     const serialized = [_]u8{ 8, 0, 0, 0, 4, 0, 0, 0 };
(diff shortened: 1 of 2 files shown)


src/ssz/memory_safety_test.zig
```diff
@@ -8,16 +8,37 @@ const ArmOnSizeAllocator = @import("testing_allocators").ArmOnSizeAllocator;
...
 const Checkpoint = FixedContainerType(struct {
     epoch: UintType(64),
     root: ByteVectorType(32),
 });
 
+test "VariableList deserializeFromBytes should free offsets on malformed later offset" {
+    const ListType = VariableListType(ByteListType(8), 4);
+    // Offset 8 declares two elements, so decoding allocates the offsets array.
+    // Offset 4 then moves backward, triggering offsetNotIncreasing after allocation.
+    const serialized = [_]u8{ 8, 0, 0, 0, 4, 0, 0, 0 };
```
spiral-ladder, line 27: 👍

spiral-ladder (approved):
lgtm, something that is out of scope that we can do is also make `readVariableOffsets` not a pub fn since it is used internally only, we can make that another PR

wemeetagain (approved):

PR #597, opened 2026-08-25
fix(state-transition): clean up upgraded payload headers

GrapeBaBa:
## Motivation

Capella and Deneb state upgrades clone `latest_execution_payload_header.extra_data` into a temporary value. `setLatestExecutionPayloadHeader()` copies that value into the persistent Merkle tree, but the temporary header was never deinitialized, leaking non-empty `extra_data` buffers.

## Description

- Deinitialize the temporary upgraded payload header on success and error paths.
- Remove the misleading Deneb ownership-transfer comment.
- Preserve the existing clone and TreeView ownership contracts.
- Add a production-boundary regression covering both fork upgrades.

### DIFF
--- a/src/state_transition/memory_safety_test.zig
+++ b/src/state_transition/memory_safety_test.zig
@@ -17,7 +17,6 @@ const upgradeStateToCapella = @import("slot/upgrade_state_to_capella.zig").upgra
 const upgradeStateToDeneb = @import("slot/upgrade_state_to_deneb.zig").upgradeStateToDeneb;
 
 test "upgradeStateToCapella and upgradeStateToDeneb should release temporary payload headers" {
-    // std.testing.allocator makes the test fail on leaks even without an explicit expect.
     const allocator = std.testing.allocator;
     const chain_config = if (active_preset == .mainnet)
         config.mainnet.chain_config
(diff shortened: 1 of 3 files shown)


src/state_transition/memory_safety_test.zig
```diff
@@ -1,11 +1,87 @@
...
+const EpochCache = @import("cache/epoch_cache.zig").EpochCache;
+const PubkeyCache = @import("cache/pubkey_cache.zig").PubkeyCache;
 const deserializeContainerOverrideFieldsWithRanges =
     @import("ssz_container.zig").deserializeContainerOverrideFieldsWithRanges;
 const processRewardsAndPenalties =
     @import("epoch/process_rewards_and_penalties.zig").processRewardsAndPenalties;
+const upgradeStateToCapella = @import("slot/upgrade_state_to_capella.zig").upgradeStateToCapella;
+const upgradeStateToDeneb = @import("slot/upgrade_state_to_deneb.zig").upgradeStateToDeneb;
+
+test "upgradeStateToCapella and upgradeStateToDeneb should release temporary payload headers" {
```
spiral-ladder, line 19: nit: im starting to feel like these should be in `<file_name>_test.zig` instead of here

spiral-ladder (approved):
lgtm

PR #598, opened 2026-08-25
refactor(ssz): rewrite offset iteration for variable element iteration

spiral-ladder:
While reviewing https://github.com/ChainSafe/lodestar-z/pull/595 it made me rethink about offsets.

Replace `OffsetIterator` and `readVariableOffsets` with `VariableElementIterator`, which yields borrowed element slices in one pass without allocating. Initialization checks the first offset and collection bounds; each subsequent offset is checked before its element slice is returned. Equal offsets remain valid for empty elements.

Use the iterator across validation, deserialization, serialized hashing, and tree construction. Add cleanup for partially decoded list and vector values when iteration or element decoding fails.


src/ssz/type/list.zig
```diff
@@ -761,82 +761,46 @@ pub fn VariableListType(comptime ST: type, comptime _limit: comptime_int) type {
...
-                const first_offset = try iterator.next();
-                const len = first_offset / 4;
-
-                var curr_offset = first_offset;
-                var prev_offset = first_offset;
-                while (iterator.pos < len) {
-                    prev_offset = curr_offset;
-                    curr_offset = try iterator.next();
-
-                    try Element.serialized.validate(data[prev_offset..curr_offset]);
```
spiral-ladder, line 810: imo we should never have to do things like this with an iterator api, the new iterator is much cleaner

src/ssz/type/list.zig
```diff
@@ -761,82 +761,46 @@ pub fn VariableListType(comptime ST: type, comptime _limit: comptime_int) type {
...
-            while (iterator.pos < len) {
-                offsets[iterator.pos] = try iterator.next();
+            var i: usize = 0;
+            while (elements.next()) |element_bytes| : (i += 1) {
+                try Element.deserializeFromBytes(allocator, element_bytes, &out.items[i]);
             }
-            offsets[len] = @intCast(data.len);
-
-            return offsets;
+            std.debug.assert(i == len);
```
spiral-ladder, line 778: use post-asserts bc if they're not eq at this point, it's a implementation error

src/ssz/type/variable_element_iterator.zig
```diff
@@ -0,0 +1,233 @@
...
+        index: usize,
+        start: usize,
+
+        const Self = @This();
+
+        /// Initializes an iterator and validates the complete SSZ offset table.
+        ///
+        /// Returns an error if the encoding has invalid offsets or violates the list or vector
+        /// bounds of `ST`.
+        pub fn init(data: []const u8) !Self {
```
spiral-ladder, line 82: this is pretty much doing what the offset iterator does while it iterates, but at init, so the code is roughly similar

src/ssz/type/list.zig
```diff
@@ -761,83 +761,53 @@ pub fn VariableListType(comptime ST: type, comptime _limit: comptime_int) type {
...
 
             pub fn length(data: []const u8) !usize {
-                if (data.len == 0) {
-                    return 0;
-                }
-                var iterator = OffsetIterator(Self).init(data);
-                return try iterator.firstOffset() / 4;
+                var elements = try VariableElementIterator(Self).init(data);
+                const len = elements.len;
+                while (try elements.next()) |_| {}
```
wemeetagain: this line can be removed? and same in other length methods

wemeetagain (approved):

PR #604, opened 2026-08-26
fix(ssz): reject invalid list growth lengths

GrapeBaBa:
## Motivation

Basic and composite list TreeViews did not handle invalid caller-provided lengths consistently. `growTo()` asserted when asked to shrink and accepted lengths above the SSZ type limit, despite exposing an error-returning public API.

## Description

- Return `error.InvalidLength` when the requested length would shrink the list.
- Return `error.LengthOverLimit` before publishing a length above the SSZ limit.
- Extend the existing basic and composite mutation tests with both error cases.

The implementation and regression tests were primarily AI-authored and independently reviewed with AI assistance.

### DIFF
--- a/src/ssz/tree_view/list_basic.zig
+++ b/src/ssz/tree_view/list_basic.zig
@@ -93,7 +93,7 @@ pub fn ListBasicTreeView(comptime ST: type) type {
         /// Grows the list; new positions read as zero. Shrinking must go through sliceTo —
         /// a bare length cut would leave stale chunk data in the merkleized root.
         pub fn growTo(self: *Self, new_length: usize) !void {
-            std.debug.assert(new_length >= self._len);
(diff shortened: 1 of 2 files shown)


src/ssz/tree_view/list_basic.zig
```diff
@@ -93,7 +93,8 @@ pub fn ListBasicTreeView(comptime ST: type) type {
         /// Grows the list; new positions read as zero. Shrinking must go through sliceTo —
         /// a bare length cut would leave stale chunk data in the merkleized root.
         pub fn growTo(self: *Self, new_length: usize) !void {
-            std.debug.assert(new_length >= self._len);
+            if (new_length < self._len) return error.InvalidLength;
```
GrapeBaBa, line 96: Thought here using assert is not correct

wemeetagain (approved):

PR #605, opened 2026-08-26
do not merge: fallible one-pass variable elem iteration

spiral-ladder:
for context see #598

### DIFF
--- a/src/ssz/type/list.zig
+++ b/src/ssz/type/list.zig
@@ -766,9 +766,13 @@ pub fn VariableListType(comptime ST: type, comptime _limit: comptime_int) type {
 
             try out.resize(allocator, len);
             @memset(out.items[0..len], Element.default_value);
+            errdefer {
+                Self.deinit(allocator, out);
+                out.* = default_value;
+            }
 
             var i: usize = 0;
-            while (elements.next()) |element_bytes| : (i += 1) {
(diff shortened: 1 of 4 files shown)


src/ssz/type/list.zig
```diff
@@ -777,13 +781,16 @@ pub fn VariableListType(comptime ST: type, comptime _limit: comptime_int) type {
...
+                while (try elements.next()) |element_bytes| {
                     try Element.serialized.validate(element_bytes);
                 }
             }
 
             pub fn length(data: []const u8) !usize {
-                return (try VariableElementIterator(Self).init(data)).len;
+                var elements = try VariableElementIterator(Self).init(data);
+                const len = elements.len;
+                while (try elements.next()) |_| {}
```
spiral-ladder, line 792: idk why this is here, codex was hallucinating

src/ssz/type/vector.zig
```diff
@@ -512,8 +512,13 @@ pub fn VariableVectorType(comptime ST: type, comptime _length: comptime_int) typ
             }
 
             var elements = try VariableElementIterator(Self).init(data);
+            errdefer {
+                Self.deinit(allocator, out);
+                out.* = default_value;
+            }
```
spiral-ladder, line 518: this is what's not as nice if we dont do upfront validation: failure halfway means we have to cleanup here in an errdefer

PR #611, opened 2026-08-27
test: verify allocation failure coverage with standard checks

GrapeBaBa:
## Motivation

Fixed allocation indexes can silently move away from the intended cleanup boundary when allocation order changes. Replace suitable tests with allocation sweeps while retaining their original error-path checks.

## Description

Use `std.testing.checkAllAllocationFailures` for 23 scenarios across SSZ, fork types, state transition, clock, and BlockStateCache. Retain explicit target-operation OOM, ownership, rollback, pool-slot, double-free, and forced resize/remap-failure checks where applicable.

Each replacement was checked against the original test using targeted negative modifications: 25 comparisons fail under both versions, and the replacement passes once the normal implementation is restored. The two EpochShuffling failure stages and both bulk-read cleanup stages were checked separately.

Keep the original PMT same-pool recovery test: a free-list corruption probe exposed a coverage loss when OOM and subsequent reuse were separated. Specialized retry, no-allocation, and CoW tests remain unchanged; the sync-committee rotation sweep also awaits #631. No production behavior changes.

Validation: SSZ (346), PMT (63), fork types (16), clock (41), beacon-node (36), mainnet state-transition (153), and focused minimal EpochShuffling tests pass, as do tidy, lint, and changed-file formatting. Lint has 10 existing warnings; repository-wide formatting reports existing issues in ignored dependency sources.

The test changes were primarily AI-authored.


src/ssz/memory_safety_test.zig
```diff
@@ -377,69 +391,50 @@ test "TreeView composite list set(index, ownedView) - failed set leaves the elem
...
+            else => {
+                elem_view.deinit();
+                return err;
+            },
         }
-        try std.testing.expect(!oom.double_free);
-    }
-}
-
-test "TreeView composite list clone(transfer_cache) - OOM does not double-free cached children" {
```
nazarhussain: Would be better to rewrite this test rather completely removing it.

src/ssz/memory_safety_test.zig
```diff
@@ -318,26 +317,33 @@ test "TreeView composite list setValue - OOM does not double-free the element vi
...
         defer view.deinit();
 
-        // We only care that no path double-frees; the OOM itself is expected.
-        view.setValue(0, &newval) catch {};
-        try std.testing.expect(!oom.double_free);
+        // The setValue path creates a child view before storing it in the list cache. Let it
+        // allocate, then fail the cache reservation. setValue must free the child once. Teardown
+        // must not free it again.
+        oom.failing.fail_index = oom.failing.alloc_index + 1;
+        try std.testing.expectError(error.OutOfMemory, view.setValue(0, &newval));
```
nazarhussain: The test fixed  allocation is the risky one'; if the code's allocation order shifts, the test keeps passing while guarding the wrong spot. 

Instead of betting on one magic number, better to loop: fail allocation #1, run the scenario, check nothing double-freed.

nazarhussain (changes requested):

src/beacon_node/chain/state_cache/block_state_cache.zig
```diff
@@ -794,21 +794,17 @@ test "BlockStateCache add - insert/prune/duplicate paths free the owned state ex
...
-        var failing = std.testing.FailingAllocator.init(testing.allocator, .{ .fail_index = fail_index });
-        if (BlockStateCache.init(failing.allocator(), .{ .max_states = 4 })) |cache| {
-            var c = cache;
-            c.deinit();
-            break;
-        } else |err| {
-            try testing.expectEqual(error.OutOfMemory, err);
+    var saw_operation_oom = false;
+    try testing.checkAllAllocationFailures(testing.allocator, struct {
+        fn run(allocator: std.mem.Allocator, saw_oom: *bool) !void {
```
nazarhussain: The flag is used in two different ways 


`errdefer saw_oom.* = true` vs `errdefer |err| saw_oom.* = err == error.OutOfMemory)`

We should make it consistent across all files.

src/beacon_node/chain/state_cache/block_state_cache.zig
```diff
@@ -794,21 +794,17 @@ test "BlockStateCache add - insert/prune/duplicate paths free the owned state ex
...
-    var fail_index: usize = 0;
-    while (true) : (fail_index += 1) {
-        var failing = std.testing.FailingAllocator.init(testing.allocator, .{ .fail_index = fail_index });
-        if (BlockStateCache.init(failing.allocator(), .{ .max_states = 4 })) |cache| {
-            var c = cache;
-            c.deinit();
-            break;
-        } else |err| {
-            try testing.expectEqual(error.OutOfMemory, err);
+    var saw_operation_oom = false;
```
nazarhussain, line 797: `checkAllAllocationFailures` already enforces OOM-on-injection, SwallowedOutOfMemoryError and MemoryLeakDetected. So the `saw_operations_oom` is mostly redundant. 


The flag is used in two different ways `errdefer saw_oom.* = true` vs `errdefer |err| saw_oom.* = err == error.OutOfMemory)`. If we keep this pattern we should make it consistent across all files.

src/beacon_node/chain/state_cache/block_state_cache.zig
```diff
@@ -794,21 +794,17 @@ test "BlockStateCache add - insert/prune/duplicate paths free the owned state ex
...
+                saw_oom.* = err == error.OutOfMemory;
+            }
+            var cache = try BlockStateCache.init(allocator, .{ .max_states = 4 });
+            defer cache.deinit();
         }
-    }
-    try testing.expect(fail_index > 0); // sanity: init does allocate, so the sweep hit OOM at least once
+    }.run, .{&saw_operation_oom});
+    try std.testing.expect(saw_operation_oom);
 }
```
nazarhussain, line 808: How about this version of the test. 


```zig

fn countAllocations(comptime test_fn: anytype, extra_args: anytype) !usize {
    var failing = std.testing.FailingAllocator.init(testing.allocator, .{});
    try @call(.auto, test_fn, .{failing.allocator()} ++ extra_args);
    return failing.alloc_index;
}

test "BlockStateCache init releases its reservations on OOM at every allocation point" {
    const Body = struct {
        fn initThenDeinit(allocator: Allocator) !void {
            var cache = try BlockStateCache.init(allocator, .{ .max_states = 4 });
            cache.deinit();
        }
    };

    try testing.expect(try countAllocations(Body.initThenDeinit, .{}) > 0);
    try testing.checkAllAllocationFailures(testing.allocator, Body.initThenDeinit, .{});
}
```


The function `countAllocations` can be reused across all the tests. And rest of the codebase will be simplified.

nazarhussain (changes requested):

PR #614, opened 2026-08-28
fix(ssz): keep clone destinations deinit-safe

GrapeBaBa:
## Motivation

Variable list cloning exposed the full destination length before every child was initialized. If a later child clone ran out of memory, the destination contained an undefined tail and could not be safely deinitialized by its caller.

## Description

- Define the caller-owned clone contract: initialize the destination with its SSZ default and clean it up after success or error.
- Add destination-typed `cloneInto()` for structural container, vector, and list clones.
- Publish each list element as its destination default before recursively cloning it.
- Migrate cross-fork execution payload header clones and add a deterministic second-child OOM regression.

The implementation and regression test were primarily AI-authored and independently reviewed with AI assistance.


wemeetagain (dismissed):
looks good

src/ssz/type/bit_list.zig
```diff
@@ -286,10 +286,19 @@ pub fn BitListType(comptime _limit: comptime_int) type {
...
-        ///
-        /// Caller owns the memory.
+        /// The caller initializes `out` with `default_value`; this uses `cloneInto`'s contract.
         pub fn clone(allocator: std.mem.Allocator, value: *const Type, out: *Type) !void {
+            return cloneInto(@This(), allocator, value, out);
+        }
+
+        /// The caller initializes `out` with `DestinationST.default_value` and deinitializes it
+        /// after success or error. Errors leave `out` safe to deinitialize.
+        pub fn cloneInto(
```
spiral-ladder, line 296: Should this be a pub fn?

GrapeBaBa: That is designed for non same type clone

wemeetagain (approved):

src/ssz/type/bit_list.zig
wemeetagain: Yes, looks like a new operation for cloning from similar, but distinct types. Used in STF for things like migrating between forks. Eg upgrading the ExecutionPayloadHeader from one fork to another.

PR #615, opened 2026-08-28
perf(ssz): hash dirty basic field roots directly

GrapeBaBa:
## Motivation

Dirty basic field-root reads materialize a temporary PMT node only to copy its root into the view cache. This adds a pool allocation, a `PoolExhausted` failure point, and matching cleanup.

## Description

- Hash dirty Uint and Bool field values directly into `field_root_cache`.
- Preserve the dirty basic snapshot lifetime and all other field-root paths.

The implementation was primarily AI-authored and independently reviewed with AI assistance.


spiral-ladder (approved):
nice easy fix

PR #617, opened 2026-08-31
refactor(pmt): make node pool non-growable

GrapeBaBa:
## Motivation

`Node.Pool` runtime growth can relocate its `MultiArrayList` columns and invalidate borrowed node-root pointers. The production addon instead expects one bounded pool allocation during initialization.

## Description

- Allocate the configured user slots exactly once and return `PoolExhausted` instead of resizing.
- Keep the existing `pool_size` option, initialize the addon from `LODESTAR_Z_NODE_POOL_CAPACITY` or a 10,000,000-slot default, and remove the runtime growth API.
- Keep ERA Reader ownership unchanged and raise its fixed Pool capacity for state loading.
- Update affected callers and ownership regressions for deterministic fixed-capacity exhaustion.
- Make the all-field snapshot expansion proposed in #613 unnecessary because live Pool root-column addresses no longer move.

The implementation and regression tests were primarily AI-authored and independently reviewed with AI assistance.


bindings/napi/pool.zig
```diff
@@ -1,16 +1,22 @@
...
 const Node = @import("persistent_merkle_tree").Node;
 const RefCount = @import("state_transition").RefCount;
 
 /// Backs the `PoolRc` wrapper allocation only — `Node.Pool` uses its own
 /// `InitOptions` allocators.
 const allocator = std.heap.page_allocator;
 
-const default_pool_size: u32 = 0;
+const pool_size_environment_variable = "LODESTAR_Z_NODE_POOL_CAPACITY";
+const default_pool_size: u32 = 10_000_000;
```
GrapeBaBa, line 11: Is this a reasonable default size?

wemeetagain: I think this is fine for now. we will have to tune this.

src/persistent_merkle_tree/Node.zig
```diff
@@ -269,13 +273,18 @@ pub const Pool = struct {
...
         allocator: Allocator = std.heap.smp_allocator,
-        // Default 0 = lazy growth via preheat/ensureCapacity. Pre-PR NAPI
-        // bindings relied on this to avoid hundreds of MB upfront. Bench
-        // sites that want a pre-sized pool pass an explicit value.
-        pool_size: u32 = 0,
+        /// Usable user slots. The zero-hash sentinels are allocated in addition.
+        pool_size: u32,
     };
 
     pub fn init(opts: InitOptions) Error!Pool {
```
wemeetagain, line 280: maybe tighten the error union? InvalidPoolCapacity, OutOfMemory ?

src/persistent_merkle_tree/Node.zig
```diff
@@ -39,8 +39,12 @@ pub const Error = error{
...
     /// Attempt to use a length beyond the tree's length at a given depth.
     InvalidLength,
+    /// The configured pool size cannot be represented by the free list.
+    InvalidPoolCapacity,
     /// Attempt to increment the reference count beyond `max_ref_count`.
     RefCountOverflow,
+    /// The fixed-capacity pool has no free user slots.
+    PoolExhausted,
     /// Out of memory.
     OutOfMemory,
```
wemeetagain, line 49: may no longer be necessary?

bindings/napi/pool.zig
spiral-ladder: might want to add a comment saying that this is arbitrarily sized for now so the next reader doesn't wonder where we got 10,000,000 from

src/persistent_merkle_tree/node_test.zig
```diff
@@ -7,21 +7,20 @@ const Node = @import("Node.zig");
 const Gindex = @import("gindex.zig").Gindex;
 const ChunkedLeaf = @import("ChunkedLeaf.zig");
 
-// Allocate until the pool is full, so the next request has to grow (and fail). Returns the filler.
+// Allocate until the fixed pool is full. Returns the filler nodes to the caller for cleanup.
 fn drainPoolToFull(pool: *Node.Pool, out: *std.ArrayList(Node.Id)) !void {
```
spiral-ladder: nit: out of scope of changes for this PR but draining to full sounds counter intuitive because draining implies emptying 😁 maybe rename to `allocatePoolTillFull` or something?

src/persistent_merkle_tree/Node.zig
GrapeBaBa: still have dynamic allocation used by chunked leaf Node, need to evaluate if it is possible to static

src/persistent_merkle_tree/Node.zig
GrapeBaBa: I felt a bit long such kind of syntax `Allocator.Error||error{InvalidPoolCapacity}!Pool`, so changed to `Error!Pool`.

spiral-ladder (approved):

PR #618, opened 2026-09-01
refactor(ssz): simplify progressive bit list type

guha-rahul:
## Summary
- expose `ProgressiveBitList` directly as a value type instead of a zero-argument type factory
- remove the redundant negative API-presence test requested in ChainSafe/lodestar-z#99
- add focused coverage for direct value-type use

## Verification
- `zig build test:ssz`
- `zig fmt --check src/ssz/root.zig src/ssz/type/progressive_bit_list.zig`
- `pnpm lint` (passes with 10 existing warnings)

Addresses https://github.com/ChainSafe/lodestar-z/pull/99#discussion_r3897161271 and https://github.com/ChainSafe/lodestar-z/pull/99#discussion_r3897162738
This pr from lodekeeper-z was left behind during merging https://github.com/ChainSafe/lodestar-z/pull/99 into main.

### DIFF
--- a/src/ssz/type/progressive_bit_list.zig
+++ b/src/ssz/type/progressive_bit_list.zig
@@ -11,102 +11,100 @@ const mixInLength = @import("hashing").mixInLength;
 const Node = @import("persistent_merkle_tree").Node;
 const progressive = @import("progressive.zig");
 
-pub fn ProgressiveBitList() type {
-    return struct {
-        data: std.ArrayListUnmanaged(u8),
-        bit_len: usize,
+pub const ProgressiveBitList = struct {
+    data: std.ArrayListUnmanaged(u8),
+    bit_len: usize,
 
-        pub const empty: @This() = .{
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

src/ssz/type/progressive_bit_list.zig
```diff
@@ -408,6 +406,11 @@ pub fn ProgressiveBitListType() type {
     };
 }
 
+test "ProgressiveBitList is a directly usable value type" {
+    const bits: ProgressiveBitList = .empty;
```
spiral-ladder, line 410: is this meant to be for public use though? Seems like `ProgressiveBitList` is the internal impl while `ProgressiveBitListType` is what is exposed to consumers. if it's private then i think the test is unnecessary

spiral-ladder (approved):
lgtm w 1 question

PR #619, opened 2026-09-01
fix: reject withdrawal count mismatches

GrapeBaBa:
## Motivation

I thought it could be from the malicious block, `assert` may be not suitable

### DIFF
--- a/src/state_transition/block/process_block.zig
+++ b/src/state_transition/block/process_block.zig
@@ -80,6 +80,10 @@ pub fn processBlock(
                     .full => blk: {
                         const actual_withdrawals = block.body().executionPayload().inner.withdrawals;
                         if (withdrawals_result.withdrawals.items.len != actual_withdrawals.items.len) {
+                            std.log.err("withdrawal count mismatch: expected {d}, actual {d}", .{
+                                withdrawals_result.withdrawals.items.len,
+                                actual_withdrawals.items.len,
+                            });
                             return error.WithdrawalsLengthMismatch;
                         }
                         var root: Root = undefined;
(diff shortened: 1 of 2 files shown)


src/state_transition/block/process_block.zig
```diff
@@ -79,7 +79,9 @@ pub fn processBlock(
                 const payload_withdrawals_root = switch (block_type) {
                     .full => blk: {
                         const actual_withdrawals = block.body().executionPayload().inner.withdrawals;
-                        std.debug.assert(withdrawals_result.withdrawals.items.len == actual_withdrawals.items.len);
+                        if (withdrawals_result.withdrawals.items.len != actual_withdrawals.items.len) {
+                            return error.WithdrawalsLengthMismatch;
+                        }
```
nflaig, line 88: I suppose that is equivalent to the check we have [here](https://github.com/ChainSafe/lodestar/blob/1e9a530f9897d50d2d6337cd4143c8dc53667c66/packages/state-transition/src/block/processWithdrawals.ts#L76-L78) in lodestar?

```ts
if (expectedWithdrawals.length !== payload.withdrawals.length) {
  throw Error(`Invalid withdrawals length expected=${numWithdrawals} actual=${payload.withdrawals.length}`);
}
```

is it possible to add more details to the error itself, ie. in this case, what the length mismatch is

GrapeBaBa: Can't add context in pure error type in Zig unless changing to use struct, add a log for better debugging.

wemeetagain (approved):

PR #625, opened 2026-09-02
refactor: extract tests to _test.zig, add tidy lint

wemeetagain:
**Motivation**

Tests were crowding out the implementation in several files (`tree_view/list_basic.zig` was 75% tests), and there was no rule for where tests go or how they get wired in. Several files had tests that were never being run.

**Description**

Moves tests into sibling `_test.zig` files and adds a `tidy` lint to enforce the layout.

- A module has at most one `test` block. Past that, tests move to `<module>_test.zig`, wired from the module with `test { _ = @import("<module>_test.zig"); }`
- 56 `_test.zig` files added. Test bodies are unchanged, preludes rebuilt as imports/aliases
- 9 files keep tests inline because they test private declarations (`proto_array.zig`, `fork_choice.zig`, `cpu_count.zig`, ...). Listed in `inline_test_allowlist` in `test/tidy.zig`
- Add `test/tidy.zig`, run as `zig build test:tidy`, added to CI next to `zig fmt --check`
  - every `_test.zig` is imported from a `test` block, pairs with a module, and is wired from that module
  - a `root.zig` whose siblings have tests has a `test` block
  - every `src/` and `bindings/` file is reachable from a build root
  - every module in `build.zig.zon` has a CI step
  - no module has more than one `test` block
  - reads the AST, takes the file list from `git ls-files`, each rule has its own test
- Fixes found along the way
  - `era/root.zig` had no `test` block, its 19 tests never ran
  - `ssz/tree_view/root.zig` had no `test` block, `bit_list.zig`'s 13 tests never ran
  - `bounded_array` was never added to CI
  - Remove `ssz/type/serialized.zig` and `bindings/napi/napi_property_descriptor.zig`, unused
- Update AGENTS.md and CONTRIBUTING.md

TODO (separate PRs):
- 100 column limit (2338 violations, needs a reformat first), banned substrings, file/function length ratchet

**AI Assistance Disclosure**

Primarily AI-authored with Claude Code, reviewed and directed by me.

https://claude.ai/code/session_012RTf1wetpGaMYtMzCVxBFV


test/tidy.zig
```diff
@@ -0,0 +1,872 @@
...
+        const stem = file.basename[0 .. file.basename.len - "_test.zig".len];
+        const snake = try pairedModule(gpa, file.dir, stem, .snake);
+        const title = try pairedModule(gpa, file.dir, stem, .title);
+
+        const pair = findFile(files, snake) orelse findFile(files, title) orelse {
+            errors.addUnpairedTestFile(file.path, snake);
+            continue;
+        };
+        if (!pair.importsBasename(file.basename)) {
+            errors.addTestFileWiredElsewhere(file.path, pair.path);
```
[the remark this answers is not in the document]
wemeetagain: Confirmed and fixed in 956e95bd.

Verified against real Zig first, since the whole rule depends on this distinction:

```zig
// A: test block                          -> "All 2 tests passed."
test { _ = @import("foo_test.zig"); }

// B: unreferenced top-level const        -> "All 0 tests passed."
const moved_tests = @import("foo_test.zig");
```

So the accepted form ran nothing while tidy reported success, which is the same failure the rule exists to catch.

`Import` now records `in_test_block`, computed by testing whether the `@import` token falls inside a root-level `test_decl` span. The wiring rules use `wiresTests()` rather than `importsBasename()`. Re-checked against the tree: all existing wiring already used the test-block form, so no violations surfaced.

test/tidy.zig
```diff
@@ -0,0 +1,872 @@
...
+
+        var imported = false;
+        for (files) |other| {
+            if (std.mem.eql(u8, other.path, file.path)) continue;
+            if (other.importsBasename(file.basename)) {
+                imported = true;
+                break;
+            }
+        }
+        if (!imported) errors.addDeadFile(file.path);
```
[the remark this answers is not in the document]
wemeetagain: Confirmed and fixed in 956e95bd.

Reproduced against the real tree first: two mutually importing files under `src/` passed cleanly before the fix, and are reported after.

`tidyDeadFiles` now BFS-traverses from the roots declared in `build.zig.zon` instead of asking whether each file has an incoming import. Added `resolveImport`, which resolves a literal against the importing file's directory and returns null for module-name imports like `@import("hashing")`, since the build resolves those to another module's root rather than a path. Every module root is itself a declared root, so cross-module edges do not need following.

Diagnostic reworded to "not reachable from any build root", which is what the rule now actually checks.

test/tidy.zig
```diff
@@ -0,0 +1,872 @@
...
+    }
+}
+
+/// Every module declared in `build.zig.zon` must have a CI step, otherwise its
+/// tests never run on a pull request.
+fn tidyCiCoverage(gpa: Allocator, zon: []const u8, ci: []const u8, errors: *Errors) !void {
+    const modules = try declaredTestModules(gpa, zon);
+    for (modules) |module| {
+        const needle = try std.fmt.allocPrint(gpa, "zig build test:{s}", .{module});
+        if (std.mem.indexOf(u8, ci, needle) == null) errors.addModuleNotInCi(module);
```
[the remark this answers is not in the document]
wemeetagain: Confirmed and fixed in 956e95bd.

Verified both cases you named:

```
test:ssz removed from CI -> substring still matched "zig build test:ssz_generic_spec_tests"
test:bls removed from CI -> substring still matched "zig build test:bls_spec_tests"
```

Replaced the raw `indexOf` with `containsStep`, which requires the target name to end at the match rather than continue into a longer identifier. Deleting the real `test:ssz` step is now caught.

CONTRIBUTING.md
```diff
@@ -92,6 +92,64 @@ zig build run:write_ssz_static_spec_tests
...
+- Put the `test { _ = @import(...); }` block in the module under test rather than in the package
+  `root.zig`, so moving or deleting a module carries its tests along with it.
+- Do not mark a declaration `pub` just so a test can reach it from the sibling file. Tests that
+  deliberately exercise private internals belong inline; `src/cpu_count.zig` and
+  `src/state_transition/load_state.zig` are examples.
+- Zig only runs tests in files reached through an analyzed `test` block. A `_test.zig` that nothing
+  imports compiles fine, runs nothing, and still reports success, so check that
+  `zig build test:<module> --summary all` reports more tests after you add one.
+
+These rules are enforced, not just documented. Run:
```
wemeetagain: Don't include the description of tidy behavior as a sub-paragraph of "Test layout". Especially if tidy is more of a custom linter / formatter. It really should be (only briefly) described as a separate item.

AGENTS.md
```diff
@@ -301,6 +308,66 @@ Zig code must make allocator and ownership boundaries explicit:
...
+}
+```
+
+  Keeping the import local means the pairing survives moving or deleting the module, and no
+  package root has to be updated.
+- Move only the tests. Leave the test bodies unchanged and rebuild the prelude they relied on in
+  the new file as imports and aliases.
+- Private test helpers move with the tests they serve.
+- Do not widen a declaration to `pub` only to relocate a test. Tests that exercise private
+  internals stay inline. `cpu_count.zig`, `load_state.zig`, `proto_array.zig` and `fork_choice.zig`
```
wemeetagain: doesn't need examples here

AGENTS.md
```diff
@@ -301,6 +308,66 @@ Zig code must make allocator and ownership boundaries explicit:
...
+  are deliberate examples.
+- `memory_safety_test.zig` files cover a whole module rather than one file, so they stay wired
```
wemeetagain: state this rule generally, rather than just using this example.

AGENTS.md
```diff
@@ -301,6 +308,66 @@ Zig code must make allocator and ownership boundaries explicit:
...
+  from the package `root.zig`.
+
+Test discovery is easy to break silently. Zig only collects tests from files reached through an
```
wemeetagain: Everything below this point is not needed. Running tidy will expose all of this.

CONTRIBUTING.md
wemeetagain: Moved to its own item next to `zig fmt` in ff41ed5e, and cut to four lines:

> `tidy` is a custom linter for repo-wide rules that a formatter cannot express, currently covering test layout and test discovery. CI runs it, and so should you:
> ```sh
> zig build test:tidy
> ```
> It takes its file list from `git ls-files`, so run it after `git add` when you add new files.

"Test layout" is back to just the layout rules.

AGENTS.md
wemeetagain: Dropped in ff41ed5e. The rule now ends at "Tests that exercise private internals stay inline."

AGENTS.md
wemeetagain: Generalized in ff41ed5e:

> A test file covering a whole module rather than one sibling module stays wired from the package `root.zig`.

AGENTS.md
wemeetagain: Removed in ff41ed5e, everything from "Test discovery is easy to break silently" to the end of the section.

Tidy is still referenced from the critical rules, the build commands, and the pre-push checklist, so it stays discoverable without restating what it checks. AGENTS.md is down 33 lines.

.github/workflows/CI.yml
```diff
@@ -45,6 +45,13 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: zig fmt --check .
 
+      # Repo-wide lint rules that `zig fmt` cannot express. Platform independent,
+      # so it runs once rather than across the matrix.
```
spiral-ladder:
```suggestion
      # Custom repo-wide lint rules that `zig fmt` cannot express.
```

AGENTS.md
```diff
@@ -21,6 +21,10 @@
   trust boundary, native dependency, persistence path or format, shared mutable cache or pool,
   externally influenced native input, or supported integration. Update the normative threat model
   only when the security contract changes; otherwise update the implementation map.
+- **Test file layout:** once a file's `test` blocks reach roughly 100 lines or 8 tests, move them
```
spiral-ladder: im wondering if this is a good idea or if we should just move all tests regardless of number of lines/number of tests to a different file. I think consistency might be more important here

test/tidy.zig
```diff
@@ -0,0 +1,1056 @@
...
+//! through an analyzed `test` block, so a test file that nothing imports
+//! compiles cleanly, runs nothing, and still reports success. That failure is
+//! invisible in CI, which makes it worth a lint rather than a code review.
+//!
+//! Rules read the parsed AST rather than raw lines, so a brace or the word
+//! `test` inside a comment or string literal cannot skew them. Every rule
+//! reports all of its violations before failing, so one run shows the full list
+//! instead of only the first item. Each rule also has a test below that feeds it
+//! synthetic source and asserts the exact diagnostics, so a rule that stops
+//! catching anything fails here rather than going quiet.
```
spiral-ladder: had codex rewrite this a bit, its much shorter and clearer

```suggestion
   //! Lints that ensure Zig test files are discoverable.
   //!
   //! Zig only discovers tests in files imported by an analyzed `test` block.
   //! An unreferenced test file can compile and report success without running.
   //!
   //! These rules inspect the parsed AST, report all violations in one run, and
   //! include tests that verify the exact diagnostics.
```

nflaig (dismissed):

.github/workflows/CI.yml
spiral-ladder, line 49: i kinda think having it not in matrix is self explanatory


```suggestion
      # Repo-wide lint rules that `zig fmt` cannot express
```

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+/// A module holds at most this many `test` blocks. One inline test is a usage
+/// example; a second makes it a suite, and suites live in a sibling
+/// `<module>_test.zig`. The wiring block counts, so a module is either inline
+/// (one test) or extracted (only the wiring block), never both. Documented in
+/// AGENTS.md.
+const max_test_blocks = 1;
+
+/// Files that deliberately keep their tests inline. Their tests exercise private
+/// declarations that a sibling `_test.zig` cannot reach, and widening those
+/// declarations to `pub` purely to relocate a test is not a trade worth making.
```
spiral-ladder, line 38: this would be my only concern with this change, m worried that despite the "Never mark a declaration `pub` only to relocate a test." line AI might do it still, but we can see how it works out long term

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+/// runaway walk or a pathological file from turning a lint into an OOM.
+const max_tracked_files = 4096;
+const max_file_bytes: std.Io.Limit = .limited(8 << 20);
+const max_git_output_bytes = 4 << 20;
+
+/// A module holds at most this many `test` blocks. One inline test is a usage
+/// example; a second makes it a suite, and suites live in a sibling
+/// `<module>_test.zig`. The wiring block counts, so a module is either inline
+/// (one test) or extracted (only the wiring block), never both. Documented in
+/// AGENTS.md.
```
spiral-ladder, line 33:
```suggestion
/// Maximum number of `test` blocks allowed in a module.
///
/// A module is either inline (one test) or extracted (wiring block).
```

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+//! catching anything fails here rather than going quiet.
+
+const std = @import("std");
+const testing = std.testing;
+const Allocator = std.mem.Allocator;
+const Ast = std.zig.Ast;
+const assert = std.debug.assert;
+
+/// Upper bounds. The lint walks a source tree of known size, and these keep a
+/// runaway walk or a pathological file from turning a lint into an OOM.
```
spiral-ladder, line 24: these seem like its for the below 3 consts so it shouldnt just docuemnt `max_tracked_files`
```suggestion
// Upper bounds. The lint walks a source tree of known size, and these keep a
// runaway walk or a pathological file from turning a lint into an OOM.
```

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+const InlineTests = struct {
+    lines: usize = 0,
+    count: usize = 0,
+    /// One-based line of the first `test` declaration, for the diagnostic.
+    first_line: usize = 0,
+};
+
+/// One `@import` in a file.
+const Import = struct {
+    /// The literal, exactly as written.
```
spiral-ladder, line 185:
```suggestion
```

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+    path: []const u8,
+    /// Whether the import sits inside a root-level `test` declaration.
+    ///
+    /// This is the difference between wiring that works and wiring that only
+    /// looks like it does. Zig analyses a `test` block, so an import inside one
+    /// pulls in that file's tests. A plain top-level `const x = @import(...)`
+    /// that nothing references is never analysed, and its tests never run.
```
spiral-ladder, line 192: pretty confused by the last 4 lines, not sure if its needed anyway

```suggestion
    /// Whether the import sits inside a root-level `test` declaration.
```

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+        .inline_tests = inline_tests,
+    };
+}
+
+/// Lists the repository's tracked `.zig` files.
+///
+/// Uses `git ls-files` rather than a directory walk so the lint sees exactly
+/// what is committed: no build output, no vendored packages, no generated spec
+/// tests, and no untracked scratch files.
+fn listTrackedZigFiles(gpa: Allocator, io: std.Io) ![]const []const u8 {
```
spiral-ladder, line 296: hmm, kinda dislike that this has a hidden git dep

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+            else => return err,
+        };
+        try files.append(gpa, try analyze(gpa, path, text, errors));
+    }
+    return files.items;
+}
+
+// -------------------------------------------------------------------------
+// Helpers
+// -------------------------------------------------------------------------
```
spiral-ladder, line 353:
```suggestion
```

test/tidy.zig
```diff
@@ -0,0 +1,1044 @@
...
+            }
+        },
+    }
+    try name.appendSlice(gpa, ".zig");
+    return name.items;
+}
+
+// -------------------------------------------------------------------------
+// Rules
+// -------------------------------------------------------------------------
```
spiral-ladder, line 418:
```suggestion
```

spiral-ladder (approved):
lgtm and approving, just comments on removing ai slop comments

another possible (opinionated) way to organize is to have a `tidy.zig` and a `rules.zig`, feels like rules are big enough to deserve its own file, but not a must fix

PR #628, opened 2026-09-03
fix(bindings): support blinded state transitions

spiral-ladder:
We were always deserializing to full in the binding.


GrapeBaBa (approved):
LGTM

PR #635, opened 2026-09-03
feat(napi): expose forkSeq on BeaconStateView binding

lodekeeper:
Adds a `forkSeq` getter to the native `BeaconStateView` binding, requested for ChainSafe/lodestar#9927 (cc @nflaig).

**Why:** On the Lodestar TS side, `NativeBeaconStateView.forkSeq` now reads `binding.forkSeq` directly (rather than deriving it from `forkName` via `ForkSeq[forkName]`), and `IBeaconStateViewNative` no longer omits `forkSeq` — so the compiled `.node` binding needs to expose it.

**Change:**
- `bindings/napi/BeaconStateView.zig`: register the `forkSeq` property + a getter returning `@intFromEnum(state.forkSeq())` — the same `state.forkSeq()` the existing `forkName` getter already uses (`.forkSeq().name()`), just returned as the ordinal instead of the name.
- `bindings/src/index.d.ts`: add `forkSeq: ForkSeq` to the `BeaconStateView` interface + a local `ForkSeq` enum mirroring the existing `ForkName` enum.

The `ForkSeq` enum ordinals (`phase0=0 … gloas=7`) match both `src/config/fork_seq.zig`'s `ForkSeq` and `@lodestar/params`' `ForkSeq`, so `binding.forkSeq` equals the old `ForkSeq[forkName]` value.

Heads up: I don't have a Zig toolchain in this environment so I couldn't build locally — leaving CI to verify the addon build/tests. The change mirrors the existing `forkName` prop/getter exactly.

🤖 Generated with AI assistance

### DIFF
--- a/bindings/napi/BeaconStateView.zig
+++ b/bindings/napi/BeaconStateView.zig
@@ -25,6 +25,7 @@ pub const js_meta = js.class(.{ .properties = .{
     .slot = js.prop(.{ .get = true, .set = false }),
     .fork = js.prop(.{ .get = true, .set = false }),
     .forkName = js.prop(.{ .get = true, .set = false }),
+    .forkSeq = js.prop(.{ .get = true, .set = false }),
     .epoch = js.prop(.{ .get = true, .set = false }),
     .genesisTime = js.prop(.{ .get = true, .set = false }),
     .genesisValidatorsRoot = js.prop(.{ .get = true, .set = false }),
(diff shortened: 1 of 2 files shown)


nflaig (approved):
seems good to me, noticed the ci runs for external contributors automatically, this is a security risk cc @wemeetagain @matthewkeil

wemeetagain (approved):

PR #641, opened 2026-09-04
perf(ssz): size progressive fixed lists without materializing values

wemeetagain:
Calculating a fixed-element progressive list tree's serialized size materialized the entire list, allocating node-ID and value buffers.

Read the length leaf and use checked multiplication by the fixed element size.


spiral-ladder (approved):
nice, much simpler

PR #642, opened 2026-09-04
perf(ssz): avoid scratch allocation in chunked list serialization

wemeetagain:
Chunked fixed-list serialization allocated a temporary array of node IDs even when the caller supplied the output buffer. Stream those IDs through the existing depth iterator instead, preserving full chunks, zero subtrees, and the partial tail.


src/ssz/type/list_test.zig
```diff
@@ -1113,3 +1113,110 @@ test "FixedListType opts.chunked_leaf=true: serialize -> deserialize round-trip"
     defer pool.unref(round_id);
     try std.testing.expectEqualSlices(u8, tree_id.getRoot(&pool), round_id.getRoot(&pool));
 }
+
+test "FixedListType chunked serialization needs no scratch allocation" {
```
GrapeBaBa, line 1117: Currently I put these kinds of memory specific tests in separate test file, do we need it?

wemeetagain: good question. imo after #625, we should reconsider the separate memory test file?

cc @spiral-ladder @nazarhussain

spiral-ladder: i prefer putting all tests in `<file_name>_test.zig` rather than consolidating memory safety tests in a `memory_safety_test.zig`

GrapeBaBa: Also have questions about what kind test should be put in `<file_name>_test.zig`, is that such as behavior tests? Mixing all the tests seems will make the `<file_name>_test.zig` big quickly, though I noticed roc-compiler and ghostty indeed just did it.

wemeetagain: One practical thing I noticed is that having a separate memory_safety_test.zig more easily introduced merge conflicts between PRs. :joy: Not the end of the world, but it is a little annoying with our stricter branch protection that independent PRs can't easily be merged without two parties (author and reviewer) syncing the `review PRs->merge one->fix next PR->merge next`.

Yes I think the test file will get massive, but I suppose the benefit is that the behaviors being tested are "closer" to the implementation. And its a simpler rule to follow: "all unit / behavior tests go in a sibling test file".

GrapeBaBa (approved):

PR #644, opened 2026-09-04
refactor(pmt): compute proof descriptors with integer paths

wemeetagain:
Proof descriptors previously converted every generalized index and sibling path into allocated binary strings, then assembled another string before packing bytes. This also leaked a duplicated key when a following map allocation failed.

Keep paths as integers, sort them in the same lexical path order, and write descriptor bits directly into the final buffer. Bound map growth and checked size arithmetic.


src/persistent_merkle_tree/proof.zig
```diff
@@ -261,171 +261,66 @@ pub const Proof = union(ProofType) {
...
-                for (0..i) |_| {
-                    try descriptor_bitstring.append(allocator, '0');
-                }
-                try descriptor_bitstring.append(allocator, '1');
-                break;
-            }
+    for (gindices) |gindex| {
+        var current = @intFromEnum(gindex);
+        try proof_indices.put(current, {});
+        for (0..gindex.pathLen()) |_| {
```
GrapeBaBa, line 284: can we add a bit comment for such bit operation trick

src/persistent_merkle_tree/proof_test.zig
```diff
@@ -190,6 +190,54 @@ test "computeDescriptor - should convert gindices to a descriptor" {
...
+    });
+    defer testing.allocator.free(descriptor);
+
+    var expected: [(depth * 2 + 1 + 7) / 8]u8 = @splat(0);
+    for (depth..depth * 2 + 1) |bit| expected[bit / 8] |= @as(u8, 0x80) >> @intCast(bit % 8);
+    try testing.expectEqualSlices(u8, &expected, descriptor);
+}
+
+test "computeDescriptor cleans up every allocation failure" {
+    try testing.checkAllAllocationFailures(testing.allocator, computeDescriptorWithAllocator, .{});
```
GrapeBaBa, line 229: same separate memory tests question

GrapeBaBa (approved):
Looks good

PR #648, opened 2026-09-05
fix(bindings): make native STF setup and memory ownership safe

wemeetagain:
Lodestar's experimental native STF integration needs configuration and state ownership that remain valid across historical workers, reloads, and garbage collection. This adds immutable `StateTransition` setup handles, isolates mutable worker resources, and reports native state memory to V8. Each environment allocates its tree pool on first state construction so default TypeScript/BLS workers remain lightweight.

The binding accepts full and blinded signed blocks, validates numeric and serialized inputs, preserves SSZ bigint fields, and rejects unsupported Gloas transitions. Legacy `config.set` affects subsequent construction only. Sync committee lookups return the complete declared cache, including duplicate validator positions, and canonical `SLOT_DURATION_MS` takes precedence over deprecated seconds defaults.

Bing's refreshed `bing/stf` branch and current `main` are merged. Explicit `release()` is idempotent and preserves independently retained descendants. Native operations retain their resources while JavaScript input getters or output callbacks run, so a callback can request release without freeing an active call's state. Subsequent access fails immediately; cleanup completes when the active calls return. Regressions cover the callback crashes found during integration.

This supports controlled pre-Gloas fleet testing through [Lodestar #10014](https://github.com/ChainSafe/lodestar/pull/10014), targeting [#9632](https://github.com/ChainSafe/lodestar/pull/9632). Gloas transition parity remains outside this change. Ownership and host integration requirements are documented beside the owning APIs. Validator monitors and their metrics follow the same per-environment lifetime as the other mutable STF resources.

The earlier devnet validation used native `d7e17aa0` and Lodestar `e836ad8780`, before this refresh. That three-node Fulu Kurtosis devnet passed Assertoor stability and proposal checks, matched native/TypeScript sync duties and historical state, published native sync contributions, and agreed at slot 777 / finalized epoch 22. The enclave has been removed. Repeated observer replay hit the two-peer topology's request limits; checkpoint recovery restored agreement.

> This PR was written primarily by Codex.


bindings/napi/BeaconStateView.zig
```diff
@@ -75,160 +73,215 @@ pub const js_meta = js.class(.{ .properties = .{
...
         .{
-            .config = &config.state.config,
+            .config = beacon_config,
             .pubkey_cache = &pubkey.state.cache,
         },
         null,
     );
 }
 
+fn unsignedInteger(value: js.Number) !u64 {
```
GrapeBaBa: Maybe this should be implemented in zapi like `exactU32`

bindings/napi/BeaconStateView.zig
```diff
@@ -75,160 +73,215 @@ pub const js_meta = js.class(.{ .properties = .{
...
+    return fork_seq;
+}
+
 // -------------------------
 // Class Methods
 // -------------------------
 /// Requires state bytes with trusted provenance; SSZ decoding does not authenticate them.
-pub fn createFromBytes(bytes: js.Uint8Array) !BeaconStateView {
+pub fn createFromBytes(bytes: js.Uint8Array, setup: ?*const StateTransition) !BeaconStateView {
+    const config_rc = if (setup) |value| value.config_rc else config.state.current orelse return error.ConfigNotInitialized;
```
GrapeBaBa: After introduce setup, we don't need keep `config.state.current`?

bindings/napi/BeaconStateView.zig
```diff
@@ -557,78 +643,59 @@ pub fn nextProposers(self: *const BeaconStateView) !?js.Array {
...
 }
 
-pub fn currentSyncCommittee(self: *const BeaconStateView) !js_types.SyncCommittee {
+pub fn currentSyncCommittee(self: *BeaconStateView) !js_types.SyncCommittee {
     const env = js.env();
-    const cached_state = try self.requireState();
+    const cached_state = try self.acquireState();
+    defer self.finishState();
+    const allocator = cached_state.allocator;
+    defer self.pool_rc.?.instance.reportMemory();
```
GrapeBaBa: Seems some get functions can also trigger memory allocation such create treeview internal but don't call `reportMemory`

bindings/napi/config_snapshot.zig
```diff
@@ -0,0 +1,167 @@
...
+const js = @import("zapi:zapi").js;
+const active_preset = @import("preset").active_preset;
+const c = @import("config");
+const BeaconConfig = @import("config").BeaconConfig;
+const ChainConfig = @import("config").ChainConfig;
+const Preset = @import("preset").Preset;
+
+const max_blob_schedule_entries = 16;
+
+const allocator = std.heap.c_allocator;
```
GrapeBaBa: Why here need a global allocator?

bindings/napi/config_snapshot.zig
```diff
@@ -0,0 +1,167 @@
...
+    }
+    if (chain_config.SLOT_DURATION_MS == 0 or chain_config.SLOT_DURATION_MS % 1000 != 0) {
+        return error.InvalidSlotDuration;
+    }
+    chain_config.SECONDS_PER_SLOT = @divExact(chain_config.SLOT_DURATION_MS, 1000);
+    snapshot.instance.config = BeaconConfig.init(chain_config, root_slice[0..32].*);
+    return snapshot;
+}
+
+fn valueToU64(value: napi.Value) !u64 {
```
spiral-ladder: Feels like this should be in zapi

bindings/napi/metrics.zig
```diff
@@ -2,27 +2,40 @@ const std = @import("std");
...
+/// JS: metrics.init({historical?: boolean}) → void
+pub fn init(options: ?js.Value) !void {
+    var use_historical_prefix = false;
+    if (options) |value| {
+        const raw = value.toValue();
+        if (try raw.hasNamedProperty("historical")) {
+            use_historical_prefix = try (try raw.getNamedProperty("historical")).getValueBool();
+        }
+    }
+    if (historical) |previous| {
```
spiral-ladder: nit: feels confusing to have both historical and use_historical_prefix

bindings/test/stateRelease.test.ts
```diff
@@ -0,0 +1,227 @@
...
+])("retains the active state through release in a %s", {timeout: 40_000}, (_name, source) => {
+  runProcess(`
+    const state = new bindings.StateTransition(stfConfig, new Uint8Array(32)).createFromBytes(bytes);
+    ${source}
+    assert.throws(() => state.slot, {code: "InvalidState"});
+    state.release();
+  `);
+});
+
+function runProcess(source: string, exposeGc = false): void {
```
spiral-ladder: this looks fine but i wonder if there's a better way to do this than to parse in source code as a string? It looks ok when it's a few lines of template strings but we have a few test files here that are almost entirely template strings, I guess it isn't a huge concern bc we can get AI to audit these lines but still it's not very easily human readable/understandable, worth a followup PR but not blocking this for now

bindings/napi/tracked_allocator.zig
```diff
@@ -0,0 +1,42 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const Alignment = std.mem.Alignment;
+const TrackedAllocator = @This();
+
+backing: Allocator,
+bytes_in_use: usize = 0,
+
+pub fn allocator(self: *TrackedAllocator) Allocator {
+    return .{ .ptr = self, .vtable = &.{ .alloc = alloc, .resize = resize, .remap = remap, .free = free } };
```
spiral-ladder:
```suggestion
    return .{ .ptr = self, .vtable = &.{ .alloc = alloc, .resize = resize, .remap = remap, .free = free, } };
```

fmt

bindings/napi/tracked_allocator.zig
spiral-ladder: i think maybe a doc comment on what this allocator is specialized for helps:

```
//! A standard Zig allocator wrapped with a byte counter to expose to V8.
```

bindings/napi/config_snapshot.zig
```diff
@@ -0,0 +1,167 @@
...
+const c = @import("config");
+const BeaconConfig = @import("config").BeaconConfig;
+const ChainConfig = @import("config").ChainConfig;
+const Preset = @import("preset").Preset;
+
+const max_blob_schedule_entries = 16;
+
+const allocator = std.heap.c_allocator;
+
+pub const ConfigSnapshot = struct {
```
spiral-ladder: snapshot sounds vague, can we name this OwnedConfig or something maybe?

spiral-ladder (commented):
Looks really good apart from some nits and minor stuff, making the cache and metrics `threadlocal` means a lot of the concerns around multi-threaded access between main thread and the historical regen thread are gone

bindings/napi/config_snapshot.zig
spiral-ladder: https://github.com/ChainSafe/zapi/pull/80, had my claude run a local pull, build and test on this branch, works fine

src/persistent_merkle_tree/Node.zig
```diff
@@ -347,12 +349,8 @@ pub const Pool = struct {
     }
 
     /// Returns the number of nodes currently in use (not free).
-    pub fn getNodesInUse(self: *Pool) usize {
-        var count: usize = 0;
-        for (self.nodes.items(.state)) |s| {
-            if (!s.isFree()) count += 1;
-        }
-        return count;
+    pub fn getNodesInUse(self: *const Pool) usize {
```
GrapeBaBa: I still want original loop approach can be used in test to avoid the mistake of count usage

PR #651, opened 2026-09-07
fix: prevent duplicate Altair sync cache cleanup

GrapeBaBa:
## Motivation

During the Phase0-to-Altair upgrade, an allocation failure after the next sync cache transfers to its reference-counted wrapper can release the same cache twice.

## Description

- Scope raw-cache cleanup to RC construction for both sync caches, so each resource has one cleanup owner. Preserve allocation order and error propagation. 
- Add allocation-failure regression coverage for both RC allocations and the first allocation after ownership transfer, checking unchanged existing caches and successful subsequent setup.


wemeetagain (approved):
LGTM. It does make me wonder if there's some refactoring potential around `SyncCommitteeCacheAllForks` and `SyncCommitteeCacheRc` based on this unergonomic initialization.

PR #652, opened 2026-09-07
fix: preserve sync committee ownership on rotation OOM

GrapeBaBa:
## Motivation

Sync-committee rotation can corrupt shared references on OOM, causing candidate cleanup to invalidate the pre-state cache. 

## Description

- Construct the replacement cache and RC before transferring current/next references, and release the raw cache if RC allocation fails.
- Add regression coverage for allocation failure, candidate teardown, and successful rotation.


spiral-ladder (approved):
Lgtm, honestly we can consider just allocating a 512 length array for the indices so we don't have to use an allocator here, reduces chance of error as well but maybe in a future PR? Granted the overall struct still needs an allocator

PR #655, opened 2026-09-07
fix: report proposer rewards from block transitions

GrapeBaBa:
## Motivation

`BeaconStateView.proposerRewards` reports zero even when block processing credits proposer rewards.

## Description

- Record attestation, sync aggregate, and slashing rewards using the existing computed amounts, and reset reporting on state clone to match Lodestar TS. 

- Add Zig regressions and a native binding test comparing rewards and state roots with Lodestar TS, including successive blocks.


spiral-ladder (approved):
need this for metrics, lgtm, merging first

PR #662, opened 2026-09-08
feat(metrics): record exported but unset state transition gauges

spiral-ladder:
These metrics were declared and scraped but not written to:

- `lodestar_stfn_state_cloned_count`
- `epoch_transition_commit_seconds`,
- `effective_balance_updates_count`
- `validators_in_{activation,exit}_queue`,
- `new_seen_attesters_per_block_total`,
- `new_seen_attesters_effective_balance_per_block_total` and
- `attestations_per_block_total`

Note that for epoch transition commit, we deviate slightly, since we do not commit automatically. We use a var to track if a transition took place and then time the commit only if an epoch was processed.


wemeetagain (dismissed):
looks good, merge conflict

src/state_transition/state_transition.zig
```diff
@@ -140,7 +143,13 @@ pub fn processSlots(
         }
     }
 
+    // Unlike lodestar-ts, the epoch transition above does not commit on its own
+    // so we manually commit and report here
+    const commit_timer = time.start(io);
     try state.commit();
+    if (epoch_transition_processed) {
+        metrics.state_transition.epoch_transition_commit.observe(time.durationSeconds(time.since(io, commit_timer)));
```
GrapeBaBa: This may not be epoch transition commit, just last slot commit. Why we don't use the lodestar's approach?

spiral-ladder: not sure i understand, this is the equivalent of this

https://github.com/ChainSafe/lodestar/blob/30d2b0bf521ce2b6e96f33f6b945b137cf579a2d/packages/state-transition/src/stateTransition.ts#L301-L305

 no?

GrapeBaBa: This commit more like https://github.com/ChainSafe/lodestar/blob/v1.47.0/packages/state-transition/src/stateTransition.ts#L183, it commit after all slots processed. If slot from 31 -> 33 or even cross multi epochs, this timer seems not record correct epoch transition like lodestar.

spiral-ladder: oh yeah good catch, this should be fixed in https://github.com/ChainSafe/lodestar-z/pull/662/changes/ed9df7a37abc5be93a9876150652214be64633a8

GrapeBaBa (approved):
Looks good

PR #670, opened 2026-09-09
test: colocate memory safety tests with their modules

wemeetagain:
Moves the 64 tests from all four `memory_safety_test.zig` files into 23 sibling `<module>_test.zig` suites, with imports wired from the modules they cover. Every relocated test name starts with `memory_safety: `, so the suite remains selectable with filters such as `zig build test:ssz -Dssz.filters=memory_safety`.

Preserves test behavior and existing private visibility, relocates the inline tests that now belong to suites, removes the obsolete tidy exemption, and documents the convention in AGENTS.md. Runtime behavior is unchanged.

AI assistance: this implementation was primarily AI-authored using Codex.


GrapeBaBa (approved):
Looks goood

Issue #158, opened 2026-01-05
SSZ API improvements


wemeetagain:
There are a few rough edges I'd like to resolve, this list is pretty opinionated:
- Currently, `VariableST.hashTreeRoot`, `VariableST.serialized.hashTreeRoot`, `VariableST.tree.fromValue`, `VariableST.tree.deserializeFromBytes` all require internal allocations. This can and should be avoided such that the end result is only `bytes -> value` and `tree -> value` require allocations (to resize the output value) and no other operations need to allocate.
  - for `hashTreeRoot`, implementation can be either fixed-size, internally stack-allocated chunks, or pass in managed hasher memory

PR #671, opened 2026-09-09
perf(ssz): stream list tree reads into output values

wemeetagain:
Reading an SSZ list tree into a value allocates a temporary node-ID array even when the destination already has sufficient capacity. Use the existing bounded depth iterator for plain, chunked, and variable-element lists so allocations are limited to the output value and its children. Packed decoding, zero-subtree handling, and partial-value cleanup retain their existing behavior.

Follow-up to #158. This PR is independent of the open proof and hashing work.


GrapeBaBa (approved):
LGTM

PR #647, opened 2026-09-04
refactor(pmt): bound lazy hashing and compact proof depth

wemeetagain:
Lazy branch-root hashing used recursion, and compact-proof reconstruction could create trees deeper than the supported limit. Use a fixed `max_depth` postorder stack and a pool-capacity visit budget for hashing, preserving cached roots, shared subtrees, chunked-leaf scratch, and opaque callbacks.

Validate compact-proof shape, canonical padding, and depth before allocating the decoded descriptor or tree nodes. Excessive depth returns `InvalidProofDepth` at the fallible construction boundary.

`getRoot()` keeps its existing infallible signature. Its panic guards invalid native graph construction; raw native callers must preserve acyclicity and the supported depth. The implementation map records these contracts.

Implementation primarily AI-authored, with independent AI review.

PR #672, opened 2026-09-09
perf(ssz): bound list hashing scratch with batched accumulation

wemeetagain:
SSZ list value and serialized hashing currently allocate a chunk array proportional to the input. Stream fixed and variable list contents through a Merkle accumulator with 64-chunk batches and one root per supported tree level, keeping list-level scratch around 4 KiB. The accumulator hashes partial batches without expanding missing leaves and uses cached roots for empty trees. Child hashers retain their own allocation behavior, and existing allocator-taking entry points remain available.

Follow-up to #158. This branches from main and composes independently with #644, #647, and #671. Managed `Hasher`, bit-list, and progressive hashing remain separate follow-ups.


GrapeBaBa (approved):
Looks great

PR #673, opened 2026-09-09
refactor(ssz): share byte-list hashing with generic uint8 lists

wemeetagain:
`ByteListType` duplicates the value and serialized hashing algorithms for `FixedListType(UintType(8))`. Alias those two entry points to the generic implementation, preserving the public signatures and keeping one implementation of byte packing, padding, and length mixing. This can merge independently; together with #672 it gives byte lists the bounded hashing kernel without a second implementation.


GrapeBaBa (approved):
LGTM

spiral-ladder (approved):

PR #674, opened 2026-09-10
fix: stabilize ptc benchmark seeds

lodekeeper-z:
**Motivation**

The PTC performance fixtures call `crypto.randomBytes` without importing Node's crypto module. On Node 22, `crypto` is the Web Crypto global, so benchmark collection fails:

```text
TypeError: crypto.randomBytes is not a function
```

`@chainsafe/benchmark` 2.0.1 still exits successfully and persists only 24 results, silently dropping the shuffle suite.

**Description**

Use `deterministicBenchmarkSeed` for the per-slot and full-epoch PTC fixtures, with labels separated by operation and validator count. Two complete benchmark executions now each produce all 49 results, including the 25 shuffle cases. The measured implementations are unchanged.

**AI Assistance Disclosure**

The implementation and verification were primarily AI-authored.


GrapeBaBa (approved):
Yes, this is one PR merged after deterministic seed change, good catch

PR #678, opened 2026-09-10
perf: bound bitlist hashing workspace

wemeetagain:
**Motivation**

BitList value and byte hashing allocate a complete chunk buffer for every hash. Ordinary lists already have bounded accumulation through #672.

Related to #158.

**Description**

Use `MerkleAccumulator` in both BitList hashing paths, with workspace independent of input length.

- `BitListType.hashTreeRoot` asserts the typed value's length and storage invariants, passes full chunks directly to the accumulator, and pads only a partial tail.
- `BitListType.serialized.hashTreeRoot` validates the bytes before hashing and passes complete 256-bit chunks directly. A final chunk containing the delimiter is copied and stripped before accumulation. Both paths mix in the bit length.
- Extend allocation-disabled comparisons against tree roots around byte, chunk, and 63/64/65-chunk boundaries.

**AI Assistance Disclosure**


src/ssz/type/bit_list.zig
```diff
@@ -275,14 +275,18 @@ pub fn BitListType(comptime _limit: comptime_int) type {
...
-            const chunks = try allocator.alloc([32]u8, (chunkCount(value) + 1) / 2 * 2);
-            defer allocator.free(chunks);
-
-            @memset(chunks, [_]u8{0} ** 32);
-            @memcpy(@as([]u8, @ptrCast(chunks))[0..value.data.items.len], value.data.items);
-
-            try merkleize(@ptrCast(chunks), chunk_depth, out);
+        pub fn hashTreeRoot(_: std.mem.Allocator, value: *const Type, out: *[32]u8) !void {
+            if (value.bit_len > limit) return error.tooLarge;
+            std.debug.assert(value.data.items.len == (value.bit_len + 7) / 8);
```
GrapeBaBa, line 280: why this check could use `assert` but above need `error`

src/ssz/type/bit_list.zig
```diff
@@ -275,14 +275,18 @@ pub fn BitListType(comptime _limit: comptime_int) type {
...
+            var accumulator = MerkleAccumulator.init(chunk_depth);
+            for (0..chunkCount(value)) |i| {
+                var chunk: [32]u8 = @splat(0);
+                const start = i * 32;
+                const len = @min(32, value.data.items.len - start);
+                @memcpy(chunk[0..len], value.data.items[start..][0..len]);
```
twoeths: we don't really need this copy from 0 to `chunk_count - 1`, as `MerkleAccumulator.append()` already handled it 
is it worth to do the for loop to `chunk_count - 1`, and treat the last chunk separately with an extra copy?

src/ssz/type/bit_list.zig
wemeetagain: Good point. Both seem to assert invalid state / improper usage of `value` (which is a `BitList(limit)`).

GrapeBaBa (dismissed):

GrapeBaBa (approved):

PR #679, opened 2026-09-10
perf: stream bitlist tree reads

wemeetagain:
**Motivation**

BitList tree-to-value conversion allocates a node-ID array even when the destination already has enough capacity. Ordinary list conversion was addressed in #671.

Related to #158.

**Description**

Read chunks with `Node.DepthIterator` directly into the resized BitList output. Boundary and implicit-zero tests disable allocations after reserving output capacity and verify that source roots remain unchanged.

**AI Assistance Disclosure**


GrapeBaBa (dismissed):

twoeths (commented):
there is a conflicting file @wemeetagain

GrapeBaBa (approved):

PR #681, opened 2026-09-10
refactor: bound fixed vector hashing

wemeetagain:
**Motivation**

Fixed-element vectors reserve hashing scratch proportional to their declared size on the stack. Large vectors therefore consume large stack frames even though Merkleization can stream chunks.

Related to #158.

**Description**

Keep direct hashing for at most 64 chunks and use `MerkleAccumulator` above that bound. Stream packed values and composite roots, and reject incorrect serialized lengths before reading bytes.

A local ReleaseSafe uint8-vector microbenchmark (20,000 calls per size) preserved the small-vector timings. Larger samples show the tradeoff: 2,080 bytes took 3.15 us with the previous full-buffer algorithm versus 3.80 us with accumulation; 32,768 bytes took 44.10 us versus 50.63 us. Workspace no longer grows with vector length. These samples are not a whole-client throughput prediction.

**AI Assistance Disclosure**


GrapeBaBa (dismissed):

GrapeBaBa (approved):

GrapeBaBa:
@wemeetagain conflict need fix to merge

PR #683, opened 2026-09-10
perf: stream compact proof generation

wemeetagain:
**Motivation**

Compact-proof generation expands descriptor bits and recursively allocates and concatenates arrays of leaves. Descriptor validation from #647 already establishes the shape and depth needed for a bounded traversal.

The descriptor is caller-controlled through `BeaconStateView.createMultiProof`. Reserving its entire declared output before navigating the source tree can amplify malformed input into a large allocation, as reported in [review](https://github.com/ChainSafe/lodestar-z/pull/683#discussion_r3979609729).

Related to #158.

**Description**

Traverse packed descriptor bits with a bounded pending-node stack and grow the output as traversal reaches witnesses. Invalid source paths release partial output; opaque materialization retains its existing temporary ownership.

Add regressions for impossible paths before and after a valid witness, output growth failures, and opaque-materialization failures. A descriptor declaring roughly 128 KiB of output now reports the invalid path with a 1 KiB output budget.

**AI Assistance Disclosure**


nazarhussain (approved):

nazarhussain:
Just a side note

The **1 MiB depth-22 descriptor requests roughly 128 MiB** scenario isn't reachable through the shipped API. `beacon-node/src/api/impl/proof/index.ts:15-22` already rejects `descriptor.length / 2 > maxGindicesInProof` (default 512), so HTTP-reachable descriptors are capped at ~1 KiB.

PR #684, opened 2026-09-10
refactor: iterate compact proof reconstruction

wemeetagain:
**Motivation**

Compact-proof reconstruction still allocates an expanded descriptor and recursively builds nodes after validating the proof depth.

Related to #158.

**Description**

Read packed descriptor bits with a bounded stack of unfinished left subtrees. Validate the leaf count before node allocation and release every unfinished subtree on failure. Regressions cover both maximum-depth spine shapes, allocation-free scratch, and every pool-exhaustion point on shorter spines.

**AI Assistance Disclosure**


src/persistent_merkle_tree/proof.zig
```diff
@@ -426,61 +426,48 @@ pub fn createCompactMultiProof(
...
+        }
+        if (depth == 0) {
+            std.debug.assert(bit_index + 1 == bit_length);
+            std.debug.assert(leaf_index == leaves.len);
+            try pool.ref(current);
+            return current;
+        }
+        left_roots[depth - 1] = current;
+    }
+    unreachable;
```
nazarhussain, line 470: How about returning an error instead, as the future use case will based on user input and in in `ReleaseSafe` the `unreachable` is a `panic -> process abort` .

```suggestion
    return error.InvalidWitnessLength;;
```

Error return will cost nothing extra and keep the process alive.

src/persistent_merkle_tree/proof.zig
```diff
@@ -426,61 +426,48 @@ pub fn createCompactMultiProof(
...
 pub fn createNodeFromCompactMultiProof(
     pool: *Node.Pool,
     leaves: [][32]u8,
     descriptor: []const u8,
 ) (Node.Error || Error)!Node.Id {
-    var arena = std.heap.ArenaAllocator.init(pool.allocator);
-    defer arena.deinit();
-    const temp_allocator = arena.allocator();
-
-    const bitlist = try descriptorToBitlist(temp_allocator, descriptor);
```
nazarhussain, line 469: The function `descriptorToBitlist` had only usage here, now it's a dead code. Should we remove it?

nazarhussain (changes requested):

PR #686, opened 2026-09-10
perf: stream progressive tree serialization

wemeetagain:
**Motivation**

Fixed-element progressive-list tree serialization first materializes the entire list value, including a node-ID traversal array.

Related to #158.

**Description**

Add `progressive.NodeIterator` and stream directly into the caller's output. Bound traversal depth, check size arithmetic and output capacity, expand implicit zero subtrees, and validate the right-spine terminator. Child serializers keep their own workspace contracts. Share `chunkCountForLength()` across value and tree paths.

**AI Assistance Disclosure**


src/ssz/type/progressive_list.zig
```diff
@@ -232,12 +232,26 @@ pub fn FixedProgressiveListType(comptime ST: type) type {
...
-
-                try toValue(allocator, node, pool, &value);
-                return Self.serializeIntoBytes(&value, out);
+                const len = try length(node, pool);
+                const size = try std.math.mul(usize, len, Element.fixed_size);
+                if (out.len < size) return error.InvalidSize;
+                const chunk_count = if (comptime isBasicType(Element))
+                    size / 32 + @intFromBool(size % 32 != 0)
+                else
+                    len;
```
spiral-ladder: seems like we have multiple ways to derive `chunk_count` at this point and it's all inline, we could merge them into an internal fn and use that instead

src/ssz/type/progressive_list_test.zig
```diff
@@ -256,3 +256,57 @@ test "memory_safety: variable progressive list tree.toValue preserves out on OOM
     try std.testing.expect(saw_failure);
     try std.testing.expect(saw_success);
 }
+
+test "fixed progressive tree serialization streams without temporary values" {
```
spiral-ladder: Maybe this is better? Can't use no-alloc bc the pool allocates, but at least we can state this

```suggestion
test "fixed progressive tree serialization streams without temporary allocations" {
```

GrapeBaBa (approved):

PR #690, opened 2026-09-10
refactor(ssz): centralize chunked-leaf mutation bookkeeping

wemeetagain:
**Motivation**

The packed SSZ setter currently coordinates copying, dirty tracking, payload mutation, and hash invalidation through direct access to pool internals. Move that bookkeeping into scoped operations at the layers that own it, as part of #158.

**Description**

Centralize chunked-leaf mutation bookkeeping while preserving commit and snapshot semantics.

- `TreeViewState.editChunkedLeaf` prepares private storage and registers the pending change.
- `Node.Id.editChunkedLeaf` invokes the packing callback, updates the valid length, and invalidates the cached root.

The SSZ caller passes its existing packing function directly. All fallible preparation completes before the callback runs. Shared leaves are copied on the first write; subsequent pending writes reuse storage without allocating.

Add coverage for snapshot isolation, repeated writes, hash invalidation, sparse leaves, length growth, allocation failures, and every supported basic element type.

**AI Assistance Disclosure**

Implementation primarily AI-authored.


GrapeBaBa (dismissed):
Looks great for simplify

GrapeBaBa (approved):

PR #693, opened 2026-09-14
refactor: extract bit array module

wemeetagain:
**Motivation**

Packed-bit values live inside SSZ schema modules, coupling ordinary bit operations to SSZ. Bounded and progressive bitlists also duplicate the same growable storage and mutation code.

This continues the library organization and coupling work in #158.

**Description**

Extract packed-bit values into a standalone `bit_array` module that depends only on `std`.

- `BitList(.{ .limit = n })` and `BitList(.{ .limit = bit_array.unlimited })` share one implementation, with the optional limit resolved at compile time.
- `BitVector(n)` retains inline storage and allocator-free mutation.
- `setAssumeCapacity()` is infallible and asserts its existing-length precondition; callers use `set()` when growth or recoverable limit errors are needed.
- `BitListType`, `ProgressiveBitListType`, and `BitVectorType` use the standalone values. Remove the old SSZ value exports and update callers directly, without compatibility wrappers. Serialization, hashing, and tree views remain in SSZ.

**AI Assistance Disclosure**

Implementation primarily AI-authored, with maintainer-directed API design.


src/bit_array/bit_list.zig
```diff
@@ -0,0 +1,236 @@
...
+
+pub const BitListOptions = struct {
+    /// Maximum logical length in bits; null imposes no limit beyond usize.
+    limit: ?usize,
+};
+
+/// Owns packed bytes with least-significant-bit-first indexing.
+/// Keep data.items.len equal to ceil(bit_len / 8) and unused trailing bits zero.
+pub fn BitList(comptime options: BitListOptions) type {
+    const limit = options.limit orelse std.math.maxInt(usize);
```
GrapeBaBa, line 11: `BitList(.{ .limit = null })` to `BitList(.{ .limit = unlimited })`(using const) or `BitList(.{ .limit = .unbounded })`(using enum) seems more readable API

src/bit_array/bit_list.zig
```diff
@@ -0,0 +1,236 @@
...
+                if (remainder_bits != 0) {
+                    const mask: u8 = (@as(u8, 1) << @intCast(remainder_bits)) - 1;
+                    self.data.items[byte_len - 1] &= mask;
+                }
+            }
+            self.bit_len = bit_len;
+        }
+
+        /// Requires bit_index below bit_len; does not grow or allocate.
+        pub fn setAssumeCapacity(self: *@This(), bit_index: usize, bit: bool) !void {
```
GrapeBaBa, line 162: Refer to Zig std, `*Assume*` kind of function should not throw error

src/bit_array/bit_list.zig
wemeetagain: Not really a fan of using a const or enum.

Preference for keeping as-is or making the limit non-nullish (eg `pub fn BitList(comptime limit: usize) type`).

GrapeBaBa (approved):

Issue #622, opened 2026-09-01
Fix STF error observability


wemeetagain:
I suppose that is equivalent to the check we have [here](https://github.com/ChainSafe/lodestar/blob/1e9a530f9897d50d2d6337cd4143c8dc53667c66/packages/state-transition/src/block/processWithdrawals.ts#L76-L78) in lodestar?

```ts
if (expectedWithdrawals.length !== payload.withdrawals.length) {
  throw Error(`Invalid withdrawals length expected=${numWithdrawals} actual=${payload.withdrawals.length}`);
}
```

is it possible to add more details to the error itself, ie. in this case, what the length mismatch is

_Originally posted by @nflaig in https://github.com/ChainSafe/lodestar-z/pull/619#discussion_r3902203221_
            

---

We should figure out how to expose richer error metadata from the STF.
Eg: better logs, using a metadata struct, catching errors and inspecting the state/caches/block at a higher level, etc

spiral-ladder:
I would suppose the most basic/naive approach for now is to do a error log along side errors thrown for now, eg. https://github.com/tigerbeetle/tigerbeetle/blob/47aeb2212a255273dda508288412e537d11e4b7c/src/storage.zig#L323-L327

PR #694, opened 2026-09-15
feat: add typed stf diagnostics

GrapeBaBa:
**Motivation**

STF validation currently reports errors such as:

```text
WithdrawalsRootMismatch
```

The error omits the expected and actual roots needed to diagnose the mismatch. Related to #622.

**Description**

Capture per-module typed error details during Zig state transition, initially covering withdrawals validation.

- `stateTransition` creates local diagnostics for block processing and warns with the slot and captured details before returning the original error.
- `TransitionOpts.diagnostics` allows Zig callers to supply a target for structured inspection.
- `withdrawalsRootMismatch` captures both roots by value; callers can inspect or format them after failure cleanup.

**AI Assistance Disclosure**

Implementation primarily authored with AI assistance.


src/state_transition/block/process_withdrawals.zig
```diff
@@ -42,7 +44,11 @@ pub fn processWithdrawals(
     try types.capella.Withdrawals.hashTreeRoot(allocator, &expected_withdrawals_result.withdrawals, &expected_withdrawals_root);
 
     if (!std.mem.eql(u8, &expected_withdrawals_root, &payload_withdrawals_root)) {
-        return error.WithdrawalsRootMismatch;
+        return diagnostics.state_transition.withdrawalsRootMismatch(
```
wemeetagain, line 47: I like the ux of this

spiral-ladder (approved):
pretty nice, i think could be the start of a logger too addressing #413 , the only nit is that maybe we should have a simpler name rather than `diagnostics`? No idea yet

PR #696, opened 2026-09-15
docs: retire security implementation map

wemeetagain:
**Motivation**

The [security-maintenance rule](https://github.com/ChainSafe/lodestar-z/blob/02d61c2b9c9ca7361ea852b8f5458854f8522197/AGENTS.md#L20-L23) requires routine implementation changes to update a central inventory:

```text
Update the normative threat model only when the security contract changes; otherwise update the implementation map.
```

This leaves contributors unsure which changes need an entry and makes independent PRs conflict in the same table. Much of the inventory duplicates code, tests, and existing API documentation.

**Description**

Retire the implementation map and make security documentation maintenance depend on changes to the contract.

- Delete `docs/security/IMPLEMENTATION_MAP.md` and its references.
- Update `AGENTS.md` and `THREAT_MODEL.md` so changes that preserve security objectives, trust assumptions, trust boundaries, and supported caller obligations require no security-documentation update. Keep the requirement to verify actual call paths before downgrading findings.
- Preserve the single-addon-version restriction in `THREAT_MODEL.md`. Keep addon compatibility, `BeaconStateView.createFromBytes()` provenance, and `EpochTransitionCache` lifetime requirements beside the owning modules and APIs. Other enduring preconditions already have local documentation.
- Use one-line public binding comments for STF concurrency and cache teardown, including both teardown exports and phase0 checkpoint computation. Preserve the requirement to coordinate across workers.

**AI Assistance Disclosure**

This change was primarily AI-authored with Codex.


GrapeBaBa:
I thought if we could just transfer that docs to code comment for module/struct

bindings/src/index.d.ts
```diff
@@ -378,7 +383,17 @@ export declare class BeaconStateView {
   hashTreeRoot(): Uint8Array;
   createMultiProof(descriptor: Uint8Array): CompactMultiProof;
 
+  /**
+   * Callers must serialize this call with other `processSlots()`,
```
wemeetagain: can you make the comments added here more succinct?

Something like "Callers must serialize this call with other STF mutation functions"

nflaig (approved):
green diff looks alright to me, haven't looked at what was deleted

PR #697, opened 2026-09-15
perf: defer next epoch proposers

GrapeBaBa:
**Motivation**

Pre-Fulu cache creation and epoch transitions sample all next-epoch proposers even when a state is only used for block processing or historical replay.


GrapeBaBa:
Not sure if it is still worth for pre-FULU optimzation

wemeetagain:
yeah I think pre-fulu is not important to optimize

Issue #621, opened 2026-09-01
State transition audit follow-ups


GrapeBaBa:
## Context

Tracking issue for the state-transition audit consolidated on 2026-08-31.

- Audited snapshot: `f7e9a8bd97f4a50e556953b7e987a0626020f046`
- Consensus baseline: `consensus-specs` `v1.7.0-alpha.11`
- Scope: Phase0 through Fulu under `src/state_transition`, including reachable N-API callers

Locations came from the audited snapshot and should be revalidated against current `main` when each item is picked up.

## Security, isolation, and consensus correctness

- [x] **1. High: validate short seed validator buffers before slicing.** `src/state_transition/load_state.zig` can panic on an aligned but undersized N-API seed buffer. In progress: #653.
- [ ] **2. Critical: remove process-global epoch scratch race and UAF paths.** `src/state_transition/cache/epoch_transition_cache.zig` exposes borrowed slices after releasing the resize mutex. In progress: #616. Related implementation: #648 uses thread-local scratch and environment teardown. Both PRs remain open.
- [x] **3. High: reject hostile withdrawal counts without asserting.** Done: #619.
- [x] **4. High: keep the public expected-withdrawals query read-only.** Done: #620.
- [x] **5. High: accept exact two-thirds participation for justification.** `process_justification_and_finalization.zig` uses `>` where the spec requires `>=`. Done #623. 
- [x] **6. Medium: align Capella-through-Fulu execution and withdrawal processing.** Default execution-header data can currently skip processing; confirm the reachable production pre-state before changing behavior.Done #624 
- [x] **7. High: fix shuffled-index source cache capacity and metadata publication.** `committee_indices.zig` can repeatedly scan and replace entries for large validator sets. Done: #638. Tracked by #629.
- [x] **8. Medium: replace the Phase0 floating-point square root with exact integer arithmetic.** The current `f64` path can change base rewards. In progress: #654.
- [x] **9. High: initialize newly inserted pending-withdrawal balance entries.** Repeated requests for one validator can read an undefined `AutoHashMap` value. Done: #630. 
- [x] **10. Medium: populate proposer reward reporting.** Native attestation, sync-aggregate, and slashing processing leave the public reward fields at zero. In progress: #655.
- [ ] **11. Low: include the validator that fills the payload in the reported sweep count.** In progress: #656.
[...]

PR #701, opened 2026-09-16
fix: declare transition status options

GrapeBaBa:
**Motivation**

Addresses the declaration/documentation mismatch in #621, item 37. The binding already parses and uses execution-payload and data-availability statuses, but `TransitionOpts` omits both fields and says they are ignored. 

**Description**

Declare the status values already accepted by `parseOptions`, document their existing defaults, and remove the incorrect ignored-fields note.

**AI Assistance Disclosure**

Implementation was primarily AI-authored with Codex, with AI-assisted validation.


GrapeBaBa:
@spiral-ladder why ignore these two fields before?

spiral-ladder:
> @spiral-ladder why ignore these two fields before?

no good reason, in https://github.com/ChainSafe/lodestar-z/pull/333 when this was introduced we hardcoded the opts, but this should be the right fix

spiral-ladder:
> > @spiral-ladder why ignore these two fields before?
> 
> no good reason, in #333 when this was introduced we hardcoded the opts, but this should be the right fix

but thinking again, we should not even need these? In prod lodestar we always pass `ExecutionPayloadStatus.valid` to the caller and the validity is checked out of band

For the da option, seems like we always assume that it's available upstream: https://github.com/ChainSafe/lodestar/blob/07c3496aff12718b442ed3aa612bfb262bbf54ed/packages/beacon-node/src/chain/blocks/verifyBlock.ts#L189-L190

So we might not even need this PR

wemeetagain:
Imo we should keep it for now for compatibility's sake, determine later if its still necessary.

PR #705, opened 2026-09-16
fix(stf): floor the sync-aggregate proposer penalty and guard phase0 lookups

nazarhussain:
**Motivation**

Two sync committee bugs that crash the process instead of returning an error.

The proposer penalty uses `@max(0, proposer_balance - sync_participant_reward)`. The balance is unsigned, so the subtraction underflows before `@max` can clamp it and the node aborts.

Separately, `getValidatorIndices` and `getValidatorIndexMap` panic on a phase0 state. Both are callable from JavaScript, so a pre-Altair state kills the host process.

**Description**

Use saturating subtraction for the penalty. Return `SyncCommitteeCache.Error.SyncCommitteeNotAvailable` instead of panicking, and map it to the `NO_SYNC_COMMITTEE` error the binding already throws.

Adds a test for each; both fail against the old code. Tests move to a sibling `_test.zig` because tidy allows only one inline test block per module.

Mainnet is green. Minimal is green with #704.

Refs #621, items 44 and 33 (phase0 part).

**AI Assistance Disclosure**


GrapeBaBa (approved):
LGTM

PR #706, opened 2026-09-16
fix(bindings): declare sync-committee index map values as Uint32Array

nazarhussain:
**Motivation**

`SyncCommitteeCache.validatorIndexMap` is declared as `Map<number, number[]>`, but the binding emits `Uint32Array` values. TypeScript consumers get the wrong type with no error.

**Description**

Declare the values as `Uint32Array` to match what the binding returns.

Changing the binding to emit plain arrays would also fix the mismatch, but the map is rebuilt on every call and that path runs once per sync-committee gossip message, so this keeps the faster representation.

Refs #621, item 33.

**AI Assistance Disclosure**


wemeetagain (approved):

spiral-ladder:
so this change makes the apis kinda awkward on the lodestar side, opened https://github.com/ChainSafe/lodestar-z/pull/714, see that PR for more info

PR #708, opened 2026-09-16
fix: isolate binding benchmark suites in separate processes

lodekeeper-z:
Run each binding benchmark file sequentially in a fresh Node process using the public `--isolate` option from `@chainsafe/benchmark` 2.1.0. This prevents retained mainnet state and garbage collection from earlier suites from inflating unrelated native measurements, without changing the allocating API, deterministic inputs, benchmark IDs, or thresholds.

Use the upstream CLI directly, with no local runner adapter or internal-module coupling. The framework owns comparison, combined CSV persistence, the single GitHub report, worker lifecycle, and rejection of incomplete results. Real CLI integration tests cover isolation, cancellation, failure handling, persistence, and regression exits.

Start a new `isolated-v1` local history namespace and CI cache identity so shared-process measurements are not reused as baselines. Dedicated monitors should enable `--isolate` and select a fresh history profile; archive existing histories rather than copying them into the new baseline.

Consumes ChainSafe/benchmark#55.


scripts/benchmark.mjs
```diff
@@ -0,0 +1,49 @@
+import {fork} from "node:child_process";
+import {fileURLToPath} from "node:url";
+
+// Keep the framework's comparison, persistence and GitHub report in the parent.
+// Only its file execution is replaced; this adapter targets the pinned version.
+const {BenchmarkRunner} = await import(new URL("./benchmark/runner.js", import.meta.resolve("@chainsafe/benchmark")));
+const workerPath = fileURLToPath(new URL("./benchmarkWorker.mjs", import.meta.url));
+
+BenchmarkRunner.prototype.process = async function (files) {
```
wemeetagain: would this feature (running benchmarks in separate processes) be better suited as an upstream feature? Eg as a PR into `@chainsafe/benchmark`?

lodekeeper-z: Yes—the process-isolation feature belongs in `@chainsafe/benchmark`, rather than a Lodestar-Z-specific runner. I checked upstream: [`BenchmarkRunner.process()`](https://github.com/ChainSafe/benchmark/blob/737944321ed887fcc694ebd24c6616fbc9474bc0/src/benchmark/runner.ts) currently calls `startTests(files, this)` in one process, while [`run()`](https://github.com/ChainSafe/benchmark/blob/737944321ed887fcc694ebd24c6616fbc9474bc0/src/cli/run.ts) already owns history, comparison, and the GitHub report separately. That is a natural boundary for an opt-in, sequential process-per-file mode upstream.

That would also remove this PR’s prototype patch and internal-module coupling. The upstream mode should inherit Node flags/environment, wait for each worker to exit, and reject collection/worker failures or duplicate IDs before persisting the combined result. Lodestar-Z would then only need to enable the option and start a fresh history/cache namespace, since isolated measurements should not reuse the shared-process baseline.

I would favor that upstream-first approach over merging the local adapter as the long-term solution.

wemeetagain: Ok open a PR there.

lodekeeper-z: Opened upstream: https://github.com/ChainSafe/benchmark/pull/55.

The opt-in `--isolate` mode runs files sequentially in fresh Node processes while keeping history and the combined report in the parent. It rejects failed/crashed/missing-result workers, interrupted runs, and duplicate IDs before persistence; the existing shared-process default is unchanged.

Build, lint, type checking, and the full local unit suite passed (53 passed, 6 skipped), including 11 real CLI/process integration cases. The integration cases also passed on Node 18 and 22, and removing the worker-failure gate made the collection/execution regressions fail as intended. Independent specification/quality/security review passed. Upstream fork workflows currently need maintainer approval.

I left this Lodestar-Z PR unchanged; consumption can switch to the upstream option once it lands and is available.

lodekeeper-z:
Fixed the bindings-test timeout in 42df688e65bbe80bca3899c8b3820286da3d9564. The regression-exit test performs two CLI runs, each with a 30-second subprocess deadline, but inherited Vitest’s 5-second test timeout; attempt #2 took 5.234 seconds. Its test deadline now allows both subprocess deadlines plus cleanup. Production code and regression assertions are unchanged.

Verification: a deterministic probe adding 2.5 seconds to each CLI startup reproduced the original timeout and passed after the change (6.37 seconds). All 302 binding tests passed locally with the ERA fixture, an 8 GiB Node heap, and one worker; Biome and diff checks passed. The default local run hit the unrelated BeaconStateView heap limit before using those settings.

The pushed head is verified. CI and benchmark workflows for this head are awaiting maintainer approval; they have not passed yet. This does not change the upstream-first plan in ChainSafe/benchmark#55.

scripts/benchmark.mjs
wemeetagain: chainsafe/benchmark 2.1.0 will soon be released (CI running now). Please update this PR to use this instead.

lodekeeper-z: Updated in a9df81a4beebb6e7e0ab50773721bb78b21bff2c to the published @chainsafe/benchmark 2.1.0 and its public --isolate CLI option. Removed both local adapter/worker scripts and their lint configuration; retained the fresh isolated history/cache namespace and updated the monitor instructions and PR description.

Verification: the new direct-upstream CLI test fails with 2.0.1 (only one process runs) and passes with 2.1.0. All 303 binding tests pass, including 11 isolation integration tests. Removing --isolate makes the process-order regression fail; restoring it returns the suite to green. All three real benchmark files also passed a bounded one-iteration smoke run, not a performance comparison. Frozen-lockfile installation and changed-file lint pass; repository lint exits successfully with existing declaration-file warnings.

The signed pushed head and mergeability are verified. CI and Benchmark workflows for this head require maintainer approval; PR-title lint passed.

lodekeeper-z:
Addressed the latest CI failure in 5fea85655328d66d7e872b7a8d5e39bc56bc4551. The SIGTERM test failed before sending any signal: its fixed two-second readiness window expired during worker startup. Readiness now has a bounded 30-second poll within a 35-second test deadline; the worker-termination assertions remain unchanged. The SIGTERM fixture deliberately delays readiness by 2.2 seconds to guard against restoring the old window.

Verification: that delayed fixture failed at the original readiness assertion before the fix, then passed afterward. All 303 local binding tests pass (12 files, one worker, 8 GiB Node heap), including all 11 benchmark integration tests. Changed-file Biome and diff checks pass; independent specification, quality, and security review approved. Existing bigint native-fallback warnings remain.

The signed pushed head and mergeability are verified. New CI and Benchmark workflows require maintainer approval; they have not passed yet.

bindings/test/benchmarkRunner.test.ts
wemeetagain, line 1: @lodekeeper-z is this file still needed? I don't see it referenced

lodekeeper-z: It is still executed: the root `package.json` test script runs `vitest run bindings/test/*.test.ts`, so Vitest discovers this file without an explicit import. The latest bindings CI log confirms all 11 tests in this file passed: https://github.com/ChainSafe/lodestar-z/actions/runs/35297404972/job/105547605454.

After removing the local runner, its remaining Lodestar-Z-specific purpose is integration coverage for our `benchmark` script (`--isolate`), the fresh history namespace, and combined result persistence. The direct upstream CLI, signal-handling, and worker-failure cases now overlap with upstream coverage; those are not required here. I would keep the script/config integration checks rather than the entire original runner suite. No code changed in this reply.

wemeetagain: its extremely strange that it seems like a unit test for an upstream package... ie should not be present

lodekeeper-z: Removed the entire bindings/test/benchmarkRunner.test.ts suite in 2a53cb6eac1bf45266654490b6f4c002f3ce1401. You are right: with the runner now owned by @chainsafe/benchmark, this PR should consume its public CLI rather than retest its implementation. I did not retain the script/config test subset proposed above. The --isolate option and fresh history/cache namespace are unchanged.

All 292 remaining binding tests pass across 11 files (one worker, 8 GiB Node heap); existing bigint fallback warnings remain. The staged diff check passed. The pushed PR head and five-file PR scope are verified; no production code changed.

bindings/perf/README.md
wemeetagain, line 1: we don't need this file either

lodekeeper-z: Removed bindings/perf/README.md in b9bf8e1a471238fa51fce63a0c4c03c893a146b1. Verified the pushed head and four-file PR scope; this commit only deletes the README. No tracked references to the file remain, and the diff check passes. Runtime tests were not rerun for this documentation-only deletion.

wemeetagain (approved):

PR #709, opened 2026-09-16
feat(bindings): implement computeBlockRewards

nazarhussain:
**Motivation**

`computeBlockRewards` throws `not implemented`, so the beacon API block-rewards endpoint fails on the native path. #648 leaves the reward APIs out of scope.

**Description**

Add a native block-rewards computation and wire it to the binding. Attestation and sync-aggregate rewards are taken from the proposer-reward cache when the caller supplies one, otherwise recomputed on a throwaway clone. Slashing rewards use the fork's whistleblower quotient.

The binding takes signed block SSZ bytes and a blinded flag, matching `stateTransition`, and derives the fork from the state rather than parsing the block prefix. It returns the result synchronously; the adapter wraps it in the Promise the host interface declares.

Covers `computeBlockRewards` only. `computeAttestationsRewards` and `computeSyncCommitteeRewards` remain stubs and follow separately.

Refs #621, item 36.

**AI Assistance Disclosure**


src/state_transition/rewards/block_rewards.zig
```diff
@@ -0,0 +1,190 @@
...
+        .full_gloas => .gloas,
+    };
+}
+
+fn blockTypeOf(comptime tag: @typeInfo(AnyBeaconBlock).@"union".tag_type.?) BlockType {
+    return switch (tag) {
+        .blinded_bellatrix, .blinded_capella, .blinded_deneb, .blinded_electra, .blinded_fulu => .blinded,
+        else => .full,
+    };
+}
```
wemeetagain: These seem unnecessary. (Shouldn't there be a `block.fork()` and `block.blockType()`?

src/state_transition/rewards/block_rewards.zig
```diff
@@ -0,0 +1,190 @@
...
+
+fn computeAttestationsReward(
+    comptime fork: ForkSeq,
+    comptime block_type: BlockType,
+    allocator: Allocator,
+    io: std.Io,
+    pre_state: *CachedBeaconState,
+    body: *const fork_types.BeaconBlockBody(block_type, fork),
+) !u64 {
+    if (comptime !fork.gte(.altair)) return Error.BlockAttestationRewardUnsupportedFork;
```
GrapeBaBa:
```suggestion
    if (comptime !fork.gte(.altair)) return error.BlockAttestationRewardUnsupportedFork;
```

GrapeBaBa: more popular in Zig std

GrapeBaBa:
Looks good otherwise cayman's comment

wemeetagain (approved):

PR #715, opened 2026-09-18
chore: remove benchmark CI

spiral-ladder:
**Motivation**

It's [not even working](https://github.com/ChainSafe/lodestar-z/pull/714#issuecomment-5727046995), and we have lodekeeper-z giving daily reports now.

**Description**

yeet benchmark.yml


GrapeBaBa (dismissed):

wemeetagain (changes requested):
wait, we have #708 which at least resolves lodekeeper's reporting

wemeetagain:
Ah, I misread the PR motivation earlier. Sorry about that. I agree with this too! branch conflict now, will approve after

spiral-ladder:
> Ah, I misread the PR motivation earlier. Sorry about that. I agree with this too! branch conflict now, will approve after

Seems like your previous PR fixed the missing comment I think:

https://github.com/ChainSafe/lodestar-z/pull/715#issuecomment-5731653175

but still not sure if this is entirely useful?

wemeetagain (approved):

wemeetagain:
Yeah I think we just rely on daily benchmark runs (which can be tuned and tweaked to be signal) rather than also add flaky CI benchmarks (which will be ignored noise).

PR #716, opened 2026-09-20
perf: avoid proposer lookahead allocation

GrapeBaBa:
**Motivation**

`processProposerLookahead` copied the fixed-size Fulu `proposer_lookahead` vector into a heap allocation and immediately freed it after epoch processing.

**Description**

Use a caller-owned stack array and add `proposerLookaheadInto` helpers on `BeaconState` and `AnyBeaconState` to fill it through `getAllInto`.

**AI Assistance Disclosure**

Implementation and validation were assisted by AI.


src/fork_types/any_beacon_state.zig
```diff
@@ -896,6 +896,11 @@ pub const AnyBeaconState = union(ForkSeq) {
         return @ptrCast(try lookahead_view.getAll(allocator));
     }
 
+    pub fn proposerLookaheadInto(self: *AnyBeaconState, proposer_lookahead: *[ct.fulu.ProposerLookahead.length]u64) !void {
```
spiral-ladder: I think `out` is good enough here

```suggestion
    pub fn proposerLookaheadInto(self: *AnyBeaconState, out: *[ct.fulu.ProposerLookahead.length]u64) !void {
```

spiral-ladder (commented):
While unrelated to this PR, we have a naming issue across this repo, our `*Slice` methods allocate (when slicing shouldn't)

spiral-ladder (approved):

PR #719, opened 2026-09-21
docs: call a source file a file, not a module

spiral-ladder:
**Motivation**

We were confusing two Zig concepts in our comments:

A Zig [module] is a collection of source files with one root source file.

A [source file] is just a struct.

[module]: https://ziglang.org/documentation/0.16.0/#Compilation-Model
[source file]: https://ziglang.org/documentation/0.16.0/#Source-File-Structs

**Description**

Rename appropriate file vs module.

**AI Assistance Disclosure**

Found with claude while reviewing #718


test/tidy.zig
nazarhussain: This still mentions `module`

nazarhussain: This still mentions the module.

nazarhussain:
One addition to the context, a module can be a single source file module and that's not required to be a `root.zig`, that name is just a convention. e.g. in our case we have `src/hex.zig` that is source file as well as module itself.

nazarhussain (approved):

PR #721, opened 2026-09-21
refactor(epoch-cache): make effectiveBalanceIncrementsSet append instead

### LINKED ISSUE #720: investigate process pending deposits
Slight regression on process pending depositsi. The peaks are almost the same as old version and on average it seems worse

feat3-sas (left) versus stable-sas (right)

Image

spiral-ladder:
**Motivation**

Before this, adding `preset.MAX_PENDING_DEPOSITS_PER_EPOCH` new validators could copy up to
`preset.MAX_PENDING_DEPOSITS_PER_EPOCH` of
`effective_balance_increments`. This commit adds a fast path that makes
use of the fact that in `beforeEpochTransition()` we clone and init a
new `effective_balance_increments` before processing deposits, which
allow us to extend the private array in place instead of copying.

Beyond that we also rename `effectiveBalanceIncrementsSet` into `effectiveBalanceIncrementsAppend`
because it was already append-only anyway, and this makes the contract 
much clearer (we can assert `index == old.items.len`), and makes the conditionals in the
code cleaner too (see `grow` block)

Napkin math:

Each entry is a u16, so the array uses approximately: validator count * 2 bytes.
For mainnet, that's roughly 1,000,000 validators: about 1.907 MiB. 16
full copies move about 1.907 MiB * 16 ~= 30.52MiB

**Description**

Add a fast path that allow us to grow `effective_balance_increments` in-place
after `beforeEpochTransition()`, when the ref count is 1.

Closes #720

**AI Assistance Disclosure**

Local benchmarks + investigation done with fable


src/state_transition/cache/epoch_cache.zig
```diff
@@ -903,7 +903,24 @@ pub const EpochCache = struct {
...
+            // Fast path: allow `self.effective_balance_increments` to grow
+            // in-place while this is the only reference.
+            // Avoids up to `preset.MAX_PENDING_DEPOSITS_PER_EPOCH`
+            // full-array copies of `effective_balance_increments`
+            // (which is a function of validator count) every epoch.
+            //
+            // NOTE: We run `beforeEpochTransition()` before this,
+            // which guarantees a ref count of 1 since we init a new
+            // rc prior to processing pending deposits.
+            rc._ref_count.load(.acquire) == 1)
```
nazarhussain: Directly accessing a private field does not feel right. If needed we should add a public accessor for it.

nazarhussain (commented):
**Two of the three guards have no test coverage.** I removed each in turn:

| guard removed | result |
|---|---|
| `rc._ref_count.load(.acquire) == 1` | caught ✅ |
| `@memset(rc.instance.items[old_len..], 0)` | **not caught** |
| `index < rc.instance.capacity` | **not caught** |

src/state_transition/cache/epoch_cache.zig
spiral-ladder: yeah this is valid critique, [d0c173e](https://github.com/ChainSafe/lodestar-z/pull/721/commits/d0c173eecd8db7baa09870f899a5181682294557) should address this, it's based on Rust's [Arc::get_mut](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) design

spiral-ladder:
This ended up being more involved, edeited the description

spiral-ladder:
> **Two of the three guards have no test coverage.** I removed each in turn:
> 
> guard removed	result
> `rc._ref_count.load(.acquire) == 1`	caught ✅
> `@memset(rc.instance.items[old_len..], 0)`	**not caught**
> `index < rc.instance.capacity`	**not caught**

the bottom 2 guards are now removed, so i think those aer not necessary anymore

src/state_transition/utils/ref_count.zig
```diff
@@ -46,6 +46,15 @@ pub fn RefCount(comptime T: type) type {
             return self.instance;
         }
 
+        /// Mutable access to the instance when the caller holds the only reference, else null.
+        ///
+        /// The caller must ensure exclusive access to `self` and its instance
+        /// for the lifetime of the returned pointer.
+        pub fn getMutIfUnique(self: *@This()) ?*T {
+            if (self._ref_count.load(.acquire) != 1) return null;
```
GrapeBaBa, line 54: I suspect this is not quite useful when allowing concurrent access

spiral-ladder: Do we have concurrent access (other than from historical regen) atm? I think we need to rework the cache to support historical regen, but for the current case this seems ok

GrapeBaBa: For this naive RC impl, I thought it is not easy to support `getMutIfUnique` like rust and we don't need it right now, without this check seems also OK for now

GrapeBaBa (approved):

nazarhussain (approved):

PR #722, opened 2026-09-21
refactor(stf): make allocator ownership explicit

GrapeBaBa:
**Motivation**

[#621](https://github.com/ChainSafe/lodestar-z/issues/621) item 45 identified allocator mismatches in state-transition caches. The shared N-API allocator masks them, but distinct Zig callers can free memory through the wrong allocator.

**Description**

Make allocator ownership explicit at state-transition cache boundaries.

- `EpochCache` uses its own allocator for cache-owned sync-committee and effective-balance storage.
- Caller-owned attesting-index results receive the output allocator.
- Borrowed process-global compounding-validator storage retains its owner allocator and exposes `append(value)` without an allocator argument.
- `EpochTransitionCache` retains the allocator for its owned lists and `balances`.
- Regression tests cover distinct allocators, cache growth, and OOM cleanup.

No consensus behavior or N-API surface changes.

**Validation**

- `zig fmt --check .`
- `zig build test:tidy`
- `zig build test:state_transition`
- `zig build`

**AI Assistance Disclosure**

Implementation was primarily AI-authored.


spiral-ladder:
honestly with the way the epochcache is used, it might even be better to consider having it be unmanaged (not have its own allocator), this way we avoid this class of errors where we misuse allocators

spiral-ladder:
It also appears now EpochCache is half managed and half unmanaged, which is bad, let's make it fully one or the other (i prefer fully unmanaged)

GrapeBaBa:
> honestly with the way the epochcache is used, it might even be better to consider having it be unmanaged (not have its own allocator), this way we avoid this class of errors where we misuse allocators

Why you prefer unmanaged here, I thought managed seems better for epochcache

GrapeBaBa:
> It also appears now EpochCache is half managed and half unmanaged, which is bad, let's make it fully one or the other (i prefer fully unmanaged)

It actually need two different allocators, I thought this is not simple managed/unmanaged comparision

wemeetagain:
> let's make it fully one or the other

I don't think it needs to be this way?

If the cache creates some caller-owned memory, then it makes sense to pass an allocator in which allocates that memory.
If the cache maintains some internal memory that gets updated occasionally, then using its own allocator makes sense.

PR #723, opened 2026-09-21
refactor: transfer shuffling indices on success

GrapeBaBa:
**Motivation**

`computeEpochShuffling` freed the caller's `active_indices` on error through an unconditional `errdefer`. That made the failure contract ambiguous and could free a slice with an allocator that did not own it. A later reference-count allocation also occurred after the shuffling had taken ownership.

**Description**

Keep `active_indices` with the caller until all fallible initialization succeeds, then transfer ownership to the resulting shuffling.

- Remove the callee-side free from `computeEpochShuffling`.
- Bind each caller-owned slice before initialization and free it only if initialization fails.
- Preallocate the reference-count cell and use infallible `RefCount.initIn` for the final transfer.
- Keep the repeated `ArrayList` to owned-slice setup in `initEpochShufflingRcFromList`.
- Add a memory-safety regression test for the failed initialization path.

`RefCount.init` keeps its existing behavior and delegates to `initIn`; the `EpochShuffling` pointer layout remains unchanged.

Verification:

```text
zig fmt --check
zig build test:tidy
zig build test:state_transition -- --test-filter "computeEpochShuffling"
```

**AI Assistance Disclosure**

Implementation primarily AI-authored, directed and reviewed by the author.


src/state_transition/cache/epoch_cache.zig
```diff
@@ -147,6 +147,18 @@ pub const EpochCache = struct {
...
     }
 
+    fn initEpochShufflingRcFromList(
+        allocator: Allocator,
+        state: *AnyBeaconState,
+        active_indices: *std.ArrayList(ValidatorIndex),
+        epoch: Epoch,
+    ) !*EpochShufflingRc {
+        const indices = try active_indices.toOwnedSlice(allocator);
+        errdefer allocator.free(indices);
```
spiral-ladder, line 159: can't we just do these 2 lines inline instead of a new function?

GrapeBaBa:
```zig
        const previous_shuffling_rc = rc: {
            const active_indices = try previous_active_indices_array_list.toOwnedSlice(allocator);
            errdefer allocator.free(active_indices);

            break :rc try initEpochShufflingRc(allocator, state, active_indices, previous_epoch);
        };
```
The original version is three scoped label block, so you prefer this?

spiral-ladder: I think verbose doesn't necessarily mean bad, I'd argue this is better because:

1) this fn is only used 3 times in the same file

2) more importantly it doesn't shove allocations down the call stack. I think it's generally a better pattern to not pass allocators as much as possible (I know this doesn't apply here since we already do pass an allocator down) but I think dealing with allocations as early as possible is better than stacking allocating functions

GrapeBaBa:
seems not correctly handle the ownership than before

PR #726, opened 2026-09-22
fix(stf): self-heal progressive balance drift

GrapeBaBa:
**Motivation**

Port the state-transition behavior from Lodestar PRs #10046 and #10143. Progressive target balance caches are maintained incrementally and can drift from the authoritative participation data used during epoch transition. The Lodestar behavior records the drift and repairs the cache so fork-choice consumers do not retain stale values.

**Description**

Add non-fatal progressive balance validation and metric attribution to the Zig state transition.

- `EpochTransitionCache.init`
  - Recomputes current and previous unslashed target balances for Altair and later forks.
  - Increments `lodestar_stfn_progressive_balances_mismatches_total` with `target="current"` or `target="previous"` for each mismatch.
  - Replaces mismatched cache values with the recomputed values.
- State-transition metrics
  - Adds the labeled mismatch counter and its cleanup/init plumbing.
- Regression coverage
  - Forces a current-balance mismatch at an epoch boundary and verifies the labeled metric.

**AI Assistance Disclosure**

Implementation and validation were primarily authored with Codex assistance.


nazarhussain (approved):

src/state_transition/metrics.zig
```diff
@@ -130,6 +136,14 @@ pub fn init(allocator: Allocator, io: std.Io, comptime opts: m.RegistryOpts) !vo
         metric_opts,
     );
     errdefer proposer_rewards.deinit();
+    var progressive_balances_mismatches = try Metrics.ProgressiveBalancesMismatches.init(
+        allocator,
+        io,
+        "stfn_progressive_balances_mismatches_total",
+        .{ .help = "Total count of progressive balance cache mismatches by target balance" },
+        metric_opts,
+    );
```
nazarhussain: One observation. We have metric with same name in JS that increments that behave a bit differently. As per my understanding it increment once per epoch and don't have labels. Here we introduced labels but no dashboards are consuming it. 

https://github.com/ChainSafe/lodestar/blob/52a50304a414f5e7c2f633657c9310940e7415cc/packages/state-transition/src/metrics.ts#L151-L154

GrapeBaBa: #10143 add the label, but still open

wemeetagain (approved):

PR #727, opened 2026-09-22
perf: overlap Fulu shuffling with epoch processing

GrapeBaBa:
**Motivation**

Fulu epoch shuffling currently runs on the state-transition thread after the other epoch steps, so the shuffling work extends the epoch boundary wall-clock time. The shuffling input can be snapshotted after `before_process_epoch` and does not require access to the mutating `BeaconState` while it runs.

**Description**

Start Fulu and later epoch shuffling with `std.Io.async` after the active-index snapshot is ready. The main epoch processing continues while the worker builds `EpochShuffling`; `processProposerLookahead` awaits the future and publishes the result, and `afterProcessEpoch` reuses the reference-counted result.

- Keep pre-Fulu processing on the existing synchronous path.
- Give the worker an owned active-index copy and precomputed attester seed.
- Propagate worker errors through the existing state-transition error path.
- Cancel and clean up an unfinished future from `EpochTransitionCache.deinit`.
- Extend the epoch benchmark to start the same pipeline before `processEpoch`.

Local ReleaseSafe benchmark evidence on the Fulu ERA fixture:

```text
epoch(segmented):     about 244 ms -> 216 ms
proposer_lookahead:   about 133 ms -> 106 ms while overlapping other epoch work
```

The result is a wall-clock overlap measurement. The total shuffle CPU work remains the same.

Validation:

```text
zig build test:state_transition -Doptimize=ReleaseSafe
zig build test:swap_or_not_shuffle -Doptimize=ReleaseSafe
zig build test:tidy
zig fmt --check <changed files>
```

**AI Assistance Disclosure**

Implementation and code exploration were assisted by AI; the changes were reviewed against the existing Zig 0.16 `std.Io` and ownership patterns.


GrapeBaBa:
@codex review

wemeetagain:
@lodekeeper-z can you run before/after benchmarks and post the results here? Should be a visible performance improvement by doing epoch processing and shuffling computation in parallel.

lodekeeper-z:
@wemeetagain measured a repeatable improvement: **200.762 → 173.832 ms per epoch, 13.4% less wall time (1.155× speedup)** on this Fulu fixture.

### Before / after

- Before: base `0f7ce2d2fe45430d728c748644249c4e01e7e132`
- After: PR head `0d9763675e30dc2e0f8f31447df98f16073ecb2d`
- Same host: Intel i5-1135G7, 4 cores / 8 threads; Zig 0.16.0, `ReleaseFast`, native target, normal multi-core scheduling.
- Same ERA: `mainnet-01628-47ac89fb.era`, Fulu state at slot 13336576, 2,178,249 validators.
- Two independent processes per revision, alternating after/before/after/before. Each process performs 5 untimed warmups, then 50 measured iterations. Clone/cleanup hooks remain outside timing; cache initialization, epoch processing, cache finalization, and state-root calculation are inside it. No production code changes.

| Run | Before mean ± σ | After mean ± σ |
| --- | --- | --- |
| 1 | 201.336 ± 2.911 ms | 173.419 ± 4.455 ms |
| 2 | 200.187 ± 3.909 ms | 174.244 ± 5.999 ms |

The headline is the equal-weight mean of these runs, saving **26.930 ms/epoch**. This is a single-fixture, single-host result, not a claim about every epoch or CPU.

### Important benchmark-boundary caveat

The checked-in `epoch(non-segmented)` and `epoch(segmented)` cases omit `EpochCache.afterProcessEpoch`. On the base, that omitted method constructs the shuffling; this PR moves that work inside `processEpoch` and reuses it afterward. Comparing only the checked-in cases therefore charges the PR for work that is excluded on the base.

I ran the unchanged suite first: its second segmented run showed 124.608 ms before versus 175.401 ms after. **That is not an apples-to-apples epoch-transition regression.** The PR's segmented breakdown showed about 0.714 ms to start shuffling and 61.160 ms in proposer lookahead, versus 15.667 ms in base lookahead, but the base still owes the shuffling construction afterward.

For the comparison above, I applied the same benchmark-only adaptation in both detached worktrees: extend `ProcessEpochBench.run` immediately after `processEpoch` with the slot advance/cache finalization below, then hash the state; run only this case with five explicit warmups. This closes the moved-work boundary rather than adding an independently measured shuffle time to an unrelated timing sample.

```zig
const state = BenchState.cloned_cached_state.state;
const next_slot = ((state.slot() catch unreachable) / preset.SLOTS_PER_EPOCH + 1) * preset.SLOTS_PER_EPOCH;
state.setSlot(next_slot) catch unreachable;
BenchState.cloned_cached_state.epoch_cache.afterProcessEpoch(state, &cache) catch unreachable;
const root = (state.hashTreeRoot() catch unreachable).*;
```

This is an isolated epoch lifecycle using the existing benchmark snapshot, not a replay of all intervening slots/blocks. Every warmup/measured iteration checks that its resulting state root stays identical. Both revisions produced the same root:

```text
6f4ee97b7e2551e2685a9ee3f25349ac8f91063a705631d7fb06cd86a286359e
```

Command in each worktree (with the benchmark-only adaptation):

```sh
zig build run:bench_process_epoch -Doptimize=ReleaseFast \
  -Ddownload_era_options.era_out_dir=/path/to/shared/fixtures/era
```

ERA SHA-256: `f068f637403729af513f91940165a338f63dde471e03f5747fb90a04d731fc4d`.

The measured improvement is consistent with overlapping shuffling with other epoch work. I recommend including `afterProcessEpoch` in an end-to-end benchmark case so future before/after comparisons retain the same work boundary. Full raw logs and revision-specific harness patches are retained locally under `/home/cayman/.hermes/state/github-autopilot-artifacts/pr727/`.

GrapeBaBa:
> @lodekeeper-z can you run before/after benchmarks and post the results here? Should be a visible performance improvement by doing epoch processing and shuffling computation in parallel.

Current benchmark not include `after_epoch_process` which includes the shuffle, so it should be a bit slower than before. But from lodestar dashboard we can see the `after_epoch_process` taking 120-130ms. I also found `before_epoch_process` 50-60ms in bench but 110-140ms lodestar dashboard, I think it is more accurate to see the `epoch_transtion_time` in dashboard

wemeetagain (dismissed):
Looks good

src/state_transition/utils/epoch_shuffling.zig
```diff
@@ -123,6 +125,18 @@ pub fn computeEpochShuffling(allocator: Allocator, state: *AnyBeaconState, activ
     return EpochShuffling.init(allocator, seed, epoch, active_indices);
 }
 
+pub fn computeEpochShufflingForFork(
```
spiral-ladder: what's the difference between using this and `computeEpochShuffling`?

spiral-ladder (dismissed):
lgtm but prob worth refactoring `computeEpochShuffling` and `computeEpochShufflingForFork`

GrapeBaBa:
> lgtm but prob worth refactoring `computeEpochShuffling` and `computeEpochShufflingForFork`

Yes, I closed a PR to refactor the ownership of `computeEpochShuffling` but seems related to `RefCount` ownership, need to rework on that later

src/state_transition/utils/epoch_shuffling.zig
wemeetagain: `computeEpochShuffling` includes code blocks for all forks and a runtime switch to select the codeblock.
`computeEpochShufflingForFork` is a codeblock for a single comptime-specified fork.

wemeetagain (approved):

